PR #53, opened 2018-05-16
WIP: Enable mariadb tests

hausdorff:
WIP PR. Need to fiddle with minikube until CI passes.


joeduffy:
This is a few months old; closing out, please reopen with fresh code when the time is right.

PR #73, opened 2018-06-13
Add `getInputs` to CustomResources

hausdorff:
It is often very useful for several Kubernetes objects to be able to share data at the _inputs_ level, without taking an explicit dependency, _e.g._, an object's labels. This commit introduces the `getInputs` function, which returns the inputs given to the constructor of a Kubernetes `CustomResource`, as well as transitioning the inputs to use `pulumi.Input<T>` instead of the "naked" types `T`.


pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -18,19 +18,23 @@ export namespace {{Group}} {
...
       *
       * @param name The _unique_ name of the resource.
       * @param args The arguments to use to populate this resource's properties.
       * @param opts A bag of options that control this resource's behavior.
       */
-      constructor(name: string, args?: inputApi.{{Group}}.{{Version}}.{{Kind}}, opts?: pulumi.ResourceOptions) {
+      constructor(name: string, args: inputApi.{{Group}}.{{Version}}.{{Kind}}, opts?: pulumi.ResourceOptions) {
           let inputs: pulumi.Inputs = {};
           {{#Properties}}
           inputs["{{Name}}"] = {{{DefaultValue}}};
```
swgillespie: Is this necessary still, now that `args` is required? I see in the generated code this will generate `args ? args.prop : undefined` even though `args` can't be undefined now.

Also, are we OK with the breaking change here? I'm fine with it, as long as we're aware of it.

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -18,19 +18,23 @@ export namespace {{Group}} {
...
           {{/Properties}}
           super("kubernetes:{{APIVersion}}:{{Kind}}", name, inputs, opts);
+          this.__inputs = args;
```
swgillespie: Should this be `args` or `inputs`? Or are we expecting that they'll always be equal at this point?

hausdorff: Resolved offline. `this.__inputs` is different from `inputs`, so we'll be keeping this code as-is.

pkg/gen/node-templates/provider.ts.mustache
hausdorff: Resolved offline. We can get rid of the ternary operator, but `{{{DefaultValue}}}` needs to stay because the Mustache generation code needs to munge `inputs["apiVersion"]`.

PR #85, opened 2018-07-03
Implement await logic for various RBAC APIs

hausdorff:


pack/nodejs/package.json
```diff
@@ -1,12 +1,13 @@
 {
     "name": "@pulumi/kubernetes",
-    "version": "v0.13.1-dev-1528839233-g598db27-dirty",
+    "version": "v0.14.1-dev-1530572194-g218d72d-dirty",
```
joeduffy: @ellismg Didn't we do something special in other repos to prevent us from committing dirty versions?

pkg/await/await.go
```diff
@@ -42,22 +42,36 @@ import (
...
-	coreV1ResourceQuota                  = "v1/ResourceQuota"
-	coreV1Secret                         = "v1/Secret"
-	coreV1Service                        = "v1/Service"
+	appsV1Deployment                            = "apps/v1/Deployment"
+	appsV1Beta1Deployment                       = "apps/v1beta1/Deployment"
+	appsV1Beta2Deployment                       = "apps/v1beta2/Deployment"
+	autoscalingV1HorizontalPodAutoscaler        = "autoscaling/v1/HorizontalPodAutoscaler"
+	extensionsV1Beta1Deployment                 = "extensions/v1beta1/Deployment"
+	storageV1StorageClass                       = "storage.k8s.io/v1/StorageClass"
+	coreV1ConfigMap                             = "v1/ConfigMap"
```
joeduffy: I _thought_ these were projected as `core/v1/...`.  Does this PR change this?

pkg/await/await.go
```diff
@@ -42,22 +42,36 @@ import (
...
+	coreV1PersistentVolume                      = "v1/PersistentVolume"
+	coreV1PersistentVolumeClaim                 = "v1/PersistentVolumeClaim"
+	coreV1Pod                                   = "v1/Pod"
+	coreV1ReplicationController                 = "v1/ReplicationController"
+	coreV1ResourceQuota                         = "v1/ResourceQuota"
+	coreV1Secret                                = "v1/Secret"
+	coreV1Service                               = "v1/Service"
+	coreV1ServiceAccount                        = "v1/ServiceAccount"
+	extensionsV1Beta1Ingress                    = "extensions/v1beta1/Ingress"
+	rbacAuthorizationV1ClusterRole              = "rbac.authorization.k8s.io/v1/ClusterRole"
```
joeduffy: I'm sure you were intentional about this, and that this is just flowing through the underlying Kubernetes object model, but for my education, why the fully qualified name, versus just `rbac/v1/...`?

pkg/await/awaiters.go
```diff
@@ -406,6 +406,86 @@ func untilCoreV1ServiceInitialized(
...
+	clientForResource dynamic.ResourceInterface, obj *unstructured.Unstructured,
+) error {
+	//
+	// A ServiceAccount is considered initialized when the controller adds the default secret to the
+	// secrets array (i.e., in addition to the secrets specified by the user).
+	//
+
+	specSecrets, _ := pluck(obj.Object, "secrets")
+	var numSpecSecrets int
+	if specSecretsArr, isArr := specSecrets.([]interface{}); isArr {
```
joeduffy: What could this be otherwise, if not an array?

pkg/gen/typegen.go
```diff
@@ -255,20 +255,33 @@ func createGroups(definitionsJSON map[string]interface{}, generatorType gentype)
 				return linq.From([]KindConfig{})
 			}
 
-			// Make fully-qualified group name.
-			fqGroup := d.gvk.Group
+			// Make fully-qualified and "default" GroupVersion. The "default" GV is the `apiVersion` that
+			// appears when writing Kubernetes YAML (e.g., `v1` instead of `core/v1`), while the
+			// fully-qualified version is the "official" GV (e.g., `core/v1` instead of `v1` or
+			// `admissionregistration.k8s.io/v1alpha1` instead of `admissionregistration/v1alpha1`).
+			defaultGroupVersion := d.gvk.Group
```
joeduffy: Unless I am misreading, `defaultGroupVersion` is appears to be reassigned on all branches below (unless I'm misreading), so should it be `var defaultGroupVersion string`, like `fqGroupVersion`?

pkg/await/await.go
hausdorff: This PR does _not_ change this. Pulumi refers to these as `core/v1/...`, but the API server refers to them as `v1/...`. The await logic was built on top of the raw types exposed by the API server, rather than our projection of those types, which is why we refer to the API types using these non-fully-qualified names.

pkg/await/await.go
hausdorff: The API server does not understand `rbac/v1/...`. It requires a fully-qualified group name in this case.

pkg/await/awaiters.go
hausdorff: It can be nothing other than an array. If we get something other than an array, that's a bug in client-go.

pkg/gen/typegen.go
hausdorff: That's right, this was fixed in a subsequent push.

pack/nodejs/package.json
hausdorff: Drat. I had meant to remove this.

hausdorff: Actually, _should_ I remove this? It seems like we do want to depend on v0.14.

ellismg: This is the version of the package itself.  Whatever we check in here is a lie because the next build will overwrite it with that commit’s hash.

I will submit a PR tomorrow to fix this in a similar way to how we handle the other providers.

So either revert it or don’t it doesn’t matter either way.

pkg/await/await.go
```diff
@@ -116,6 +142,18 @@ func Creation(
...
+	case rbacAuthorizationV1Alpha1ClusterRoleBinding:
+	case rbacAuthorizationV1Alpha1Role:
+	case rbacAuthorizationV1Alpha1RoleBinding:
+	case rbacAuthorizationV1Beta1ClusterRole:
+	case rbacAuthorizationV1Beta1ClusterRoleBinding:
+	case rbacAuthorizationV1Beta1Role:
+	case rbacAuthorizationV1Beta1RoleBinding:
 		break
 
 	// TODO(hausdorff): Find some sensible default for unknown kinds.
```
lukehoban: In other repos, we've generally made a point to open issues to track `TODO`s in code.  This may be valuable here as well so we have an enumeration of the remaining areas we have similar work we need to do.  

We've been using the `// TODO[pulumi/pulumi-kubernetes#1234] Find some sensible default for unknown kinds` comment format.

pkg/await/await.go
```diff
@@ -328,6 +376,20 @@ func Deletion(
...
+	case rbacAuthorizationV1Role:
+	case rbacAuthorizationV1RoleBinding:
+	case rbacAuthorizationV1Alpha1ClusterRole:
+	case rbacAuthorizationV1Alpha1ClusterRoleBinding:
+	case rbacAuthorizationV1Alpha1Role:
+	case rbacAuthorizationV1Alpha1RoleBinding:
+	case rbacAuthorizationV1Beta1ClusterRole:
+	case rbacAuthorizationV1Beta1ClusterRoleBinding:
+	case rbacAuthorizationV1Beta1Role:
+	case rbacAuthorizationV1Beta1RoleBinding:
```
lukehoban: A shame that we have to repeat this large block of constants three times here, and looks like this could be fragile to forgetting to add/remove in one of these places in the future causing a subtle bug.  Since this appears to be here just to be able to error in case of unknown kinds, can the logic be inverted, or the ID looked up early and compared against a single maintainer list of supported kinds so that this doesn't need to be inlined in each of the operations:?

hausdorff: Sure, let's do it in a follow up PR.

pkg/await/await.go
hausdorff: Actually this is causing enough problems that we should just fix this now, in the same PR as we re-organize the `switch` statement, which you mention doing below.

Issue #10, opened 2018-02-13
Apply auto-naming to `resource.metadata.name`


lukehoban:
The "name" property in Kubernetes resources is `res.metadata.name`, not `res.name` as is common in other providers.  We should consider applying our AutoNaming to this nested property.

protometa:
Kubernetes provides a natural solution to this in the form of namespaces. There is some discussion in kubernetes/helm#2060 about the pros and cons of leveraging namespaces more directly as opposed to the the current Helm convention of prefixing names. I can see using the pulumi stack name as resource namespace as a nice default to avoid conflicts, though any default should be easy enough to override in pulumi.

hausdorff:
I believe the right way to go is to let Pulumi handle the naming itself, and to let users encode which namespace they're deploying to inside the app code or using Kubernetes contexts (which would get picked up during `pulumi update`). But, I also believe this problem is not nearly as big a deal for the Kubernetes provider as it is for (say) the AWS provider, which is why I estimate its priority to be somewhat low compared to other issues we'd consider tackling in M16.

Mostly the reason I'd like Pulumi to handle naming is that I think namespaces are most useful as an organization construct rather than an application-level construct. So, _i.e._, different people or different teams tend to deploy to a specific namespace, RBAC is managed at the level of a specific namespace, _etc_. It seems to me to make sense to make the behavior of deploying an app to some organization's namespace really easy, rather than requiring users to boot up a new namespace for every instance of some app, especially since they might not have permission to create new namespaces at all.

That said, while I agree with @lukehoban that having Pulumi generate names would solve (a) and (c), I actually think (b) is not a huge problem for us, and this is the main reason I'd rate the priority lower. Most resources do not have a significant rollout phase (_e.g._, `ConfigMap`s, `Secret`s, _etc_., are trivially replaceable) and important resources that do (_e.g._, `Pod`s, `Pod`s attached to `PersistentVolume`s, _etc_.) tend to have built-in abstractions that handle rollout on behalf of the user (_i.e._, `Deployment` and `StatefulSet`, respectively). As I understand it, these abstractions were created to keep `kubectl` fairly lightweight, so that it didn't have to handle complicated rollout protocols. Overall I recommend we keep their strategy for ourselves.

hausdorff:
Oh, also, this will make testing a bit easier. If we generate the names on the user's behalf, it means that by default we don't have to worry about subsequent/parallel runs trampling over or interfering with each other.

[...]

PR #88, opened 2018-07-04
Centralize resource await logic

### LINKED ISSUE #87: Find sensible default await conditions for API objects
Right now if we don't recognize an object, we error out. In the Pulumi service logs, missing object awaiters account for a sizable number of deploy failures, _and_ when they do fail, they orphan the object, which was often successfully created.

We should handle this more gracefully, especially since we'll need to be able to use that solution for CRDs. Some options:

1. Simply refuse to create an object type that we don't support.
1. Create the object but emit a warning telling users they might want to refresh if the object is "complex", and/or request in the issues that we implement this.

hausdorff:
Following up from PR 85, this commit will pull the Kubernetes resource
await logic into a single, centralized place. Currently this await logic
is specified in 3 switch statements that each exhaustively list every
Kubernetes group/version/kind and the await logic for each.

This commit wil also fix #87. The current behavior is to error out if
the GVK is not recognized as having specific await logic, and our
resolution for this is to simply assume the resource initialized
correctly.


pkg/await/await.go
```diff
@@ -91,74 +57,21 @@ func Creation(
...
-				Group:   "",
-				Version: "v1",
-				Kind:    "Event",
-			}, obj.GetNamespace())
-			if err != nil {
-				return nil, err
+	if awaiter, exists := awaiters[id]; exists {
+		if awaiter.awaitCreation != nil {
+			conf := initAwaitConfig{
+				pool: pool, disco: disco, clientForResource: clientForResource, currentInputs: obj,
```
lukehoban: Nit: Newlines for more familiar struct layout?

pkg/await/await.go
```diff
@@ -91,74 +57,21 @@ func Creation(
 		return nil, err
 	}
 
-	// Wait until create resolves as success or error.
+	// Wait until create resolves as success or error. Note that the conditional is set up to log only
+	// if we don't have an entry for the resource type; in the event that we do, but the await logic
+	// is blank, simply do nothing instead of logging an error.
```
lukehoban: Minor - This and other comments say "instead of logging an error" - but this doesn't actually error in the `else` - just logs a info notice.

pkg/await/awaiters.go
```diff
@@ -16,15 +16,149 @@ package await
...
+	"github.com/pulumi/pulumi-kubernetes/pkg/openapi"
 	"github.com/pulumi/pulumi-kubernetes/pkg/watcher"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 )
 
+type initAwaitConfig struct {
```
lukehoban: Perhaps add comments on these even though the are internal?  Useful to understand how these are used.

Also - why the word `init` in this struct?  Is that referring to this being the `Create` operation?  If so, perhaps use `create` in the name instead?

pkg/await/await.go
```diff
@@ -253,53 +166,24 @@ func Update(
...
-		break
-
-	// TODO(hausdorff): Find some sensible default for unknown kinds.
-	default:
-		return nil, fmt.Errorf("Could not find object of type '%s'", id)
+	} else {
+		glog.V(1).Infof("No initialization logic found for object of type '%s'; defaulting to assuming initialization successful", id)
 	}
 
 	if waitErr != nil {
```
lukehoban: Any reason not to move this return next to the single place where `waitErr` is now returned?  (And keep `waitErr` localized to that scope).

pkg/await/await.go
```diff
@@ -353,48 +237,17 @@ func Deletion(
...
-		return fmt.Errorf("Could not find object of type '%s'", id)
+	if awaiter, exists := awaiters[id]; exists {
+		if awaiter.awaitDeletion != nil {
+			waitErr = awaiter.awaitDeletion(clientForResource, name)
+		}
+	} else {
+		glog.V(1).Infof("No deletion logic found for object of type '%s'; defaulting to assuming deletion successful", id)
 	}
 
 	return waitErr
```
lukehoban: Similar comment as other `waitErr` locations above.

pkg/await/awaiters.go
```diff
@@ -16,15 +16,149 @@ package await
...
+	coreV1ReplicationController: {
+		awaitCreation: untilCoreV1ReplicationControllerInitialized,
+		awaitUpdate:   untilCoreV1ReplicationControllerUpdated,
+		awaitDeletion: untilCoreV1ReplicationControllerDeleted,
+	},
+	coreV1ResourceQuota: {
+		awaitCreation: untilCoreV1ResourceQuotaInitialized,
+		awaitUpdate:   untilCoreV1ResourceQuotaUpdated,
+	},
+	coreV1Secret: {},
```
lukehoban: Probably worth being consistent on whether to include `/* NONE */`.

lukehoban: Potentially worth just definining a `noAwaiters` or similar empty struct once, and then using that in each place where no awaiters are needed?

pkg/await/awaiters.go
```diff
@@ -448,33 +577,41 @@ func untilCoreV1ServiceAccountInitialized(
...
 
-func untilExtensionsV1Beta1IngressInitialized(
-	clientForResource, clientForEvents dynamic.ResourceInterface, obj *unstructured.Unstructured,
-) error {
+func untilExtensionsV1Beta1IngressInitialized(c initAwaitConfig) error {
+	clientForEvents, err := client.FromGVK(c.pool, c.disco, schema.GroupVersionKind{
+		Group:   "",
+		Version: "v1",
+		Kind:    "Event",
+	}, c.currentInputs.GetNamespace())
```
lukehoban: Perhaps minor - should this be a helper function defined on `initAwaitConfig` that can be reused in the few places it is currently inlined?

pkg/provider/provider.go
```diff
@@ -458,20 +473,32 @@ func (k *kubeProvider) Update(
...
-		return nil, err
+	initialized, awaitErr := await.Update(k.pool, k.client, oldInputs, newInputs)
+	if awaitErr != nil {
+		var getErr error
+		initialized, getErr = k.readLiveObject(newInputs)
+		if getErr != nil {
+			// Object update/creation failed.
+			return nil, awaitErr
+		}
+		// If we get here, resource successfully created, but failed to initialize.
```
lukehoban: `created => `updated`?

More generally, "created" here and above means something like "was received by the Kubernetes master" and "failed to initialize" means something like "failed to reach the expected healthy state for the resource" - right?

pkg/provider/provider.go
```diff
@@ -532,6 +559,20 @@ func (k *kubeProvider) gvkFromURN(urn resource.URN) schema.GroupVersionKind {
...
+	clientForResource, err := client.FromResource(k.pool, k.client, obj)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the "live" version of the last submitted object. This is necessary because the server may
+	// have populated some fields automatically, updated status fields, and so on.
+	return clientForResource.Get(obj.GetName(), metav1.GetOptions{})
+}
+
```
lukehoban: Does it make sense for the implementation of `Get` that has a copy of this code to now call this function?

pkg/provider/provider.go
```diff
@@ -458,20 +473,32 @@ func (k *kubeProvider) Update(
 	newInputs := propMapToUnstructured(newResInputs)
 
 	// Apply update.
-	liveObj, err := await.Update(k.pool, k.client, oldInputs, newInputs)
-	if err != nil {
-		return nil, err
+	initialized, awaitErr := await.Update(k.pool, k.client, oldInputs, newInputs)
+	if awaitErr != nil {
+		var getErr error
+		initialized, getErr = k.readLiveObject(newInputs)
```
swgillespie: I think maybe I'm not understanding the semantics of `readLiveObject`. I see that `readLiveObject` keys off the resource name in order to read it, but is it always the case here that the resource name in `newInputs` (uniquely) refers to the resource that we just failed to create?

I'm thinking this is OK, I'm just trying to wrap my head around this.

swgillespie: I guess what's on my mind is that it seems to me that an object's fully qualified name is the ID that is ultimately sent to the engine, but here's we're just using a regular name?

pkg/provider/provider.go
```diff
@@ -567,3 +608,12 @@ func parseCheckpointObject(obj resource.PropertyMap) (oldInputs, live *unstructu
...
 	return
 }
+
+func initializationError(id string, err error, inputsAndComputed *structpb.Struct) error {
+	detail := pulumirpc.ErrorResourceInitFailed{
+		Id:         id,
+		Properties: inputsAndComputed,
+		Reasons:    []string{err.Error()},
+	}
+	return rpcerror.WithDetails(rpcerror.New(codes.Unknown, err.Error()), &detail)
```
swgillespie: I think we should probably avoid using `Unknown` in general, but we use it so pervasively it's probably not worth the time and effort to clean up.

pkg/provider/provider.go
hausdorff: @swgillespie Yes, it should always uniquely refer to the resource we failed to create. This will change when we address #10.

pkg/provider/provider.go
hausdorff: So the reason we're returning `codes.Unknown` is because returning a code that makes "more sense" would cause us to return `status.OK`, per [this code](https://github.com/pulumi/pulumi/blob/ad3b5e7ee88346bc7e960de9f953957a72f84516/pkg/resource/plugin/provider_plugin.go#L574).

This is obviously gross, but I was not comfortable changing this code, because it seems like our provider depends on this behavior, and changing our protocol did not seem like a good thing to do when introducing partial update.

I've filed https://github.com/pulumi/pulumi/issues/1615 to follow up.

pkg/await/awaiters.go
hausdorff: The thought was that it awaited resource _initialization_ rather than _creation_, where initialization implies that the object has been created _and_ initialized.

But, you're right that it's confusing. Let's change the name and just think of creation as having distinct failure modes -- one for invalid schema errors (_i.e._, the API server rejects the object) and one for initialization errors.

pkg/await/awaiters.go
hausdorff: Ah, interesting idea. I'd actually prefer to pass it in this way though. I think package-level variables are an anti-pattern.

pkg/provider/provider.go
hausdorff: If I understand the suggestion, I think it's not worth sharing since (1) the code is pretty trivial and this does not seem to add to the clarity IMO, and (2) in the other scenarios we are passed the client from the client pool, so obtaining it again is kind of a waste?

Issue #99, opened 2018-07-21
Update logic for service is incorrect


hausdorff:
* Changing type `"LoadBalancer"` -> `"ClusterIP"` should trigger a replace, not an update.
* `update`'ing should wait for the service IP to be allocated, but it returns instantly

Issue #90, opened 2018-07-06
Add end-to-end tests


hausdorff:
There are several components we want to nail down for M16:

- [ ] Historically we have used minikube to run CI/CD tests, but there are a host of problems with this, like the fact that we can't create PVC, `LoadBalancer` type services, and so on. For more complicated `examples/`, it would be nice to be able to run tests on a "real" cluster.
- [ ] We want to test all of: success conditions, failure conditions, and partial failure conditions (once partial-update lands).

hausdorff:
Getting a "real" Kubernetes cluster up for CI turned out to be harder than expected, so we ended up using CarbonQL to record every intermediate state of critical resources under different update/create circumstances, and turned those into tests. So we can close this for now with #112.

Issue #81, opened 2018-06-29
Storage class rename failure


lukehoban:
Changing storage class should force replace.  Currently it errors.

```
Plan apply failed: PersistentVolumeClaim "jenkins" is invalid: metadata.annotations.volume.beta.kubernetes.io/storage-class: Invalid value: "default": field is immutable
```

PR #103, opened 2018-07-22
Add resource autonaming, fix various replacement issues

### LINKED ISSUE #68: `pulumi update` does not wait for pod restarts
If you change (_e.g._) the environment variables of a container in some pod, it should trigger a hard reset of that pod. Unfortunately, `pulumi update` will usually not wait for the pod to restart, instead simply reporting immediately that the resource was created.

hausdorff:
Add autonaming (_cf_., #10). The design of this feature (which is fairly simple) is specified in that issue.

Fix #68, #99, and #81 (correctness and replacement issues).

Add end-to-end tests for delete-before-replace semantics (partial for #90).

NOTE: We would add end to end tests for service and PVs, but they won't work until we write up the test ci cluster.

[asciicast](https://asciinema.org/a/slOPaYVTX6pZu49m1FIyHYCci?autoplay=1)

### DIFF
--- a/pkg/await/awaiters.go
+++ b/pkg/await/awaiters.go
@@ -141,6 +141,8 @@ var awaiters = map[string]awaitSpec{
 		awaitCreation: untilCoreV1PersistentVolumeClaimBound,
 	},
 	coreV1Pod: {
+		// NOTE: Because we replace the Pod in most situations, we do not require special logic for the
+		// update path.
 		awaitCreation: untilCoreV1PodInitialized,
 		awaitDeletion: untilCoreV1PodDeleted,
 	},
--- a/pkg/provider/naming.go
+++ b/pkg/provider/naming.go
@@ -1,0 +1,66 @@
+package provider
+
+import (
+	"fmt"
+	"math/rand"
+	"time"
+
+	"github.com/pulumi/pulumi/pkg/tokens"
+	"github.com/pulumi/pulumi/pkg/util/contract"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const annotationInternalPrefix = "pulumi.com/"
+const annotationInternalAutonamed = "pulumi.com/autonamed"
+
+var dns1123Alphabet = []rune("abcdefghijklmnopqrstuvwxyz0123456789")
+
+// assignName generates a name for an object. Uses DNS-1123-compliant characters. All auto-named
+// resources get the annotation `pulumi.com/autonamed` for tooling purposes.
+func assignNameIfAutonamable(obj *unstructured.Unstructured, base tokens.QName) {
+	contract.Assert(base != "")
+	if obj.GetName() == "" {
+		obj.SetName(fmt.Sprintf("%s-%s", base, randString(8)))
+		setAutonameAnnotation(obj)
+	}
+}
+
+// adoptOldNameIfUnnamed checks if `newObj` has a name, and if not, "adopts" the name of `oldObj`
+// instead. If `oldObj` was autonamed, then we mark `newObj` as autonamed, too.
+func adoptOldNameIfUnnamed(newObj, oldObj *unstructured.Unstructured) {
+	contract.Assert(oldObj.GetName() != "")
+	if newObj.GetName() == "" {
+		newObj.SetName(oldObj.GetName())
+		if isAutonamed(oldObj) {
+			setAutonameAnnotation(newObj)
+		}
+	}
+}
+
+func setAutonameAnnotation(obj *unstructured.Unstructured) {
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[annotationInternalAutonamed] = "true"
+	obj.SetAnnotations(annotations)
+}
+
+func isAutonamed(obj *unstructured.Unstructured) bool {
+	annotations := obj.GetAnnotations()
+	autonamed := annotations[annotationInternalAutonamed]
+	return autonamed == "true"
+}
+
+func randString(n int) string {
+	b := make([]rune, n)
+	for i := range b {
+		b[i] = dns1123Alphabet[rand.Intn(len(dns1123Alphabet))]
+	}
+	return string(b)
+}
+
+// Seed RNG to get different random names at each suffix.
+func init() {
+	rand.Seed(time.Now().UTC().UnixNano())
+}
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -148,8 +148,8 @@ func (k *kubeProvider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (
 	//    k8s.core.v1.Service
 	// 2. req.Olds, the last version submitted from a custom resource.
 	//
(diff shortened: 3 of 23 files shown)


pkg/provider/naming.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+	}
+}
+
+func randString(n int) string {
+	var alphabet = []rune("abcdefghijklmnopqrstuvwxyz0123456789")
+
+	b := make([]rune, n)
+	for i := range b {
+		b[i] = alphabet[rand.Intn(len(alphabet))]
```
pgavlin: Note that the default `rand` source is seeded with the same value each time the process starts, so this will be essentially deterministic. If that matters, you should seed the source in this package's `init` function.

pkg/provider/naming.go
```diff
@@ -0,0 +1,44 @@
...
+func assignName(obj *unstructured.Unstructured, base tokens.QName) {
+	if !renameBlacklist.Has(obj.GroupVersionKind().String()) {
+		replaceNameIfUnnamed(obj, fmt.Sprintf("%s-%s", base, randString(8)))
+	} else {
+		replaceNameIfUnnamed(obj, string(base))
+	}
+}
+
+// replaceNameIfUnnamed will replace an object's name if it doesn't already have one.
+func replaceNameIfUnnamed(obj *unstructured.Unstructured, name string) {
```
pgavlin: `setNameIfUnnamed` might be a better name--in a Pulumi context, `replace` has pretty specific connotations.

pkg/provider/provider.go
```diff
@@ -205,7 +208,6 @@ func (k *kubeProvider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (
 		}
 	}
 
-	// We currently don't change the inputs during check.
 	return &pulumirpc.CheckResponse{Inputs: newResInputs, Failures: failures}, nil
```
pgavlin: will `newResInputs` contain the autoname, then?

pkg/provider/naming.go
```diff
@@ -0,0 +1,44 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+var renameBlacklist = sets.NewString("v1/Namespace")
+
+// assignName generates a name for an object if it doesn't already have one. Uses DNS-1123-compliant
+// characters.
+func assignName(obj *unstructured.Unstructured, base tokens.QName) {
+	if !renameBlacklist.Has(obj.GroupVersionKind().String()) {
```
swgillespie: Okay, so for my own understanding, this reads to me as "autogenerate a random suffix for this thing's name if it's not a v1/Namespace". I thought [here](https://github.com/pulumi/pulumi-kubernetes/issues/10#issuecomment-407160010) you said that `v1/Namespace` isn't an exception - should we be doing the same for those, too?

pkg/provider/naming.go
```diff
@@ -0,0 +1,44 @@
...
+		if meta == nil {
+			obj.Object["metadata"] = map[string]interface{}{"name": name}
+		} else {
+			meta.(map[string]interface{})["name"] = name
+		}
+	}
+}
+
+func randString(n int) string {
+	var alphabet = []rune("abcdefghijklmnopqrstuvwxyz0123456789")
```
swgillespie: This can be a global `var`

tests/integration/autonaming/step1/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: partial_state
```
swgillespie: nit: `autonaming`

tests/integration/pod/step1/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: partial_state
```
swgillespie: same here. Don't think this really matters, it's just hygiene.

pkg/await/awaiters.go
```diff
@@ -141,6 +141,8 @@ var awaiters = map[string]awaitSpec{
 		awaitCreation: untilCoreV1PersistentVolumeClaimBound,
 	},
 	coreV1Pod: {
+		// NOTE: Because we replace the Pod in most situations, we do not require special logic for the
+		// update path.
```
swgillespie: Is this to say that there's pretty much no input change to a `Pod` that won't cause a replacement?

pkg/provider/naming.go
swgillespie: `rand.Intn` is deterministic by default (https://github.com/golang/go/issues/4509) - since I don't think we seed rand anywhere we should either 1) seed it explicitly on startup or 2) use `crypto/rand` so that we don't get in bad situations with name collisions?

swgillespie: Ah, OK, looks like we are seeding:

https://github.com/pulumi/pulumi-kubernetes/blob/8c0b67e07348696c273cfb109d1997c22a248c44/pkg/watcher/watcher.go#L188

Never mind, then.

pgavlin: We should still be explicit about the seeding in this package if it requires it.

pkg/provider/provider.go
hausdorff: It will, yes.

pkg/provider/naming.go
hausdorff: I haven't updated this yet, I'm removing it now.

pkg/provider/naming.go
hausdorff: I don't like global `var`, but this is also minor so let's do that.

pkg/await/awaiters.go
hausdorff: It's basically saying there is no input change to `Pod` won't result in a replacement, _that also requires an await_.

pkg/provider/provider.go
```diff
@@ -251,9 +253,13 @@ func (k *kubeProvider) Diff(
 	}
 	newInputs := propMapToUnstructured(newResInputs)
 
-	// Naive replacement strategy. We will kill and recreate a resource only if the name or namespace
-	// has changed.
-	replaces, err := forceNewProperties(oldInputs.Object, newInputs.Object, oldInputs.GroupVersionKind())
+	// If the new object doesn't have a name, it was probably autonamed, so we give it the old
+	// object's name.
+	newName := newInputs.GetName()
+	replaceNameIfUnnamed(newInputs, oldInputs.GetName())
```
pgavlin: I think this is in the wrong place: I'm pretty sure this needs to happen in Check.

pkg/await/awaiters.go
swgillespie: got it, thanks.

pkg/provider/provider.go
hausdorff: Ah, I see what's going on here. Initially when I mutated the inputs to have the new name, they weren't getting propagated to calls to `Diff`, `Create`, _etc_. The reason for this is that we weren't marshalling those mutations back and returning those -- we were just returning the inputs verbatim.

hausdorff:
Unfortunately my laptop is too wimpy to run the CI tests without becoming catastrophically unusable, so I'll have to nudge things along until the CI passes, but modulo that I think we agree that this approach is the one we want to go with.

PR #108, opened 2018-07-27
Add Helm chart bridge

hausdorff:


pack/nodejs/yarn-error.log
```diff
@@ -0,0 +1,737 @@
+Arguments: 
+  /Users/alex/.nvm/versions/node/v9.10.1/bin/node /Users/alex/.nvm/versions/node/v9.10.1/bin/yarn install
```
pgavlin: Probably don't want to check this file in :)

examples/helm/index.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright 2016-2018, Pulumi Corporation.  All rights reserved.
+
+import * as helm from "@pulumi/kubernetes/helm";
+
+const nginx = new helm.v2.Chart("simple-nginx", {
+  // Represents chart `stable/nginx-lego@v0.3.1`.
+  repo: "stable", chart: "nginx-lego", version: "0.3.1",
```
lukehoban: Curious - any particular reason to break these out vs. just accept as a single string?  I feel like I always see these mentioned by `stable/jenkins`.

examples/helm/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  values: {
+    // Override for the Chart's `values.yml` file. Zero out resource requests so it can be scheduled
+    // on our (wimpy) CI cluster.
+    nginx: {resources: null},
+    default: {resources: null},
+    lego: {resources: null},
```
lukehoban: How does the equivalent look when done directly using Helm?  Is this `null` thing something that is Pulumi-specific, or is there a model for this directly with Helm `values.yml`?

pack/nodejs/helm.ts
```diff
@@ -0,0 +1,715 @@
...
+        repo: string;
+        chart: string;
+        version: string;
+
+        values?: any;
+        transformations?: transform.Transform[];
+        fetchOpts?: FetchOpts;
+    }
+
+    // A CustomResource representing a Helm Chart.
```
lukehoban: This is actually a `ComponentResource` - though for the docs might be better to just say a "Component".

pack/nodejs/helm.ts
```diff
@@ -0,0 +1,715 @@
...
+    // server-side validity testing is executed.
+    //
+    // Important caveat: Charts often assign hard-coded names for objects; since Pulumi is not able
+    // to assign such names, any time a resource must be replaced, it will be marked as
+    // delete-before-replace. For example, in a normal Pulumi program, a change in a ConfigMap's
+    // would result in: (1) a new ConfigMap being created, (2) all Deployments referencing that
+    // ConfigMap being updated to refer to the new ConfigMap, and (3) deleting the old ConfigMap
+    // only after all the rollouts of those Deployments have succeeded. In contrast, since the
+    // chart asks for a specific name for this ConfigMap, it must be deleted before being replaced
+    // with a ConfigMap of the same name.
```
lukehoban: I'd focus more on "preserving the same semantics as Helm" here and simplify the comment.

pack/nodejs/helm.ts
```diff
@@ -0,0 +1,715 @@
...
+    export class Chart extends pulumi.ComponentResource {
+        public readonly resources: pulumi.CustomResource[];
+
+        constructor(releaseName: string, config: ChartOpts, opts?: pulumi.ResourceOptions) {
+            super("kubernetes:helm.sh/v2:Chart", releaseName, config, opts);
```
lukehoban: Why `helm.sh` here?

pack/nodejs/helm.ts
```diff
@@ -0,0 +1,715 @@
...
+        ...opts.keyring != null     ? [`--keyring ${shell.quote([opts.keyring])}`]         : [],
+        ...opts.password != null    ? [`--password ${shell.quote([opts.password])}`]       : [],
+        ...opts.repo != null        ? [`--repo ${shell.quote([opts.repo])}`]               : [],
+        ...opts.untardir != null    ? [`--untardir ${shell.quote([opts.untardir])}`]       : [],
+        ...opts.username != null    ? [`--username ${shell.quote([opts.username])}`]       : [],
+        ...opts.home != null        ? [`--home ${shell.quote([opts.home])}`]               : [],
+        ...opts.devel == true       ? [`--devel`]                                          : [],
+        ...opts.prov == true        ? [`--prov`]                                           : [],
+        ...opts.verify == true      ? [`--verify`]                                         : [],
+    ] : [];
```
lukehoban: This is pretty strange code :-).  Can this just be a bunch of simple `if`s?

Also:
1. `!==` and `===`
2. `undefined` instead of `null`

examples/helm/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  },
+  transformations: [
+    // Don't expose any services publicly, i.e., turn all services into ClusterIP instead of
+    // LoadBalancer. (This cuts time to run CI tests significantly.)
+    helm.transform.ifGvkEqual("v1", "Service", (svc: any) => {
```
lukehoban: Let's get rid of `helm.transform`, and just write the `if` inline here.  It'll be *a lot* clearer what's going on here and how to do other similar things.

pack/nodejs/helm.ts
```diff
@@ -0,0 +1,715 @@
...
+    execSync(`helm fetch ${chart} ${flags.join(" ")}`);
+}
+
+export namespace transform {
```
lukehoban: I think we can get rid of this namespace.  It shouldn't be part of a Helm package here.  And I think generally these helpers (if even really needed) can be in another layer.

pkg/gen/node-templates/package.json.mustache
```diff
@@ -13,7 +13,13 @@
...
     },
     "dependencies": {
-        "@pulumi/pulumi": "^0.14.0-rc1"
+        "@pulumi/pulumi": "^0.14.0-rc1",
+        "@types/js-yaml": "^3.11.2",
+        "js-yaml": "^3.12.0",
+        "shell-quote": "^1.6.1",
+        "@types/shell-quote": "^1.6.0",
+        "tmp": "^0.0.33",
+        "@types/tmp": "^0.0.33"
```
lukehoban: Minor - but all the `@types` can go in `devDependencies`.

pack/nodejs/helm.ts
```diff
@@ -0,0 +1,715 @@
...
+            const overrides = tmp.fileSync({postfix: ".yaml"});
+            const chartDir = tmp.dirSync({unsafeCleanup: true});
+
+            try {
+                // Fetch chart.
+                fetch(`${config.repo}/${config.chart}`,
+                    {destination: chartDir.name, version: config.version});
+
+                // Create temporary overrides file.
+                const overrides = tmp.fileSync({postfix: ".yaml"});
```
lukehoban: Is this duplicated from above?  Looks like you are creating it twice (and not cleaning this one up).

pack/nodejs/helm.ts
```diff
@@ -0,0 +1,715 @@
...
+
+                // Does not require Tiller. From the `helm template` documentation:
+                //
+                // >  Render chart templates locally and display the output.
+                // >
+                // > This does not require Tiller. However, any values that would normally be
+                // > looked up or retrieved in-cluster will be faked locally. Additionally, none
+                // > of the server-side testing of chart validity (e.g. whether an API is supported)
+                // > is done.
+                const chart = `${chartDir.name}/${config.chart}`;
```
lukehoban: Are these names guaranteed to be shell-safe - or should these also be `shell.quote`d?

examples/helm/index.ts
hausdorff: Parsing the repo name out of that ID is hard in general. This ID can be 5 things (URL, path, well-known ID, etc.)

examples/helm/index.ts
hausdorff: This is how you "unset" values in `values.yml` as far as I can tell. This isn't Pulumi-specific.

pack/nodejs/helm.ts
```diff
@@ -0,0 +1,715 @@
...
+                const release = shell.quote([releaseName]);
+                const values = shell.quote([overrides.name]);
+                const yamlStream = execSync(
+                    `helm template ${chart} --name ${release} --values ${values}`
+                ).toString();
+                const resourcesObjects = yaml.safeLoadAll(yamlStream);
+                this.resources = fromList(resourcesObjects, config.transformations || [], { ...opts, parent: this });
```
lukehoban: Is there value in exposing this as a map from "qualified name" to resource instead of just an array of resources?  That might make it easier for example to look up some specific resource (the exposed `Service`) and pull outputs off of it.

examples/helm/index.ts
```diff
@@ -0,0 +1,24 @@
...
+      if (svc.spec && svc.spec.type && svc.spec.type == "LoadBalancer") {
+        svc.spec.type = "ClusterIP";
+      }
+    }),
+  ]
+});
```
lukehoban: In general, I love it when these examples show off how you expose the externally addressable endpoint from the stack.  That helps complete the scenario ("deploy this then curl the endpoint it returns").  Can we do that here?  It also helps show off accessing outputs of constituent resources.

pack/nodejs/helm.ts
hausdorff: Because Kubernetes resources are organized by group/version/kind. The group is `helm.sh`, the API version is `v2`, and the kind is `Chart`. By convention `kubernetes:v2:Chart` refers to `core/v2/Chart`, which is confusing.

examples/helm/index.ts
lukehoban: Why do we need to parse it out?   The only place I see it being used these two pieces get concatenated together.

pack/nodejs/helm.ts
hausdorff: I am pretty confident everyone who is doing any kind of transformation on any Chart at all will end up using one of these three utility functions. To me this is a compelling reason to keep it in.

examples/helm/index.ts
lukehoban: Got it.  https://github.com/helm/helm/blob/master/docs/chart_template_guide/values_files.md#deleting-a-default-key

pack/nodejs/helm.ts
hausdorff: Is the problem that it's not readable?

pack/nodejs/helm.ts
lukehoban: Question really was why is it `helm.sh` here but `helm` in the module name.  (The latter of course can't have a `.`, but assuming this `helm.sh` name isn't derived from an actual Kubernetes object, why not just `helm`?

pack/nodejs/helm.ts
hausdorff: Names are supposed to be DNS-1123 compliant, but on second thought they should be escaped because people can put bad things here.

pack/nodejs/helm.ts
hausdorff: Copy+paste error.

pack/nodejs/helm.ts
lukehoban: I feel like making it clearer what the primitive is, even if it results in a little bit of boiler plate, is going to make this a lot more consumable than defining/using combinators here.  

In particular, the one used in the current sample above would be quite a bit easier to read if it was just a callback that had an `if` in it's body.

examples/helm/index.ts
hausdorff: When we move a similar example to `pulumi/examples`, yes, but we can't expose it publicly in CI until we stop using minikube, right?

lukehoban: Possibly true - though since you are forcing to ClusterIP for now, can you provide a URL based on that?  (Not an export on how ClusterIP-based Service endpoints work...).

pack/nodejs/helm.ts
hausdorff: I am not sure there is, because:

1. we'd have to create our own way of addressing them (_e.g._, `${group}/${version}/${kind}/${namespace}/${name}`) which would be opaque and unfamiliar to the user, and opaque in the sense that we'd have to decide what to do about exceptions like the `core` namespace.
1. `${name}` itself is pretty unpredictable, since it can be literally anything.

In general instead of having a bag of properties, I think it's useful to expose nice filtering primitives in the `transform` namespace.

pack/nodejs/helm.ts
hausdorff: Because groups tend to be identified by the domain that the organization that "owns" that group owns. For example, `pulumi.com` vs `pulumi`, `ksonnet.io` vs `ksonnet`, and so on.

pack/nodejs/helm.ts
hausdorff: But the issue is that the `update` semantics _don't_ match Helm's. Helm will update the configmap in place, but _not_ trigger a rolling update of deployments.

hausdorff: Oh actually, you're right. These are the same semantics as helm. We delete and then replace.

pack/nodejs/helm.ts
hausdorff: I think we've agreed to keep this as is.

examples/helm/index.ts
hausdorff: I think we've agreed offline to keep this as is.

hausdorff:
To recap, @lukehoban and agreed offline to:

* Leave the `transform.*` off the table for now.
* Make `Chart.resources` a map ID -> `CustomResource`.
* General clean ups and stuff.

With these fixes in, I think it's good enough that we can merge when CI passes.

pgavlin:
> Make Chart.resources a map ID -> CustomResource.

Where does the ID come from here?

hausdorff:
@pgavlin It's basically `${apiVersion}/${kind}::${namespace}/${name}`. This is not perfect, but within a chart it should be enough to distinguish it from other resources.

hausdorff:
We've merged for now, if we want to follow up with other issues, please make sure to ping me about them.

PR #112, opened 2018-07-29
Implement status-rich updates

hausdorff:
This PR principally contains the logic for status-rich updates of the key "complex" Kubernetes resources (_i.e._, `Pod`, `Service`, and `Deployment`), as well as fixes for a set of bugs that affect this change.

The changes related to the status-rich updates themselves essentially have two parts:

* **A series of playback tests.** We used CarbonQL to observe changes to these key resources under various circumstances, and then encoded those as tests "playing back" those events. This was designed to (1) give a transparent view of what these "complex" awaiters do, and (2) give us a rigorous backbone on top of which to ensure the correctness of this code.
* **A set of three "complex awaiter" objects.** These objects are essentially structs that are each wrapped around an event loop that is continuously observing events from several streams (_i.e._, updates to one or more resources from the API server, plus channels each for timeout and cancellation). As events are received, these objects update the known state of the resources so that helpful intermediate errors can be displayed.

The rest of the changes are ancillary fixes to make this experience good. Most of these are small bugs.

### DIFF
--- a/examples/examples_test.go
+++ b/examples/examples_test.go
@@ -86,83 +86,88 @@ func TestExamples(t *testing.T) {
 					assert.Equal(t, resource.RootStackType, stackRes.URN.Type())
 				},
 			}),
-			base.With(integration.ProgramTestOptions{
-				Dir: path.Join(cwd, "guestbook"),
-				ExtraRuntimeValidation: func(
-					t *testing.T, stackInfo integration.RuntimeValidationStackInfo,
-				) {
-					assert.NotNil(t, stackInfo.Deployment)
-					assert.Equal(t, 7, len(stackInfo.Deployment.Resources))
+			// TODO[pulumi-kubernetes#117]: Enable this when parallelism is turned on.
+			//
+			// base.With(integration.ProgramTestOptions{
+			// 	Dir: path.Join(cwd, "guestbook"),
+			// 	ExtraRuntimeValidation: func(
+			// 		t *testing.T, stackInfo integration.RuntimeValidationStackInfo,
+			// 	) {
+			// 		assert.NotNil(t, stackInfo.Deployment)
+			// 		assert.Equal(t, 7, len(stackInfo.Deployment.Resources))
 
-					sort.Slice(stackInfo.Deployment.Resources, func(i, j int) bool {
-						ri := stackInfo.Deployment.Resources[i]
-						rj := stackInfo.Deployment.Resources[j]
-						riname, _ := openapi.Pluck(ri.Outputs, "live", "metadata", "name")
-						rinamespace, _ := openapi.Pluck(ri.Outputs, "live", "metadata", "namespace")
-						rjname, _ := openapi.Pluck(rj.Outputs, "live", "metadata", "name")
-						rjnamespace, _ := openapi.Pluck(rj.Outputs, "live", "metadata", "namespace")
-						return fmt.Sprintf("%s/%s/%s", ri.URN.Type(), rinamespace, riname) <
-							fmt.Sprintf("%s/%s/%s", rj.URN.Type(), rjnamespace, rjname)
-					})
+			// 		sort.Slice(stackInfo.Deployment.Resources, func(i, j int) bool {
+			// 			ri := stackInfo.Deployment.Resources[i]
+			// 			rj := stackInfo.Deployment.Resources[j]
+			// 			riname, _ := openapi.Pluck(ri.Outputs, "live", "metadata", "name")
+			// 			rinamespace, _ := openapi.Pluck(ri.Outputs, "live", "metadata", "namespace")
+			// 			rjname, _ := openapi.Pluck(rj.Outputs, "live", "metadata", "name")
+			// 			rjnamespace, _ := openapi.Pluck(rj.Outputs, "live", "metadata", "namespace")
+			// 			return fmt.Sprintf("%s/%s/%s", ri.URN.Type(), rinamespace, riname) <
+			// 				fmt.Sprintf("%s/%s/%s", rj.URN.Type(), rjnamespace, rjname)
+			// 		})
 
-					var name interface{}
-					var status interface{}
+			// 		var name interface{}
+			// 		var status interface{}
 
-					// Verify frontend deployment.
-					frontendDepl := stackInfo.Deployment.Resources[0]
-					assert.Equal(t, tokens.Type("kubernetes:apps/v1:Deployment"), frontendDepl.URN.Type())
-					name, _ = openapi.Pluck(frontendDepl.Outputs, "live", "metadata", "name")
-					assert.Equal(t, "frontend", name)
-					status, _ = openapi.Pluck(frontendDepl.Outputs, "live", "status", "readyReplicas")
-					assert.Equal(t, float64(3), status)
+			// 		// Verify frontend deployment.
+			// 		frontendDepl := stackInfo.Deployment.Resources[0]
+			// 		assert.Equal(t, tokens.Type("kubernetes:apps/v1:Deployment"), frontendDepl.URN.Type())
+			// 		name, _ = openapi.Pluck(frontendDepl.Outputs, "live", "metadata", "name")
+			// 		assert.Equal(t, "frontend", name)
+			// 		status, _ = openapi.Pluck(frontendDepl.Outputs, "live", "status", "readyReplicas")
+			// 		assert.Equal(t, float64(3), status)
 
-					// Verify redis-master deployment.
-					redisMasterDepl := stackInfo.Deployment.Resources[1]
-					assert.Equal(t, tokens.Type("kubernetes:apps/v1:Deployment"), redisMasterDepl.URN.Type())
-					name, _ = openapi.Pluck(redisMasterDepl.Outputs, "live", "metadata", "name")
-					assert.Equal(t, "redis-master", name)
-					status, _ = openapi.Pluck(redisMasterDepl.Outputs, "live", "status", "readyReplicas")
-					assert.Equal(t, float64(1), status)
+			// 		// Verify redis-master deployment.
+			// 		redisMasterDepl := stackInfo.Deployment.Resources[1]
+			// 		assert.Equal(t, tokens.Type("kubernetes:apps/v1:Deployment"), redisMasterDepl.URN.Type())
+			// 		name, _ = openapi.Pluck(redisMasterDepl.Outputs, "live", "metadata", "name")
+			// 		assert.Equal(t, "redis-master", name)
+			// 		status, _ = openapi.Pluck(redisMasterDepl.Outputs, "live", "status", "readyReplicas")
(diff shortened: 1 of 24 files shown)


pkg/await/core_pod.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// Check if the pod is ready.
+	switch phase := status["phase"]; phase {
+	case "Running":
+		// Success.
+		pc.podReady = true
+		return
+	case "Pending":
+		pc.checkPendingPod(pod, status)
+	// case "Succeeded", "Failed", "Unknown":
```
swgillespie: was this intentional?

pkg/await/core_pod.go
```diff
@@ -0,0 +1,293 @@
...
+		name, _ = rawName.(string)
+
+		rawWaiting, isWaiting := openapi.Pluck(containerStatus, "state", "waiting")
+		waiting, isMap := rawWaiting.(map[string]interface{})
+		if isWaiting && rawWaiting != nil && isMap {
+			pc.checkWaitingContainer(name, waiting)
+		}
+
+		//
+		// TODO: terminated
```
swgillespie: we should open an issue tracking this todo IMO

pkg/await/core_pod.go
```diff
@@ -0,0 +1,293 @@
...
+
+	rawMessage, hasMessage := waiting["message"]
+	message, isString := rawMessage.(string)
+	if !hasMessage || !isString {
+		return
+	}
+
+	// Image pull error has a bunch of useless junk at the beginning of the error message. Try to
+	// remove it.
+	imagePullJunk := "rpc error: code = Unknown desc = Error response from daemon: "
```
swgillespie: o_o

pkg/await/core_pod.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+func (pia *podInitAwaiter) Await() error {
+	//
+	// We succeed when `.status.phase` is set to "Running".
+	//
+
+	podWatcher, err := pia.config.clientForResource.Watch(metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "Could set up watch for Pod object '%s'",
```
swgillespie: couldn't?

pkg/await/core_pod.go
```diff
@@ -0,0 +1,293 @@
...
+	defer podWatcher.Stop()
+
+	// Create timeout channel.
+	timeoutCh := make(chan struct{})
+	go func() {
+		time.Sleep(5 * time.Minute)
+		timeoutCh <- struct{}{}
+	}()
+
+	return pia.await(podWatcher, timeoutCh)
```
swgillespie: `pia.await(podWatcher, time.Tick(5 * time.Minute))` should work too

swgillespie: I see you use this pattern a bunch of places in here, would it be obtrusive to make timeoutCh a `chan Time` instead of a `chan struct{}`?

hausdorff: Let's do that, but the linter informs me that we want to use `time.NewTicker` instead, or else we leak the the ticker.

pkg/await/apps_deployment.go
```diff
@@ -0,0 +1,387 @@
...
+	//   2. `.status.conditions` has a status of type `Progressing`, whose `status` member is set
+	//      to `True`, and whose `reason` is `NewReplicaSetAvailable`.
+	//   3. `.status.conditions` has a status of type `Available` whose `status` member is set to
+	//      `True`.
+	//
+
+	// Create Deployment watcher.
+	deploymentWatcher, err := dia.config.clientForResource.Watch(metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "Could set up watch for Deployment object '%s'",
```
pgavlin: Nit: `Could not...`

pkg/await/apps_deployment.go
```diff
@@ -0,0 +1,387 @@
...
+		return errors.Wrapf(err,
+			"Could create watcher for Pods objects associated with Deployment '%s'",
+			dia.config.currentInputs.GetName())
+	}
+	defer podWatcher.Stop()
+
+	period := make(chan struct{})
+	go func() {
+		for {
+			time.Sleep(10 * time.Second)
```
pgavlin: Can we hook in completion/cancellation to this s.t. it doesn't run forever?

pkg/await/apps_deployment.go
```diff
@@ -0,0 +1,385 @@
...
+	pod, isUnstructured := event.Object.(*unstructured.Unstructured)
+	if !isUnstructured {
+		glog.V(3).Infof("Pod watch received unknown object type '%s'",
+			reflect.TypeOf(pod))
+		return
+	}
+
+	// Check whether this Pod was created by our Deployment.
+	podName := pod.GetName()
+	if !strings.HasPrefix(podName+"-", inputDeploymentName) {
```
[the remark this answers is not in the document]
hausdorff: Oh, shoot, I forgot to do that. Yes, I meant to hook up the `.metadata.ownerReference`.

pkg/await/apps_deployment.go
```diff
@@ -0,0 +1,385 @@
...
+			dia.processDeploymentEvent(event)
+		case event := <-podWatcher.ResultChan():
+			dia.processPodEvent(event)
+		}
+	}
+}
+
+func (dia *deploymentInitAwaiter) processDeploymentEvent(event watch.Event) {
+	// Start with a blank slate.
+	dia.rolloutComplete = false
```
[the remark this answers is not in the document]
hausdorff: Since there are so few places where we determine the rollout to be complete, it's easier to mark it `false` and then prove that wrong later.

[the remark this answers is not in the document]
hausdorff: Oh, right, this should happen after the name check though.

pkg/await/apps_deployment.go
```diff
@@ -0,0 +1,385 @@
...
+	if !hasGeneration || !isNumber {
+		// No observed generation, Deployment controller has not yet created a ReplicaSet. Do
+		// nothing.
+		dia.observedGeneration = 0
+		return
+	}
+
+	// Return if Deployment controller has not created a new ReplicaSet for `currentGeneration`.
+	if dia.observedGeneration != dia.currentGeneration {
+		return
```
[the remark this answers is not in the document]
hausdorff: If we don't get to the `currentGeneration`, we should fail. A place this could happen is if they run `kubectl rollback`, in which case `.metadata.generation` would be set to the previous generation, and `observedGeneration` would need to be _decremented_.

pkg/await/apps_deployment.go
```diff
@@ -0,0 +1,385 @@
...
+				dia.deploymentErrors[reason] = message
+			}
+		}
+	}
+
+	dia.rolloutComplete = replicaSetAvailable && deploymentAvailable
+}
+
+func (dia *deploymentInitAwaiter) processPodEvent(event watch.Event) {
+	// Start with a blank slate.
```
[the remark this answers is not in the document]
hausdorff: The pod errors are completely advisory, so no need to blank this out. We just need to keep track of each pod and check them in batch every couple seconds so we can report to the user.

pkg/await/apps_deployment.go
```diff
@@ -0,0 +1,385 @@
...
+//
+// A Deployment is a construct that allows users to specify how to execute an update to an
+// application that is replicated some number of times in a cluster. When an application is updated,
+// the Deployment will incrementally roll out the new version (according to the policy requested by
+// the user). When the new application Pods becomes "live" (as specified by the liveness and
+// readiness probes), the old Pods are killed and deleted.
+//
+// Because this resource abstracts over so much (it is a way to roll out, essentially, ReplicaSets,
+// which themseves are an abstraction for replicating Pods), the success conditions are fairly
+// complex:
```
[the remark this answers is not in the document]
hausdorff: Yeah, kind of: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment

I would not rate these docs as excellent, but they are definitely better than nothing.

pkg/await/apps_deployment.go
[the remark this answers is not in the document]
hausdorff: Though, technically, none of this matters. We will return pretty much as soon as we meet success conditions, so this should be a no-op.

examples/examples_test.go
```diff
@@ -86,83 +86,86 @@ func TestExamples(t *testing.T) {
...
+			// 		status, _ = openapi.Pluck(redisSlaveService.Outputs, "live", "spec", "clusterIP")
+			// 		assert.True(t, len(status.(string)) > 1)
+
+			// 		// Verify root resource.
+			// 		stackRes := stackInfo.Deployment.Resources[6]
+			// 		assert.Equal(t, resource.RootStackType, stackRes.URN.Type())
+			// 	},
+			// }),
+
+			// TODO(hausdorff): Enable this when parallelism is turned on.
```
lukehoban: In general we've been aiming to track TODO with issues and link them in code.  That might be nice to do here and link t existing parallelism issues.

pkg/await/core_service_test.go
```diff
@@ -0,0 +1,290 @@
...
+		{
+			description: "Should succeed if Endpoints have settled when timeout occurs",
+			do: func(services, endpoints chan watch.Event, settled chan struct{}, timeout chan time.Time) {
+				// API server passes initialized service back.
+				services <- watchAddedEvent(initializedService("default", "foo-4setj4y6"))
+
+				// Pass initialized endpoint objects.
+				endpoints <- watchAddedEvent(initializedEndpoint("default", "foo-4setj4y6"))
+
+				// DO NOT mark as settled. Instead, time out.
```
lukehoban: This doesn't (to me) match whats in the description.

pkg/await/core_service_test.go
```diff
@@ -0,0 +1,290 @@
...
+		{
+			description: "Should succeed when Service is allocated an IP address and Endpoints target a Pod",
+			do: func(services, endpoints chan watch.Event, settled chan struct{}, timeout chan time.Time) {
+				// API server passes initialized service and endpoint objects back.
+				services <- watchAddedEvent(initializedService("default", "foo-4setj4y6"))
+				endpoints <- watchAddedEvent(initializedEndpoint("default", "foo-4setj4y6"))
+
+				// Mark endpoint objects as having settled. Success.
+				settled <- struct{}{}
+			},
```
lukehoban: I like the tests here.  These feel like they will be easy to maintainable and add to as we address bugs and adapt to changes in Kubernetes behaviour + provide good regression prevention.  Thanks for adding these in a style that is reasonably easy to reason about!

pkg/await/core_service_test.go
hausdorff: This is correct. We're not passing `settled <- struct{}{}`. I'll reword though.

PR #118, opened 2018-08-06
Small set of fixes for corner cases in await logic

hausdorff:


hausdorff:
Per out-of-band discussion, I realize now that I still have a lot more context than everyone, so changes like this necessitate significant documentation. So I've added that here.

Issue #101, opened 2018-07-21
`Configure` should accept a `kubeconfig` spec as JSON


pgavlin:
e.g. I should be able to JSON-serialize a `kubeconfig` value and supply it as the value for `kubernetes:config:kubernetes`.

pgavlin:
This depends on #100.

hausdorff:
@pgavlin Somewhat relatedly, I think we should also consider making the literal `$KUBECONFIG` file an "input" (in some loose sense) to the Pulumi application. That is to say, if the user changes the active context to point at a new cluster (_e.g._, with something like `kubectl config use-context some-new-cluster`), then we'll want to report something different -- that none of the resources are available, and we will need to create new ones.

And related to that, I'm still not sure how to reason about multiple stacks on different clusters.

pgavlin:
Here's a prototype approach to the original issue: https://github.com/pulumi/pulumi-kubernetes/compare/pgavlin/kubeconfig

pgavlin:
Closed with #119.

PR #119, opened 2018-08-09
Accept a `kubeconfig` config option.

### LINKED ISSUE #100: configuration should happen in `Configure`
Currently configuration happens in `makeKubeProvider`; it should instead happen in `Configure`.

pgavlin:
This option is the literal text of a serialized `kubeconfig` value. If
this options is set, the config specified within is used in place of the
ambient config.

In addition to adding this option, these changes move the timing of
client creation, etc. to `Configure` rather than plugin creation.

Both of these changes are in service of instantiating a first-class
k8s provider that is configured based on the output of other resources
(e.g. an EKS cluster).

Fixes #100, #101.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1378,6 +1378,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "0b565c5bfab567d99e5729a9790dac6087444adb99418ccd9b0c424e0e61beef"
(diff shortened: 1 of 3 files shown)


hausdorff:
Per offline conversation with @pgavlin, the code looks like pretty much what we want, but we're going to update it to allow the user to configure `cluster`, `context`, and `namespace` as well, instead of only the kubeconfig file itself.

PR #121, opened 2018-08-09
Expose a first-class provider for k8s.

### LINKED ISSUE #120: Expose a first-class provider type.
ala `aws.Provider`.

pgavlin:
This resource can be used to create new instances of the Kubernetes
resource provider that target different clusters.

Fixes #120.

### DIFF
--- a/pack/nodejs/provider.ts
+++ b/pack/nodejs/provider.ts
@@ -5,6 +5,50 @@ import * as pulumi from "@pulumi/pulumi";
 import * as inputApi from "./types/input";
 import * as outputApi from "./types/output";
 
+/**
+ * The provider type for the kubernetes package.
+ */
+export class Provider extends pulumi.ProviderResource {
+    /**
+     * Create a Provider resource with the given unique name, arguments, and options.
+     *
+     * @param name The _unique_ name of the resource.
+     * @param args The arguments to use to populate this resource's properties.
+     * @param opts A bag of options that control this resource's behavior.
+     */
+    constructor(name: string, args: ProviderArgs, opts?: pulumi.ResourceOptions) {
+        let inputs: pulumi.Inputs = {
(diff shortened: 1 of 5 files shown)


pgavlin:
Testing this now using EKS

pack/nodejs/provider.ts
```diff
@@ -5,6 +5,50 @@ import * as pulumi from "@pulumi/pulumi";
...
+     */
+    readonly cluster?: pulumi.Input<string>;
+    /**
+     * If present, the name of the kubeconfig context to use.
+     */
+    readonly context?: pulumi.Input<string>;
+    /**
+     * The contents of a kubeconfig file. If this is set, this config will be used instead of $KUBECONFIG.
+     */
+    readonly kubeconfig?: pulumi.Input<string>;
```
pgavlin: It would be sort of nice if we could type this as `Pulumi.Input<Kubeconfig|string>` and automatically JSON-serialize in the former case. Thoughts?

hausdorff: That seems like a good idea.

pgavlin: Are the types in `k8s.io/client-go/tools/clientcmd/api` the best source for the canonical `kubeconfig` JSON schema?

hausdorff: Hmm, this _should_ be a part of the OpenAPI spec, and therefore we _should_ be generating it, but it's not, and we're not. So I'm not sure if we can generate it....

hausdorff: But yes, that is the correct place.

lukehoban: Suggest we don't block on that, keep this string for now, and open separate issue to track stronger typing (and generation of kubeconfig schema types).

pgavlin: :+1:

PR #122, opened 2018-08-10
Extend deployment API version canonicalization.

pgavlin:
In addition to "extensions/v1beta1", Deployment objects may have the
following versions, all of which we want to canonicalize to "apps/v1":
- "apps/v1beta1"
- "apps/v1beta2"


hausdorff:
Just to be clear, this is necessary because the `Deployment` controller will change the `apiVersion` of `Deployment` objects at will, and that may or may not line up with the `.metadata.ownerReferences` of the related `ReplicaSet`s, which might use a different `apiVersion`. So that's pretty cool.

PR #126, opened 2018-08-10
Enable parallelism tests

ellismg:
Pick up the very latest pulumi CLI and pulumi packages, which now have
parallelism in them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ install:
     - minikube update-context
 
     # Install Pulumi
-    - curl -L https://get.pulumi.com/ | bash -s -- --version 0.14.4-dev-1533786685-g80569fb5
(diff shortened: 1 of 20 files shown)


.travis.yml
```diff
@@ -36,7 +36,7 @@ install:
     - minikube update-context
 
     # Install Pulumi
-    - curl -L https://get.pulumi.com/ | bash -s -- --version 0.14.4-dev-1533786685-g80569fb5
+    - curl -L https://get.pulumi.com/ | bash -s -- --version 0.15.1-dev-1533943101-g58a75cbb
```
ellismg: Because we started to work on 0.15.0 yesterday, master of pulumi got tagged with 0.15.1.  This is just HEAD as of a few moments ago, and everything here will make it's way in to 0.15.0-rc2 and the final build.

PR #129, opened 2018-08-12
Accept CustomResourceOptions for K8s resources

joeduffy:
This change allows CustomResourceOptions for the core K8s resources, in
addition to Helm Charts. It strips out the provider property while
passing from the Helm Chart constructor to the super constructor,
because we reject the provider property on non-CustomResourceOptions.


joeduffy:
I'm still not able to get K8s resource provider instances to work, even with this fix. I'll keep debugging what's going on: `I0812 16:25:22.672234   71528 provider_plugin.go:197] Provider[kubernetes, 0xc420d76be0].Configure() failed: err=Unable to read kubectl config: invalid configuration: no configuration has been provided`.

joeduffy:
Oh, I'm also not sure what the right practice is for updating these dependencies. @ellismg Can you take a look and let me know? Checking in a dependency on our RC bits doesn't feel right to me, but it's replacing what we already checked in, which depends on a dev build, which feels even less right...

pack/nodejs/helm.ts
```diff
@@ -34,8 +34,9 @@ export namespace v2 {
     export class Chart extends pulumi.ComponentResource {
         public readonly resources: {[key: string]: pulumi.CustomResource};
 
-        constructor(releaseName: string, config: ChartOpts, opts?: pulumi.ResourceOptions) {
-            super("kubernetes:helm.sh/v2:Chart", releaseName, config, opts);
+        constructor(releaseName: string, config: ChartOpts, opts?: pulumi.CustomResourceOptions) {
```
lukehoban: This is a `ComponentResource`, so should be `pulumi.ComponentResourceOptions`.

I'll push a change to fix this, along with a test.

joeduffy: But the whole point is that I want to pass a provider that is used by the child resources .... Is the convention here different for components?

joeduffy: Oh I think this may be the automatic parent provider searching stuff you guys mentioned last week. If it works, great, certainly was not obvious to me.

lukehoban: Yes - all of this is related to how providers are handled in components (via parent).

You would do:

```javascript
let myk8s = new k8s.Provider("myk8s", { /* ... */ });
let chart = new helm.v2.Chart("simple-nginx", { /* ... */ }, { providers: { "kubernetes": myk8s } });
```

This is necessary, instead of `provider:` because components can create many resources from many providers.

Opened https://github.com/pulumi/docs/issues/516 to document this and other related concepts.

joeduffy:
LGTM

PR #131, opened 2018-08-13
Teach Pulumi to understand Kubernetes YAML

### LINKED ISSUE #115: Support deployment of Kubernetes YAML files
We added support for deploying Helm charts via Pulumi in https://github.com/pulumi/pulumi-kubernetes/issues/54, and should extend this to support the (simpler) case of deploying raw Kubernetes YAML files.

hausdorff:
Fixes #115.


pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -1,9 +1,110 @@
...
+ */
+export class Yaml {
+    public readonly resources: { [key: string]: pulumi.CustomResource };
+
+    constructor(config: YamlOpts, opts?: pulumi.CustomResourceOptions) {
+        const yamlTexts = [];
+
+        if (config.files !== undefined) {
+            let files: string[] = [];
+            if (Object.prototype.toString.call(config.files) === "[object String]") {
```
swgillespie: `typeof String`?

swgillespie: You get the added bonus of not needing the cast below.

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -1,9 +1,110 @@
...
+                }
+            }
+
+            for (const file of files) {
+                yamlTexts.push(fs.readFileSync(file).toString());
+            }
+        }
+
+        if (config.yaml !== undefined) {
+            if (Object.prototype.toString.call(config.files) === "[object String]") {
```
swgillespie: Same here

pgavlin: Was this meant to be a check on `config.yaml`?

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -1,9 +1,110 @@
...
+ *   2. Using a file pattern or a list of file patterns:
+ *        a. `{files: "*.yaml"}`
+ *        b. `{files: ["foo/*.yaml", "bar/*.yaml"]}`
+ *   3. Using a literal string containing YAML, or a list of such strings:
+ *        a. `{yaml: "(LITERAL YAML HERE)"}`
+ *        b. `{yaml: ["(LITERAL YAML HERE)", "(MORE YAML)"]}`
+ *   4. Any combination of files, patterns, or YAML strings:
+ *        a. `{files: "foo.yaml", yaml: "(LITERAL YAML HERE)"}`
+ */
+export class Yaml {
```
pgavlin: Can you go over the thinking behind making this a plain class rather than inheriting from `pulumi.ComponentResource`? I do think it makes sense--e.g. it allows the use of the type without introducing potentially-awkward nesting--but I'd like to hear your thoughts as well.

joeduffy: Definitely would go the component route so stuff gets parented nicely underneath each yaml template.

lukehoban: :+1: This should be a component, and should parent the resources it creates.

We do have to consider whether the Helm Chart component has as children its direct Kube resources or a set of these Yaml components (whcih themselves have kube resources as children).

pkg/gen/node-templates/provider.ts.mustache
hausdorff: We got this from StackOverflow, I'll just take your word for it.

pkg/gen/node-templates/provider.ts.mustache
hausdorff: I didn't think it added much value and we needed the YAML objects to be parented to the `helm.Chart` in that case. But it sounds like people like this idea so we can do that.

pack/nodejs/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@pulumi/kubernetes",
-    "version": "v0.15.1-dev-1533945810-gb5546fb-dirty",
```
pgavlin: Was this intentional?

examples/yaml-guestbook/index.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright 2016-2018, Pulumi Corporation.  All rights reserved.
+
+import * as k8s from "@pulumi/kubernetes";
+
+// Create resources from standard Kubernetes guestbook YAML example.
+const guestbook = new k8s.Yaml("guestbook", { files: "yaml/*.yaml" });
```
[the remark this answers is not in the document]
hausdorff: Two things:

1. Luke suggested the name, I don't like it either, but I couldn't think of anything better. I actually think that `Yaml` is a slightly better name than `yaml.Template`.
1. Is this enough of a problem to be a blocker for release?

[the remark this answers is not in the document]
ellismg: > Is this enough of a problem to be a blocker for release?

FWIW, it would kind of suck to immediately change the name of the thing in the next release, breaking anyone who used it (or immediately deprecating the name, which is what we'd probably do).

hausdorff:
@joeduffy asked for a more complex parenting relationship. Basically, it involves bucketing resources by the YAML file they were found in. For example, in the Guestbook, all of the resources are in `guestbook.yaml`:

https://asciinema.org/a/Fjnwm3FbJM77GAEyuF9eoWYMW?speed=2

pack/nodejs/provider.ts
```diff
@@ -1,9 +1,1627 @@
...
+                case "batch/v1beta1/CronJob":
+                    return <T>this.resources[`${groupVersionKind}::${id}`];;
+                case "batch/v1beta1/CronJobList":
+                    return <T>this.resources[`${groupVersionKind}::${id}`];;
+                case "batch/v2alpha1/CronJob":
+                    return <T>this.resources[`${groupVersionKind}::${id}`];;
+                case "batch/v2alpha1/CronJobList":
+                    return <T>this.resources[`${groupVersionKind}::${id}`];;
+                case "certificates.k8s.io/v1beta1/CertificateSigningRequest":
+                    return <T>this.resources[`${groupVersionKind}::${id}`];;
```
lukehoban: Double `;;`

pack/nodejs/index.ts
```diff
@@ -1 +1,3 @@
 export * from "./provider";
+import * as helm from "./helm";
+export { helm };
```
pgavlin: 🎉

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -1,9 +1,196 @@
...
+import * as fs from "fs";
+
 import * as pulumi from "@pulumi/pulumi";
 import * as inputApi from "./types/input";
 import * as outputApi from "./types/output";
+import * as jsyaml from "js-yaml";
+import * as glob from "glob";
+
+export namespace yaml {
+    export interface ConfigGroupOpts {
```
pgavlin: Can we add docs for these?

pgavlin: (fine to do so after the fact)

PR #138, opened 2018-08-15
 Move live object properties to the root of resource state. 

### LINKED ISSUE #137: Outputs not working
Output properties on k8s resources are not being populated. This is because the values of these properties do not live at the root level in the resource's state object; instead, they are nested under "live".

A simple fix here is to move the live state that contains the values of the output properties to the root level of the resource's state object (i.e. instead of storing `{ inputs: {}, live: {} }` in the checkpoint file, store `{ ...live, "__inputs": {} }`).

pgavlin:
We currently next the live properties of a Kubernetes resource under
the key "live" when storing the resource's state in the checkpoint file.
This prevents output properties from resolving when using Node.JS (and
most likely Python as well), as the Pulumi JS library only looks for
property values at the root of an object's state.

The simplest, most localized fix we can make for this is to move a
Kubernetes resource's live properties into the root of the resource's
state. This commit implements that approach.

Fixes #137.

### DIFF
--- a/examples/examples_test.go
+++ b/examples/examples_test.go
@@ -56,10 +56,10 @@ func TestExamples(t *testing.T) {
 					sort.Slice(stackInfo.Deployment.Resources, func(i, j int) bool {
 						ri := stackInfo.Deployment.Resources[i]
 						rj := stackInfo.Deployment.Resources[j]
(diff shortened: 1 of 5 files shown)


pgavlin:
Hm. Looks like some of the integration tests are broken... Investigating.

PR #139, opened 2018-08-16
Don't dirty the working tree when building

ellismg:
For the package.json we generate, don't embed the version string right
away. Instead, just wack the version in the one we place in the bin
folder. This is in line with how we handle our other projects.

To lock in these fixes, add similar logic to CI to ensure the worktree
is clean after a PR runs and for a push job, before we publish.

### DIFF
--- a/pkg/gen/node-templates/package.json.mustache
+++ b/pkg/gen/node-templates/package.json.mustache
@@ -1,6 +1,6 @@
 {
     "name": "@pulumi/kubernetes",
-    "version": "{{ProviderVersion}}",
+    "version": "${VERSION}",
     "description": "A Pulumi package for creating and managing Kubernetes resources.",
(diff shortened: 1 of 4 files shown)


pkg/gen/node-templates/package.json.mustache
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@pulumi/kubernetes",
-    "version": "{{ProviderVersion}}",
+    "version": "${VERSION}",
```
ellismg: This file needn't be a template anymore, so if you would prefer I could just check it in as is to `pack/nodejs/package.json`.  Whatever is your preference.

ellismg:
@hausdorff Let me know what you think.  If you're happy with this, I will rebase to fix the merge conflict that happens because every change touches `package.json` without this change.

PR #140, opened 2018-08-16
Accept ComponentResourceOptions in ConfigGroup.

pgavlin:
This is necessary for provider plumbing to work correctly.

### DIFF
--- a/pkg/gen/node-templates/provider.ts.mustache
+++ b/pkg/gen/node-templates/provider.ts.mustache
@@ -70,7 +70,7 @@ export namespace yaml {
         resources: { [key: string]: pulumi.CustomResource };
 
         constructor(
-            resourceType: string, name: string, config: any, opts?: pulumi.CustomResourceOptions,
+            resourceType: string, name: string, config: any, opts?: pulumi.ComponentResourceOptions,
         ) {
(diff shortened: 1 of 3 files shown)


pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -124,14 +124,14 @@ export namespace yaml {
...
-        constructor(name: string, config: ConfigGroupOpts, opts?: pulumi.CustomResourceOptions) {
+        constructor(name: string, config: ConfigGroupOpts, opts?: pulumi.ComponentResourceOptions) {
             super("kubernetes:yaml:ConfigGroup", name, config, opts);
             this.resources = parse(config, {parent: this});
         }
     }
 
     export class ConfigFile extends CollectionComponentResource {
-        constructor(name: string, config: ConfigOpts, opts?: pulumi.CustomResourceOptions) {
+        constructor(name: string, config: ConfigOpts, opts?: pulumi.ComponentResourceOptions) {
```
hausdorff: Hmm, why does this show up as changed in the .mustache but not in the .ts file? Am I missing something?

pgavlin: Fixed.

PR #149, opened 2018-08-22
BREAKING: Change resource naming scheme, fix a small set of bugs

### LINKED ISSUE #146: Presence of CRDs cause deletes to fail

hausdorff:
BREAKING CHANGE: This PR fixes a small set of bugs and changes the way resources are named in the Pulumi Kubernetes provider. Quoting from the relevant commit:

> Use `/` in fully-qualified resource names
>
> Fixes #146.
> 
> Currently we're creating canonical IDs for resources using the `.`
> character. These IDs take the form of `${namespace}.${name}`, or simply
> `${name}` if namespace does not exist.
> 
> We copied this strategy from ksonnet, which in turn copied it from some
> part of kubectl, but this turns out to be a critical flaw, because `.`
> is an allowed character in DNS-1123-compliant names. This means it's
> possible to name your resource `foo.bar`, and if it does not contain a
> namespace, then we'd decide the namespace was `foo`.
> 
> The solution is to move canonical IDs for resources to use `/` to
> delimit namespace, which is much better because `/` is not DNS-1123
> compliant, and therefore no Kubernetes resource can have the slash in
> it.

This change is absolutely necessary, but it will be painful for anyone who (1) has a stack with explicitly-namespaced resources, or (2) uses `.` in the name or namespace of a resource.

Our plan is to:

- Include this in v0.15.2
- Find all users who are affected by the change (we have the appropriate SQL query written)
- Email all of them to let them know what will happen when they upgrade
- In my personal time, I will write a migration tool so they can do something like `pulumi stack export | migrate.sh | pulumi stack import`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ build:: $(OPENAPI_FILE)
 		yarn install && \
 		yarn run tsc
 	cp README.md LICENSE ${PACKDIR}/nodejs/package.json ${PACKDIR}/nodejs/yarn.lock ${PACKDIR}/nodejs/bin/
-	sed -i 's/$${VERSION}/$(VERSION)/g' ${PACKDIR}/nodejs/bin/package.json
+	sed -i.bak 's/$${VERSION}/$(VERSION)/g' ${PACKDIR}/nodejs/bin/package.json
 
 lint::
 	$(GOMETALINTER) ./cmd/... ./pkg/... | sort ; exit "$${PIPESTATUS[0]}"
--- a/Makefile
+++ b/Makefile
@@ -49,12 +49,16 @@ install::
 	mkdir -p "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)"
 	cp -r pack/nodejs/bin/. "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)"
 	rm -rf "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)/node_modules"
+	rm -rf "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)/tests"
 	cd "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)" && \
 		yarn install --offline --production && \
 		(yarn unlink > /dev/null 2>&1 || true) && \
 		yarn link
 
-test_all::
+test_fast::
+	./pack/nodejs/node_modules/mocha/bin/mocha ./pack/nodejs/bin/tests
+
+test_all:: test_fast
 	PATH=$(PULUMI_BIN):$(PATH) $(GO) test -v -cover -timeout 1h -parallel ${TESTPARALLELISM} $(TESTABLE_PKGS)
 
 .PHONY: publish_tgz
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -38,7 +38,7 @@ build:: $(OPENAPI_FILE)
 		yarn install && \
 		yarn run tsc
 	cp README.md LICENSE ${PACKDIR}/nodejs/package.json ${PACKDIR}/nodejs/yarn.lock ${PACKDIR}/nodejs/bin/
-	sed -i 's/$${VERSION}/$(VERSION)/g' ${PACKDIR}/nodejs/bin/package.json
+	sed -i.bak 's/$${VERSION}/$(VERSION)/g' ${PACKDIR}/nodejs/bin/package.json
```
lukehoban: ?

pkg/await/apps_deployment.go
```diff
@@ -455,8 +455,32 @@ func (dia *deploymentInitAwaiter) checkReplicaSetStatus() {
 	glog.V(3).Infof("ReplicaSet '%s' requests '%v' replicas, but has '%v' ready",
 		rs.GetName(), specReplicas, readyReplicas)
 
-	dia.updatedReplicaSetReady = lastGeneration != dia.currentGeneration && udpatedReplicaSetCreated &&
-		readyReplicasExists && readyReplicas >= int64(specReplicas)
+	if dia.changeTriggeredRollout() {
+		dia.updatedReplicaSetReady = lastGeneration != dia.currentGeneration && udpatedReplicaSetCreated &&
+			readyReplicasExists && readyReplicas >= int64(specReplicas)
+	} else {
+		dia.updatedReplicaSetReady = lastGeneration == dia.currentGeneration && udpatedReplicaSetCreated &&
```
lukehoban: Do we need to enforce that they are equal on this path?  In general, how should we think about being conservative about blocking success?  Seems we should aim to have as few things be capable of blocking success as possible, and be sure they are rock solid signals of a problem if they do?

Makefile
hausdorff: The title of this commit is:

> `make install` should be macOS-compatible

The macOS implementation of the `-i` flag requires the `.bak` suffix to work at all.

pkg/await/apps_deployment.go
hausdorff: Ok, we can do that.

pack/nodejs/helm.ts
```diff
@@ -65,6 +66,9 @@ export namespace v2 {
                     yaml: [yamlStream],
                     transformations: config.transformations || [],
                 }, { parent: this });
+            } catch (e) {
+                console.error(e.toString());
+                process.exit(1);
```
pgavlin: This is pretty surprising. An uncaught exception should already tear the process down. Is this not happening?

pgavlin: Let's change this to throw a `pulumi.RunError`.

hausdorff:
Ok, so to follow up, I did manually migrate all stacks deploying Kubernetes that are affected, and emailed them their new stacks. So anyone who is interested has a smooth transition.

PR #186, opened 2018-09-06
Call them "transformations" instead of "transforms"

hausdorff:


joeduffy:
I actually prefer transforms...

hausdorff:
Sync'd with @joeduffy ... recap: we all like `transforms` better, but the plan of record is to deprecate `transformations` and transition to `transform: (inputs: any[]) => any[]` close after launch.

PR #190, opened 2018-09-06
Collection of small bug fixes

hausdorff:


pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -166,12 +166,13 @@ export namespace yaml {
 
         const kind = obj["kind"];
         const apiVersion = obj["apiVersion"];
-        if (kind == null || apiVersion == null) {
+        const meta = obj["metadata"];
+        if (kind == null || apiVersion == null || meta == null) {
```
swgillespie: I think it's more accurate to do `"metadata" in obj` instead of doing the property lookup and testing for null. You'll get `undefined` anyway here if `metadata` isn't a property of `obj` - it works out here because of the non-strict equality, but `meta === null` will be false even if the property does not exist.

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -166,12 +166,13 @@ export namespace yaml {
...
             return null;
         }
 
-        let id: string = obj["metadata"]["name"];
-        const namespace = obj["metadata"]["namespace"] || undefined;
+        let id: string = meta["name"];
+        const namespace = meta["namespace"] || undefined;
```
swgillespie: I'm assuming the empty string is not a legal namespace?

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -166,12 +166,13 @@ export namespace yaml {
...
         const apiVersion = obj["apiVersion"];
-        if (kind == null || apiVersion == null) {
+        const meta = obj["metadata"];
+        if (kind == null || apiVersion == null || meta == null) {
             return null;
         }
 
-        let id: string = obj["metadata"]["name"];
-        const namespace = obj["metadata"]["namespace"] || undefined;
+        let id: string = meta["name"];
```
swgillespie: There's no guarantee that this is a string, right?

pkg/gen/node-templates/provider.ts.mustache
hausdorff: It is. It means `"default"`.

pkg/gen/node-templates/provider.ts.mustache
hausdorff: There is if it's a Kubernetes resource.

pkg/gen/node-templates/provider.ts.mustache
swgillespie: gotcha.

pkg/gen/node-templates/provider.ts.mustache
swgillespie: OK - not sure if we care about arbitrary input here, but I'm fine with it if you're fine.

pack/nodejs/helm.ts
```diff
@@ -68,7 +68,7 @@ export namespace v2 {
                 this.resources = k8s.yaml.parse({
                     yaml: [yamlStream],
                     transformations: config.transformations || [],
-                }, { parent: this });
+                }, { parent: this, provider: opts && opts.providers && opts.providers.kubernetes });
```
pgavlin: This should not be necessary. What are you hitting?

pkg/gen/node-templates/helm.ts.mustache
```diff
@@ -68,7 +68,7 @@ export namespace v2 {
                 this.resources = k8s.yaml.parse({
                     yaml: [yamlStream],
                     transformations: config.transformations || [],
-                }, { parent: this });
+                }, { parent: this, provider: opts && opts.providers && opts.providers.kubernetes });
```
pgavlin: Likewise here.

hausdorff:
Many hours later, we tracked down the "bug" I "fixed" relating to the provider not being reported through to the `Resource` constructor. The issue turns out to be that side-by-side installs of `@pulumi/pulumi` were incompatible with provider chains. This was fixed in https://github.com/pulumi/pulumi/pull/1901.

I hate computers. I am now retiring to become a goat farmer.

PR #192, opened 2018-09-06
Consume @pulumi/pulumi 0.15.1-rc1

ellismg:


ellismg:
After we land this, @hausdorff, I will cut 0.17.0-rc1 of this package.

PR #203, opened 2018-09-10
Small Helm-related bug fixes.

hausdorff:


pack/nodejs/helm.ts
```diff
@@ -78,6 +82,81 @@ export namespace v2 {
...
+            "Deployment",
+            "StatefulSet",
+            "Job",
+            "CronJob",
+            "Ingress",
+            "APIService"
+        ];
+
+        const ordering: { [key: string]: number } = {};
+        installOrder.forEach((kind, i) => {
```
pgavlin: Nit: `s/kind/_` or `s/installOrder[i]/kind` below.

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -155,21 +163,38 @@ export namespace yaml {
...
-        if (!("kind" in obj && "apiVersion" in obj && "metadata" in obj)) {
-            return null;
+        if (!("kind" in obj && "apiVersion" in obj)) {
+            throw new Error(`Kubernetes resources require a kind and apiVersion: ${JSON.stringify(obj)}`)
         }
 
         const kind = obj["kind"];
         const apiVersion = obj["apiVersion"];
+
+        // Flatten `v1.List`. `v1.List` is an undocumented Kubernetes resource, and does not appear
```
swgillespie: wat

pack/nodejs/helm.ts
```diff
@@ -61,14 +67,12 @@ export namespace v2 {
                 const chart = path.quotePath(nodepath.join(chartDir.name, config.chart));
                 const release = shell.quote([releaseName]);
                 const values = path.quotePath(overrides.name);
+                const defaultValues = path.quotePath(nodepath.join(chartDir.name, config.chart, "values.yaml"));
```
pgavlin: Is this file guaranteed to exist? What happens if it doesn't?

pack/nodejs/helm.ts
```diff
@@ -78,6 +82,81 @@ export namespace v2 {
...
             }
         }
+
+        parseTemplate(
+            yamlStream: string,
+            config: ChartOpts
+        ): { [key: string]: pulumi.CustomResource } {
+            const objs = jsyaml
+                .safeLoadAll(yamlStream)
+                .filter(a => a != null && "kind" in a)
```
pgavlin: What does Helm do for resources that are missing `kind`?

pack/nodejs/helm.ts
hausdorff: It's guaranteed to exist or it's not a valid chart.

pack/nodejs/helm.ts
hausdorff: There is no such thing as a resource that does not have a `kind`.

hausdorff: This is basically to filter out empty YAML objects, which Helm can produce lots of.

hausdorff:
@pgavlin has verbally signed off.

PR #210, opened 2018-09-13
Load kubeconfig file instead of parsing filename

### LINKED ISSUE #209: Problem parsing kubeconfig when passed via "kubernetes:kubeconfig"
If I set `$KUBECONFIG` in my environment, then pulumi can find my cluster and things work fine.

But if I set `kubernetes:kubeconfig` in my stack's yaml, I get an error:

```
Diagnostics:
  global: global
    error: failed to parse kubeconfig: couldn't get version/kind; json parse error: json: cannot unmarshal string into Go value of type struct { APIVersion string "json:\"apiVersion,omitempty\""; Kind string "json:\"kind,omitempty\"" }
```

The first issue is certainly that it is [trying to parse the filename as a kubeconfig](https://github.com/pulumi/pulumi-kubernetes/blob/master/pkg/provider/provider.go#L131). That is almost certainly a bug. After that, I'm not clear if the underlying codec will accept yaml/json interchangably, but I will try to find out and make a pr.

captncraig:
Fixes #209

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -17,6 +17,7 @@ package provider
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


hausdorff:
@captncraig per https://github.com/pulumi/pulumi-kubernetes/issues/209, if you want to add a different config option, say `kubernetes:kubeconfigFilename` I think we'll be willing to accept this PR. :)

PR #226, opened 2018-09-26
Allow use of custom resources with getResource from TypeScript

### LINKED ISSUE #225: Cannot use custom resources with `getResource` from TypeScript
The TypeScript typing has overloads for every known resource type, but does not have a user-visible overload that allows an arbitrary string. So although it's usable from JavaScript, the TypeScript compiler will complain about (e.g.):

```
kubePrometheus.getResource("monitoring.coreos.com/v1/ServiceMonitor", "kube-prometheus-exporter-kubernetes")
```

lukehoban:
TypeScript somewhat confusingly doesn't allow the implementation signature of an overload set to be used by callers.  So we need to include a more permissive `string` overload explicitly.  Note that this does mean that mis-typing a service type will lead to reduced typing, not to a TS error.  I think that is preferrable to preventing use of these functions from TS on customer resources.

Fixes #225.

### DIFF
--- a/pkg/gen/node-templates/provider.ts.mustache
+++ b/pkg/gen/node-templates/provider.ts.mustache
@@ -94,6 +94,8 @@ export namespace yaml {
         {{/Kinds}}
         {{/Versions}}
         {{/Groups}}
+        public getResource<T extends pulumi.CustomResource>(groupVersionKind: string, namespace: string): T;
+        public getResource<T extends pulumi.CustomResource>(groupVersionKind: string, namespace: string, name: string): T;
         public getResource<T extends pulumi.CustomResource>(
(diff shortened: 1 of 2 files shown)


hausdorff:
@lukehoban Alternatively, we could also add a `getCustomResource` function? Seems like a shame to jettison the hard TS error if you fat finger something.

lukehoban:
> @lukehoban Alternatively, we could also add a getCustomResource function? Seems like a shame to jettison the hard TS error if you fat finger something.

Made that change.  It's a little funny as a JavaScript API, as these two APIs are actually identical - just with different types.  @hausdorff Up to you if you think this is net better.

lukehoban:
@hausdorff What do you think of the new approach?

PR #227, opened 2018-09-27
Rename pack/ directory to sdk/

joeduffy:
This renames the pack/ directory to sdk/, mirroring our naming
conventions in other repos.

This has the added advantage of fixing pulumi/docs#599, since
apparently our doc generator assumed this was the repo's layout
already.


hausdorff:
Looks like CI failed because `gometalinter` timed out. It sometimes does this for seemingly no good reason. Once green I'll go ahead and merge.

Issue #224, opened 2018-09-26
Don't apply patch update for empty diffs


hausdorff:
When a resource fails to fully initialize when the Pulumi CLI exits, a subsequent run of `pulumi up` will typically result in an "empty" update.

This update still results in a patch between the two versions of inputs that have not changed, and the live object; the effect is to blank out computed fields. This is nearly never a problem, as the resource didn't initialize correctly anyway, so it doesn't hurt to overwrite them.

_Nearly_ never a problem. There appears to be at least one case where it does matter: `v1/Service`'s `.spec.clusterIP` field. The issue is that this field is immutable, and you can't change it. If you try to apply a patch that blanks it out, the update will fail.

Instead of applying this patch, we should switch to a strategy where we skip the patch step for all "empty" update steps.

hausdorff:
So as I was writing a test case of this for #230, I realized that actually we're doing the right thing here -- it's _the Chart_ that has the bug.

Here's what's going on:
* For context, this issue comes to us from #215.
* If you try to deploy the kube-prometheus chart, and you update any of the `Service`s that are `type: ClusterIP` (_e.g._, `kube-prometheus`), the will fail because `spec.clusterIP: Invalid value: "": field is immutable`
* This is because the Helm Chart is supplying `""` as the default value for the `clusterIP`.
* This causes client-go to generate a strategic merge patch `{"spec":{"clusterIP":""}}`, which attempts to "unset" `.spec.clusterIP`, which will fail because it's immutable.

**This is absolutely

hausdorff:
cc @joeduffy @lukehoban 

If you want proof that doesn't work with `kubectl` btw, run `kubectl create` the following `Service`, then change something trivial (the labels, say) and run `kubectl apply`. You'll see something like `The Service "my-service" is invalid: spec.clusterIP: Invalid value: "": field is immutable` 

```
kind: Service
apiVersion: v1
metadata:
  name: my-service
spec:
  clusterIP: ""
  type: ClusterIP
  selector:
    app: MyApp
  ports:
  - protocol: TCP
    port: 80
    targetPort: 9376
```

hausdorff:
@lukehoban @joeduffy I've gone ahead and submitted the fix for (1) to upstream: https://github.com/coreos/prometheus-operator/pull/1990

I'll look at adding a little bool run an optional `helm verify` check before submitting to the API server.

hausdorff:
[...]

PR #230, opened 2018-09-27
Fixes in read and await code paths

hausdorff:
Fixes various problems in the `Read` pipeline, specifically #234 and #231. See commit messages for more details.

### DIFF
--- a/pkg/gen/node-templates/provider.ts.mustache
+++ b/pkg/gen/node-templates/provider.ts.mustache
@@ -104,7 +104,6 @@ export namespace yaml {
             return this.getResourceImpl(groupVersionKind, namespaceOrName, name);
         }
 
-        
         /**
          * getCustomResource returns a resource defined by a CRD with the given group/version/kind and name.
          *
@@ -369,6 +368,17 @@ export namespace {{Group}} {
 
       {{/Properties}}
 
+      /**
+       * Get an existing {{Kind}} resource's state with the given name, ID, and optional extra
+       * properties used to qualify the lookup.
+       *
+       * @param name The _unique_ name of the resulting resource.
+       * @param id The _unique_ provider ID of the resource to lookup.
+       */
+      public static get(name: string, id: pulumi.Input<pulumi.ID>): {{Kind}} {
+          return new {{Kind}}(name, undefined, { id });
+      }
+
       public getInputs(): inputApi.{{Group}}.{{Version}}.{{Kind}} { return this.__inputs; }
       private readonly __inputs: inputApi.{{Group}}.{{Version}}.{{Kind}};
 
@@ -379,13 +389,13 @@ export namespace {{Group}} {
       * @param args The arguments to use to populate this resource's properties.
       * @param opts A bag of options that control this resource's behavior.
       */
-      constructor(name: string, args: inputApi.{{Group}}.{{Version}}.{{Kind}}, opts?: pulumi.CustomResourceOptions) {
+      constructor(name: string, args?: inputApi.{{Group}}.{{Version}}.{{Kind}}, opts?: pulumi.CustomResourceOptions) {
           let inputs: pulumi.Inputs = {};
           {{#Properties}}
           inputs["{{Name}}"] = {{{DefaultValue}}};
           {{/Properties}}
           super("kubernetes:{{APIVersion}}:{{Kind}}", name, inputs, opts);
-          this.__inputs = args;
+          this.__inputs = <any>args;
       }
     }
 
--- a/pkg/provider/diff.go
+++ b/pkg/provider/diff.go
@@ -41,9 +41,14 @@ type properties []string
 
 var forceNew = groups{
 	"apps": versions{
-		// NOTE: These fields do NOT trigger a replace in extensions/v1beta1 or apps/v1beta1.
-		"v1beta2": kinds{"Deployment": deployment},
-		"v1":      kinds{"Deployment": deployment},
+		// NOTE: .spec.selector triggers a replacement in Deployment only AFTER v1beta1.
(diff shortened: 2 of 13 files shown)


hausdorff:
@lukehoban sorry, this was supposed to be marked [WIP]. It's not ready to merge.

hausdorff:
@lukehoban I'm moving the ignore-empty-diff-updates out of this PR, per discussion in #224. With additional investigation, I think we have conclusive evidence that the root issue was a bug in the underlying Helm Chart, and that we have no evidence that empty-diff updates have any deleterious effect.

`kubectl` will ignore `apply`s of empty-diff updates, so if we want to do that here too, we can -- but it does imply writing significant additional test code, it adds complexity to the already-complex `update` path, and there does not seem to be any benefit.

hausdorff:
Ok this is ready for review. As I said in Slack, I don't think I really understand the semantics of the state arguments meant to be passed to `get`, so I hope I've gotten those right. Otherwise, I think this should be pretty much ready to go.

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -379,13 +391,13 @@ export namespace {{Group}} {
       * @param args The arguments to use to populate this resource's properties.
       * @param opts A bag of options that control this resource's behavior.
       */
-      constructor(name: string, args: inputApi.{{Group}}.{{Version}}.{{Kind}}, opts?: pulumi.CustomResourceOptions) {
+      constructor(name: string, args?: inputApi.{{Group}}.{{Version}}.{{Kind}}, opts?: pulumi.CustomResourceOptions) {
```
swgillespie: Why do we need this?

pkg/provider/diff.go
```diff
@@ -41,9 +41,14 @@ type properties []string
...
-		"v1beta2": kinds{"Deployment": deployment},
-		"v1":      kinds{"Deployment": deployment},
+		// NOTE: .spec.selector triggers a replacement in Deployment only AFTER v1beta1.
+		"v1beta1": kinds{"StatefulSet": statefulSet},
+		"v1beta2": kinds{
+			"Deployment":  deployment,
+			"StatefulSet": statefulSet},
+		"v1": kinds{
+			"Deployment":   deployment,
+			"StatefuleSet": statefulSet},
```
swgillespie: typo: `StatefulSet`

pkg/gen/node-templates/provider.ts.mustache
hausdorff: All the other packages do it this way and it's annoying to have to manually specify an empty object when all the properties are optional.

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -369,6 +368,17 @@ export namespace {{Group}} {
 
       {{/Properties}}
 
+      /**
+       * Get an existing {{Kind}} resource's state with the given name, ID, and optional extra
+       * properties used to qualify the lookup.
```
lukehoban: I think this comment is not correct given that the extra properties are not exposed (and not needed)?

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -369,6 +368,17 @@ export namespace {{Group}} {
...
+       *
+       * @param name The _unique_ name of the resulting resource.
+       * @param id The _unique_ provider ID of the resource to lookup.
```
lukehoban: Is "provider ID" the right term in the Kubernetes context?

pkg/gen/node-templates/provider.ts.mustache
lukehoban: I believe we only do this for resources which have all optional input argument object properties.  I'm less sure this makes sense to make optional overall.

sdk/nodejs/provider.ts
```diff
@@ -1518,14 +1528,14 @@ export namespace admissionregistration {
       * @param args The arguments to use to populate this resource's properties.
       * @param opts A bag of options that control this resource's behavior.
       */
-      constructor(name: string, args: inputApi.admissionregistration.v1alpha1.InitializerConfiguration, opts?: pulumi.CustomResourceOptions) {
+      constructor(name: string, args?: inputApi.admissionregistration.v1alpha1.InitializerConfiguration, opts?: pulumi.CustomResourceOptions) {
           let inputs: pulumi.Inputs = {};
           inputs["apiVersion"] = "admissionregistration.k8s.io/v1alpha1";
-          inputs["initializers"] = args.initializers;
+          inputs["initializers"] = args && args.initializers || undefined;
```
lukehoban: Why do we need `|| undefined` here?  If `args` is not provided it will be `undefined` and so the result of the `&&` will already be `undefined` - right?

hausdorff: This is pretty much copied from the AWS provider, and I kept the `||` here to make the equational reasoning explicit, which I like.

pkg/gen/node-templates/provider.ts.mustache
hausdorff: I thought this made sense to keep, but I can certainly make the comment more clear.

pkg/gen/node-templates/provider.ts.mustache
hausdorff: @lukehoban `get` needs to call the constructor with `null` as args. I'm not sure of a really great way to make this work. One option is to do what we're doing here, which doesn't seem so bad because omitting it will result in a validation error anyway.

lukehoban: If this is how we do it in TF implementations then fine to kee it the same here. I recall thinking the approach there was awkward too - but better to be consistent.

PR #236, opened 2018-10-11
Initialization error corner case fixes

hausdorff:


pkg/await/await.go
```diff
@@ -104,10 +105,13 @@ func Read(
...
-	// `CustomResource#get`. Simply return the object.
-	if inputs == nil || len(inputs.Object) == 0 {
-		return clientForResource.Get(name, metav1.GetOptions{})
+	outputs, err := clientForResource.Get(name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	} else if inputs == nil || len(inputs.Object) == 0 {
+		// No inputs means that we do not manage the resource, i.e., it's a call to
+		// `CustomResource#get`. Simply return the object.
+		return outputs, nil
```
lukehoban: Not sure I follow what's motivating the above change - what issue is this fixing?

sdk/nodejs/provider.ts
```diff
@@ -1508,14 +1508,18 @@ export namespace admissionregistration {
...
+       * Pulumi will keep track of this resource using `name` as the Pulumi ID.
+       *
+       * @param name _Unique_ name used to register this resource with Pulumi.
+       * @param kubeId An ID for the Kubernetes resource to retrive. Takes the form
+       *  <namespace>/<name> or <name>.
        */
-      public static get(name: string, id: pulumi.Input<pulumi.ID>): InitializerConfiguration {
-          return new InitializerConfiguration(name, undefined, { id });
+      public static get(name: string, kubeId: pulumi.Input<pulumi.ID>): InitializerConfiguration {
+          return new InitializerConfiguration(name, undefined, { id: kubeId });
```
lukehoban: I think this was PR feedback based on a comment I made - but I don't actually think we should change `id` to `kubeId`.  My comment was about whether the string `"The _unique_ provider ID"` was the correct terminology to use in the comment - I think the new comment is an improvement, but doesn't require changing the parameter name.

pkg/await/await.go
hausdorff: Recapping offline discussion, in the case of `.get` we do _not_ want to run validation routines  -- we want to return the object itself, in whatever stat it is.

In the case of a `refresh`, we want to validate the outputs are in a valid state.

We run `clientForResource.Get` so that we can pass `ouptuts` to the validation logic below.

PR #241, opened 2018-10-15
WIP: Use `pulumi.Input` in Helm Chart API

hausdorff:


sdk/nodejs/yarn.lock
```diff
@@ -45,9 +45,9 @@
   version "1.1.0"
   resolved "https://registry.yarnpkg.com/@protobufjs/utf8/-/utf8-1.1.0.tgz#a777360b5b39a1a2e5106f8e858f2fd2d060c570"
 
-"@pulumi/pulumi@^0.15.1":
-  version "0.15.1"
-  resolved "https://registry.yarnpkg.com/@pulumi/pulumi/-/pulumi-0.15.1.tgz#1b94077ec0bfb8fef868d6aa66643d7f726c0678"
+"@pulumi/pulumi@dev":
+  version "0.16.1-dev-1539638908-g00a22b27"
```
lukehoban: If we want to release this package today - we would ideally not depend on unreleased `@pulumi/pulumi` builds.  Can we work around the issue in that layer for now so that we can release a version of the Kubernetes package based on 0.16 of the `@pulumi/pulumi` package?  Or is a patch release of `@pulumi/pulumi` strictly necessary to deliver this fix in `@pulumi/kubernetes`?

hausdorff: Yep. Let's do it.

pkg/gen/node-templates/provider.ts.mustache
```diff
@@ -95,12 +101,12 @@ export namespace yaml {
...
         {{#Kinds}}
-        public getResource(groupVersionKind: "{{RawAPIVersion}}/{{Kind}}", name: string): {{Group}}.{{Version}}.{{Kind}};
-        public getResource(groupVersionKind: "{{RawAPIVersion}}/{{Kind}}", namespace: string, name: string): {{Group}}.{{Version}}.{{Kind}};
+        public getResource(groupVersionKind: "{{RawAPIVersion}}/{{Kind}}", name: string): pulumi.Output<{{Group}}.{{Version}}.{{Kind}}>;
+        public getResource(groupVersionKind: "{{RawAPIVersion}}/{{Kind}}", namespace: string, name: string): pulumi.Output<{{Group}}.{{Version}}.{{Kind}}>;
         {{/Kinds}}
         {{/Versions}}
         {{/Groups}}
-        public getResource(groupVersionKind: string, namespaceOrName: string, name?: string): pulumi.CustomResource {
+        public getResource(groupVersionKind: string, namespaceOrName: string, name?: string): pulumi.Output<pulumi.CustomResource> {
```
lukehoban: Note that this is a breaking change - and will also be a bit more cumbersome to use in practice.  It's also a bit of a shame, because `CustomResource` objects are already delayed.  Not sure there is anything we can do here in the nearterm to handle this better - but another case to think about as part of https://github.com/pulumi/pulumi/issues/994.

pgavlin:
@lukehoban @hausdorff this is ready for review.

pkg/gen/nodejs-templates/helm.ts.mustache
```diff
@@ -63,63 +63,76 @@ export namespace v2 {
...
-                chartDir.removeCallback();
-                overrides.removeCallback();
-            }
+			const allConfig = pulumi.output(config);
+			const configDeps = Array.from(<Set<pulumi.Resource>>(<any>allConfig).resources());
+
+			(<any>allConfig).isKnown.then((isKnown: boolean) => {
+				if (!isKnown) {
+					// Note that this can only happen during a preview.
+					pulumi.log.info("Note: some chart inputs are unknown; the chart's resources will not be displayed.");
```
hausdorff: My only big UX feedback in this PR is:

* I'd really like this to be correlated to the actual chart resource. Right now this `info` is displayed under the stack.
* As an `info` it's kind of easy to miss because it appears in the diagnostics summary. It would be nice if this was integrated as a first-class citizen in the display, like:
  
  ```
  
       Type                                         Name                   Plan       Info
   +   pulumi:pulumi:Stack                          azure-mean-azure-mean  create     
   +   └─ kubernetes:helm.sh:Chart                  node                   create     [Can't preview] all chart values must be known ahead of time to generate an accurate preview
  ```

  If we're not willing to invest in that, maybe we should do a `warning`? I don't know. It is a warning in the sense you should tread carefully and you want to get users attention, but it's also _not_ in the sense that you can't really correct it.
* I (personally?) find the wording here a little strange. I don't think what I'm suggesting above is optimal, but I do think we should mention specifically that it's the _values_ passed to the chart that is causing this. Helm users will understand that better than "inputs".

pkg/gen/nodejs-templates/helm.ts.mustache
```diff
@@ -63,63 +63,76 @@ export namespace v2 {
...
-                throw new pulumi.RunError(e.toString());
-            } finally {
-                // Clean up temporary files and directories.
-                chartDir.removeCallback();
-                overrides.removeCallback();
-            }
+			const allConfig = pulumi.output(config);
+			const configDeps = Array.from(<Set<pulumi.Resource>>(<any>allConfig).resources());
+
+			(<any>allConfig).isKnown.then((isKnown: boolean) => {
```
hausdorff: Just so I'm super clear: this is only true if all the _values_ are known, or if all the _inputs_ are known? It seems to me we want the former, rather than the latter.

pgavlin: This is only true if all of the inputs are known. I believe that this is what we want, as any unknown input will prevent us from generating an accurate preview. On a more concrete level, any unknown input will prevent us from running the `apply` below.

pgavlin:
Here's the preview experience: https://asciinema.org/a/rFMPmQwBjz4Vo5kApBKKvULih

pkg/gen/nodejs-templates/helm.ts.mustache
hausdorff: Hmm, other than `values`, how can other fields affect the preview?

hausdorff: Got it. You're right. Let's do it your way.

pkg/gen/nodejs-templates/helm.ts.mustache
```diff
@@ -63,63 +63,76 @@ export namespace v2 {
...
+				if (!isKnown) {
+					// Note that this can only happen during a preview.
+					pulumi.log.info("[Can't preview] all chart values must be known ahead of time to generate an accurate preview.", this);
+				}
+			});
+
+			this.resources = allConfig.apply(cfg => {
+				// Create temporary directories and files to hold chart data and override values.
```
hausdorff: oh, eek, I almost missed this -- these are tabs. Can we please double-check the formatting to make sure it's consistent with what we have? :)

pgavlin: I'll fix this now--got a little itchy on the trigger.

PR #246, opened 2018-10-19
Fix v1/Secret replacement logic, apiextensions.CustomResource typing issues

hausdorff:


hausdorff:
Luke signed off verbally when GitHub was down.

PR #247, opened 2018-10-23
Allow `helm.v2.Chart` to use Charts on filesystem

### LINKED ISSUE #238: Helm: Cannot install local charts
currently the helm integration always assumes the chart is in a remote repository. However it is not uncommon to simply specify a local chart path.
A prominent example for where this actually required is to install istio https://istio.io/docs/setup/kubernetes/helm-install/ because istio is currently not published to any chart repo.

Please add support to install a chart from a local directory or file by specifying a filepath and skip the fetch step for those.

hausdorff:
This change will be to change the `Chart` constructor to take `ChartOpts
| LocalChartOpts` as arguments instead of simply `ChartOpts`.
`ChartOpts` represents a Chart in a repository, while `LocalChartOpts`
represents a Chart in a filesystem directory.

Fixes #238.


sdk/nodejs/helm.ts
```diff
@@ -10,17 +10,32 @@ import * as path from "./path";
...
-        transformations?: ((o: any) => void)[];
         fetchOpts?: FetchOpts;
     }
 
+    function isChartOpts(o: any): o is ChartOpts {
+        return "repo" in o && "chart" in o && "version" in o;
+    }
+
+    export interface LocalChartOpts extends BaseChartOpts {
+        directory: string;
```
lukehoban: Nit - is `path` or `folder` better here?  Not sure of good precedent to base on - but for some reason `directory` felt a little surprising to me.

Issue #197, opened 2018-09-09
Deployment awaiter is wrong for the case previous state had init errors


hausdorff:
We want to check if the rollout succeeded only if this is not the first time we've succeeded. If there were previously initialization errors, we currently expect that any generation > 1 will imply a rollout.

hausdorff:
Looking closer, it looks like the `extensions/v1beta1` version of the API doesn't seem to have a condition indicating that the rollout was complete. _sigh_

hausdorff:
We've punted this out because, looking at it now, I'm not actually sure this is a bug. We filed it before #199, which seems likely to be the real problem.

lukehoban:
@hausdorff What's the current plan for this and #199?  Are they the same issue?  Is there a fix we know we'll take in M19?

hausdorff:
I don't think this is a bug. Let's close.

PR #251, opened 2018-10-24
`cancellationError` should implement `InitializationError`

hausdorff:
`InitializationError` is used to signal to the provider that a resource
has successfully been created, but not initialized. Currently
`cancellationError` does not implement this interface, so if
`Deployment` or `Service` are cancelled, this is not registered as a
partial failure with the engine or service.

This did not used to be the case, but was a regression with #236.

We already have good tests to make sure that `Deployment` and `Service` return the appropriate cancellation and timeout errors under various circumstances -- it's not clear that additional tests would help here, as the problem is that the error implementations didn't technically implement the correct interface. As a safeguard against this in the future I've added a sentinel check `var _ InitializationError = (*cancellationError)(nil)`, which should cause the compiler to error out on regression.

### DIFF
--- a/pkg/await/error.go
+++ b/pkg/await/error.go
@@ -25,6 +25,7 @@ type cancellationError struct {
 
 var _ error = (*cancellationError)(nil)
 var _ AggregatedError = (*cancellationError)(nil)
+var _ InitializationError = (*cancellationError)(nil)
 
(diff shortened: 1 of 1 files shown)


hausdorff:
Let's merge this now so that I can do a new patch release, and look at adding integration tests as part of #197.

PR #253, opened 2018-10-27
Initial cut at Python support

hausdorff:


pkg/gen/python-templates/kind.py.mustache
```diff
@@ -0,0 +1,44 @@
...
+
+        __props__['kind'] = '{{Kind}}'
+        self.kind = '{{Kind}}'
+        {{#RequiredProperties}}
+
+        if not {{Name}}:
+            raise TypeError('Missing required property {{Name}}')
+        elif not isinstance({{Name}}, {{PropType}}):
+            raise TypeError('Expected property aliases to be a {{PropType}}')
+        self.{{Name}} = {{Name}}
```
swgillespie: Don't think you need this

pkg/gen/python-templates/kind.py.mustache
```diff
@@ -0,0 +1,44 @@
...
+        {{{Comment}}}
+        __props__['{{Name}}'] = {{Name}}
+        {{/RequiredProperties}}
+        {{#OptionalProperties}}
+
+        if {{Name}} and not isinstance({{Name}}, {{PropType}}):
+            raise TypeError('Expected property aliases to be a {{PropType}}')
+        self.{{Name}} = {{Name}}
```
swgillespie: Same here, this is covered by the `__props__` right below

swgillespie: or wait... maybe you do? I know that I'm working on populating properties the "node" way (a way that doesn't require this).

If you do have to set attributes on self, you should use self.__setattr__(name, {{Name}}) instead so we don't get syntax errors if any property names collide with keywords.

hausdorff: I'm pretty much copying the AWS provider here. I actually have very little knowledge of what this _should_ look like.

joeduffy: I would just copy TF providers. We can then improve both of them similarly.

Makefile
```diff
@@ -31,14 +33,26 @@ $(OPENAPI_FILE)::
 build:: $(OPENAPI_FILE)
 	$(GO) install $(VERSION_FLAGS) $(PROJECT)/cmd/$(PROVIDER)
 	$(GO) install $(VERSION_FLAGS) $(PROJECT)/cmd/$(CODEGEN)
-	for LANGUAGE in "nodejs" ; do \
-		$(CODEGEN) $(OPENAPI_FILE) pkg/gen/node-templates $(PACKDIR)/$$LANGUAGE || exit 3 ; \
+	for LANGUAGE in "python" ; do \
```
joeduffy: Shouldn't this be `for LANGUAGE in "nodejs" "python" ; do \`? In other words, we want to generate both the Node.js package _and_ Python, not just one of them.

cmd/pulumi-gen-kubernetes/main.go
```diff
@@ -25,11 +25,13 @@ import (
 )
 
 func main() {
-	if len(os.Args) < 4 {
+	if len(os.Args) < 5 {
 		log.Fatal("Usage: gen <swagger-file> <template-dir> <out-dir>")
```
joeduffy: Nit, usage needs to be updated to 5-arg form.

pkg/gen/python-templates/version__init__.py.mustache
```diff
@@ -0,0 +1,8 @@
+# coding=utf-8
+# *** WARNING: this file was generated by the Pulumi Kubernetes codegen tool. ***
+# *** Do not edit by hand unless you're certain you know what you are doing! ***
+
+# Export this package's modules as members:
+{{#Kinds}}
+from {{Kind}} import *
```
swgillespie: For Python 3: `from .{{Kind}} import *`

PR #256, opened 2018-11-01
Bring up Python 3

swgillespie:
Extension of https://github.com/pulumi/pulumi-kubernetes/pull/253 that gets us to the "hey look, this works!" stage of Python 3.

Big ideas here include:

1. Using https://github.com/pulumi/pulumi/pull/2140 to rewrite input and output bags as they flow to and from the engine, allowing us to project a facade of snake cased dictionaries while really speaking to the engine in terms of Kubernetes casing
2. Change up the codegen a little bit to avoid writing to `self`, since the Python 3 SDK does this already
3. Remove codegen type tests, since they don't work with Outputs


swgillespie:
I'm trying to fix the build now, once it's green this should be good to merge.

swgillespie:
Opened https://github.com/pulumi/pulumi-kubernetes/issues/258 to track auto-generating the tables in `tables.py` as part of the build.

pkg/gen/python-templates/kind.py.mustache
```diff
@@ -1,9 +1,11 @@
 import pulumi
 import pulumi.runtime
 
+from ...tables import _CASING_FORWARD_TABLE, _CASING_BACKWARD_TABLE
```
hausdorff: Not a huge deal, but it seems like it would be nicer to have this read `tables._CASING_FORWARD` instead.

pkg/gen/python-templates/tables.py
```diff
@@ -0,0 +1,1140 @@
+_CASING_FORWARD_TABLE = {
```
hausdorff: I think we should filter out anything that's not mixed case. I think this will greatly lower the probability of making a mistake, and it makes auditing much easier.

Probably worth lining up the semicolons too if it's not too much work...

hausdorff: Also, for stuff that's not generated from mustache, we are just putting it in the sdk directory, rather than having two copies of it.

hausdorff:
Just so we're all on the same page, @lukehoban @joeduffy, I just talked to @swgillespie my understanding here is that:

* We are going to use named parameters instead of `dict` for the arguments because right now that makes for a better autocomplete experience.
* It might be possible to use typed `dict`, but we don't know for sure, and it would be some amount of work to find out.
* It sounds like we want to do a release today, so the plan seems to be that we'll ship something like this, and then possibly add breaking changes next release with the typed `dict` stuff.

Is that right?

pkg/gen/python.go
```diff
@@ -93,3 +106,182 @@ func PythonClient(
...
+	"raise":    true,
+	"return":   true,
+	"try":      true,
+	"while":    true,
+	"with":     true,
+	"yield":    true,
+}
+
+// pyName turns a variable or function name, normally using camelCase, to an underscore_case name.
+func pyName(name string) string {
```
pgavlin: Is this copy-pasted from some other part of our codebase (e.g. `pulumi-terraform`)? Would it make more sense to stick this is a utility package somewhere to ease bugfixes?

swgillespie: yep, it totally is.

What I'd really love is a retargetable unified Pulumi code generator, but I don't mind checking in utility functions here and there that can be shared between code generators. Do you know where such a utility package would go? I feel weird putting it in `pulumi/pulumi`, since it's not used in that repo, and I'd feel weird putting it in `pulumi-terraform` too since it would be weird for this repo to depend on it.

swgillespie:
@hausdorff just addressed your feedback in the latest commit.

pkg/gen/python.go
pgavlin: I think `pulumi/pulumi` is a fine place to put it.

swgillespie: Opened https://github.com/pulumi/pulumi/issues/2151 to track.

hausdorff: FWIW I actually don't think it's a huge deal if we have this code in two places. We're using it to generate diff-able Python code, so we _should_ notice if there is an improper generation somewhere, and it _should_ be trivial to fix.

That said, whatever you all want. :)

PR #259, opened 2018-11-02
`timeoutError` should implement `InitializationError`

hausdorff:
`InitializationError` is used to signal to the provider that a resource
has successfully been created, but not initialized. Currently
`timeoutError` does not implement this interface, so if
`Deployment` or `Service` are cancelled, this is not registered as a
partial failure with the engine or service.

This did not used to be the case, but was a regression with #236.


hausdorff:
Sorry folks, should have checked that both of these implement `InitializationError`. My bad.

Issue #240, opened 2018-10-14
Add Istio to our nightly tests


geekflyer:
This is mostly related to #239 but I wanted to create a separate issue for istio in general due to the severity of problems I ran into and discuss whether there are other issues to create out of this.

In general I think istio is an excellent example of a non-trivial k8s application that pulumi should adopt as part of their test suite. Why? Istio uses helm, istio deploys more than 100 resources, it uses a lot of advanced resource types, like CRDs etc.

Currently the experience when attempting to deploying istio via pulumi is very bad.
This is the code that can be used for testing:
1. `helm repo add istio https://istio.io/charts`
2. deploy:
```ts
import { helm, core } from '@pulumi/kubernetes';

const appName = 'istio';

const namespaceName = `${appName}-system`;

const namespace = new core.v1.Namespace(namespaceName, {
  metadata: { name: namespaceName }
});

const chart = new helm.v2.Chart(
  appName,
  {
    repo: 'istio',
    chart: appName,
    namespace: namespaceName,
    version: '1.0.1',
    // for all options check https://github.com/istio/istio/tree/master/install/kubernetes/helm/istio
    values: {
         kiali: { enabled: true}
      }
  },
  { dependsOn: [namespace] }
);
```

This is the list of issues I ran into - in order:
### 1. getting the chart involves manual steps
I found no way to install a chart from local. Then I found istio actually has a not documented custom helm repo endpoint but using this in pulumi requires a manual `helm repo add` step. Those 2 issues are already addressed in #229 and #238.
### 2. Pulumi crashes because of duplicate yaml key in kiali definitions with:
```sh
Previewing changes:

     Type                         Name                                 Plan          Info
 *   pulumi:pulumi:Stack          api-cluster-dev-b-api-cluster-dev-b  no change     2 errors
 ~   └─ kubernetes:helm.sh:Chart  istio                                update        changes: ~ values

Diagnostics:
  pulumi:pulumi:Stack: api-cluster-dev-b-api-cluster-dev-b
    error: YAMLException: duplicated mapping key at line 2450, column 5:
            name: http-kiali
            ^

    error: an unhandled error occurred: Program exited with non-zero exit code: 1

error: an error occurred while advancing the preview
```

[...]

PR #262, opened 2018-11-05
Retry creation REST calls that fail

### LINKED ISSUE #239: Helm: resource ordering not working
The helm implementation does some ordering in an attempt to deploy resources in a "more likely to work" order as implemented here: https://github.com/pulumi/pulumi-kubernetes/blob/master/sdk/nodejs/helm.ts#L109 .
Unfortunately this doesn't seem to work at all or have any effect on pulumis deploy order in practice. Resources get deployed in completely random order and it hence it fails.

The best way to reproduce this is to attempt to install istio, which has a lot of resources:
1. `helm repo add istio https://istio.io/charts`
2. deploy:
```ts
import { helm, core } from '@pulumi/kubernetes';

const appName = 'istio';

const namespaceName = `${appName}-system`;

const namespace = new core.v1.Namespace(namespaceName, {
  metadata: { name: namespaceName }
});

const chart = new helm.v2.Chart(
  appName,
  {
    repo: 'istio',
    chart: appName,
    namespace: namespaceName,
    version: '1.0.1',
    // for all options check https://github.com/istio/istio/tree/master/install/kubernetes/helm/istio
    values: {}
  },
  { dependsOn: [namespace] }
);
```

hausdorff:
This commit will introduce a simple retry mechanism that makes the
Kubernetes provider more tolerant to misordered resource creation, as is
particularly common in Helm and Kubernetes YAML. For example, if
Kubernetes YAML specifies to create a `Pod` before the `Namespace` it
resides in, it is likely this will cause `pulumi up` to fail. Another
example is creating `CustomResoruces` before an operator or controller has
the opportunity to register the relevant `CustomResourceDefinition`s.

We ameliorate this bug by retrying creation REST calls that fail, with
some backoff, and some maximum retry threshold. The effects of this are:

* Makes the Kubernetes provider somewhat tolerant to undeclared
  dependencies between resources.
* Makes the Kubernetes provider somewhat tolerant to under-specified
  CRDs, custom controllers, and custom operators, since the retry logic
  can stand in for all but the most complex of these (which would still
  require await logic).

This commmit represents major progress towards #240, and should (we
believe) fix #239.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,8 @@ build:: $(OPENAPI_FILE)
 	sed -i.bak 's/$${VERSION}/$(VERSION)/g' ${PACKDIR}/nodejs/bin/package.json
 
 lint::
-	$(GOMETALINTER) ./cmd/... ./pkg/... | sort ; exit "$${PIPESTATUS[0]}"
+	$(GOMETALINTER) ./cmd/... | sort ; exit "$${PIPESTATUS[0]}"
+	$(GOMETALINTER) ./pkg/... | sort ; exit "$${PIPESTATUS[0]}"
 
 install::
 	GOBIN=$(PULUMI_BIN) $(GO) install $(VERSION_FLAGS) $(PROJECT)/cmd/$(PROVIDER)
--- a/pkg/await/await.go
+++ b/pkg/await/await.go
@@ -58,8 +58,28 @@ func Creation(
 		return nil, err
 	}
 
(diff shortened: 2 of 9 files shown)


lukehoban:
Don you have a link to similar logic in `kubectl` we could reference as baseline for the strategy and chosen constants here?  Would be nice to include in code comments as well.

PR #263, opened 2018-11-05
Pin to dev builds of pulumi/pulumi

swgillespie:
As part of the release process, we'll need to change the constraint to a non-developer build like we do for all of our other package repos.

### DIFF
--- a/pkg/gen/nodejs-templates/package.json.mustache
+++ b/pkg/gen/nodejs-templates/package.json.mustache
@@ -13,7 +13,7 @@
         "build": "tsc"
     },
     "dependencies": {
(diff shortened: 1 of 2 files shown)


swgillespie:
Whoops, looks like that package.json is autogenerated.

swgillespie:
Currently on-hold until I find some time to fix the yarn.lock issue.

ellismg:
After talking with @swgillespie the intention of this change was to ensure that we run against the yet to be released `@pulumi/pulumi`, instead of needing the next version of `@pulumi/pulumi` to build or run Kubernetes.

We'll be getting the "run tests against head" coverage a slightly different way so it is safe to close this out.

PR #270, opened 2018-11-14
Update Python dependency to released version

ellismg:


ellismg:
I believe that I know what I am doing and the comment at the top of setup.py is a lie and this file is not auto-generated, so I just edited manually.  @hausdorff if I did something wrong, let me know.

PR #271, opened 2018-11-14
Fixes in how the provider handles CRDs and CRs

hausdorff:
See commit messages for a more detailed discussion.

### DIFF
--- a/pkg/await/await.go
+++ b/pkg/await/await.go
@@ -53,11 +53,6 @@ func Creation(
 	ctx context.Context, host *provider.HostClient, pool dynamic.ClientPool,
 	disco discovery.ServerResourcesInterface, urn resource.URN, inputs *unstructured.Unstructured,
 ) (*unstructured.Unstructured, error) {
-	clientForResource, err := client.FromResource(pool, disco, inputs)
-	if err != nil {
-		return nil, err
-	}
-
 	// Issue create request. We retry the create REST request on failure, so that we can tolerate
 	// some amount of misordering (e.g., creating a `Pod` before the `Namespace` it goes in;
 	// creating a custom resource before the CRD is registered; etc.), which is common among Helm
@@ -69,10 +64,23 @@ func Creation(
 	// nolint
 	// https://github.com/kubernetes/kubernetes/blob/54889d581a35acf940d52a8a384cccaa0b597ddc/pkg/kubectl/cmd/apply/apply.go#L94
 	var outputs *unstructured.Unstructured
-	err = sleepingRetry(
+	var clientForResource dynamic.ResourceInterface
+	err := sleepingRetry(
 		func(i uint) error {
 			if i > 0 {
 				_ = host.LogStatus(ctx, diag.Info, urn, fmt.Sprintf("Creation failed, retrying (%d)", i))
+			}
+
+			// Recreate the client for resource, in case the client's cache of the server API was
+			// invalidated. For example, when a CRD is created, it will invalidate the client cache;
+			// this allows CRs that we tried (and failed) to create before to re-try with the new
+			// server API, at which point they should hopefully succeed.
+			var err error
+			if clientForResource == nil {
+				clientForResource, err = client.FromResource(pool, disco, inputs)
+				if err != nil {
+					return err
+				}
 			}
 			outputs, err = clientForResource.Create(inputs)
 			return err
--- a/pkg/await/await.go
+++ b/pkg/await/await.go
@@ -301,6 +301,24 @@ func Deletion(
 	disco discovery.DiscoveryInterface, urn resource.URN, gvk schema.GroupVersionKind, namespace,
 	name string,
 ) error {
+	// nilIfGVKDeleted takes an error and returns nil if `errors.IsNotFound`; otherwise, it returns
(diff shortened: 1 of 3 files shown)


pkg/await/await.go
```diff
@@ -299,28 +307,39 @@ func Deletion(
...
+		// deleted.
+		return nil
+	} else if err != nil {
 		return err
 	}
 
 	// Issue deletion request.
 	err = clientForResource.Delete(name, &deleteOpts)
 	if err != nil && !errors.IsNotFound(err) {
 		return fmt.Errorf("Could not find resource '%s/%s' for deletion: %s", namespace, name, err)
```
lukehoban: This first branch is confusing - it fires on ~"is found", but reports "could not find".  With the addition of the new code here, it also means the third branch will never be hit, right?

pkg/await/await.go
```diff
@@ -299,28 +307,39 @@ func Deletion(
...
 	}
 
 	// Obtain client for the resource being deleted.
 	clientForResource, err := client.FromGVK(pool, disco, gvk, namespace)
-	if err != nil {
+	if errors.IsNotFound(err) {
+		// If the resource G/V/K is not found, but was successfully created before, (e.g., it was a
+		// CR, but the CRD has since been removed) we can assume that it has now been successfully
+		// deleted.
+		return nil
```
lukehoban: This looks like it would have a much broader impact than what the comment describes - basically it means that delete will now succeed in cases where the resources was removed out-of-band of Pulumi.  I can believe that is a good semantic, but worth calling this out as a change for all resources, not just CRDs?

pkg/await/await.go
```diff
@@ -299,28 +307,39 @@ func Deletion(
...
+	// Manually set delete propagation for Kubernetes versions < 1.6 to avoid bugs.
 	deleteOpts := metav1.DeleteOptions{}
 	if version.Compare(1, 6) < 0 {
 		// 1.5.x option.
 		boolFalse := false
 		// nolint
 		deleteOpts.OrphanDependents = &boolFalse
-	} else {
+	} else if version.Compare(1, 7) < 0 {
 		// 1.6.x option. (NOTE: Background delete propagation is broken in k8s v1.6, and maybe later.)
```
lukehoban: Perhaps remove "and maybe later" if we now believe that is not a problem?

pkg/provider/provider.go
```diff
@@ -442,6 +442,13 @@ func (k *kubeProvider) Create(
 		return nil, initializationError(client.FqObjName(initialized), awaitErr, inputsAndComputed)
 	}
 
+	// Refresh the client cache if this was a CRD. This allows CRs to be created in the same program
+	// as the relevant CRDs.
```
lukehoban: Nit - technically "invalidate" not "refresh"?

pkg/await/await.go
```diff
@@ -299,28 +307,39 @@ func Deletion(
...
 		fg := metav1.DeletePropagationForeground
 		deleteOpts.PropagationPolicy = &fg
 	}
 
 	// Obtain client for the resource being deleted.
 	clientForResource, err := client.FromGVK(pool, disco, gvk, namespace)
-	if err != nil {
+	if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
hausdorff: @lblackstone Did you mean:

```
	if err != nil {
		if errors.IsNotFound(err) {
			// If the resource G/V/K is not found, but was successfully created before, (e.g., it
			// was a CR, but the CRD has since been removed) we can assume that it has now been
			// successfully deleted.
			return nil
		}
		return err
	}
```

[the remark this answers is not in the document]
hausdorff: Oh, I think you meant this snippet to be for the similar code block below.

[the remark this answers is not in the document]
lblackstone: Oops, you're right.

pkg/provider/provider.go
```diff
@@ -442,6 +442,13 @@ func (k *kubeProvider) Create(
...
+	if newInputs.GetKind() == "CustomResourceDefinition" &&
```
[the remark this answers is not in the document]
hausdorff: I couldn't find a constant for this kind so I just put it into a function, `isCRD`.

hausdorff:
Ok I think we're ready for another spin on this @lblackstone. I refactored the `isNotFound` stuff into one place and then addressed all the other comments.

pkg/await/await.go
```diff
@@ -293,36 +301,53 @@ func Deletion(
...
+	//
+	// Rationale: If we have gotten to this point, this resource was successfully created and is now
+	// being deleted. This implies that the G/V/K once did exist, but now does not. This, in turn,
+	// implies that it has been successfully deleted. For example: the resource was likely a CR, but
+	// the CRD has since been removed. Otherwise, the resource was deleted out-of-band.
+	//
+	// This is necessary for CRs, which are often deleted after the relevant CRDs (especially in
+	// Helm charts), and it is acceptable for other resources because it is semantically like
+	// running `refresh` before deletion.
+	nilIfGVKDeleted := func(err error) error {
```
lblackstone: Nice! 👍

PR #284, opened 2018-11-18
Allow more flexible chart options

### LINKED ISSUE #229: Problem with URL as a Helm chart repository
### Problem

When trying to install a helm chart that is not in the "stable" Helm repository like this:
```
    const fluxChart = new k8s.helm.v2.Chart("flux",
            {
                repo: "https://weaveworks.github.io/flux",
                chart: "flux",
                version: "0.3.3",
                namespace: "flux"
            }
        );
```

I got an error like this:

```
    Error: Command failed: helm fetch https\://weaveworks.github.io/flux/flux --untar --version 0.3.3 -- destination /var/folders/_6/...
    Error: Failed to fetch https://weaveworks.github.io/flux/flux : 404 Not Found
```

Looking into the code of "helm.js", the "fetch" method could handle the "opts.repo" option, but the constructor merges the repo + chart name like this: `${config.repo}/${config.chart}` and doesn't use that option.

### Workaround
Add the repository to the local Helm installation. E.g. for the above case:

```
helm repo add weaveworks https://weaveworks.github.io/flux
```  

and modify the Pulumi resource definition like this:

```  
const fluxChart = new k8s.helm.v2.Chart("flux",
            {
                repo: "weaveworks",
                chart: "flux",
                version: "0.3.3",
                namespace: "flux"
            }
        );
```

### LINKED ISSUE #255: Chart `version` should be optional
Helm does not require a version to be specified when installing a chart.  `helm.v2.Chart` should mark `version` as optional.

Note that this can be worked around by passing `version: <any>undefined`, since it is just an issue in the TypeScript typing not the JavaScript implementation.

### LINKED ISSUE #257: Helm: Allow installing charts simply by providing a fully qualified URL
With regular helm I can install charts without adding their corresponding custom chart repository first.
I.e.  this is valid helm usage: `helm install https://jupyterhub.github.io/helm-chart/jupyterhub-0.7.0.tgz --namespace jupyter --set proxy.secretToken=foobar` works.

The way how pulumi expects parameters to be passed it is currently always necessary to add a repository manually outside of pulumi which is cumbersome. E.g. via 
```
helm repo add jupyterhub https://jupyterhub.github.io/helm-chart/
helm repo update
```

Please add the ability to pass simple URLs to install remote charts.

lukehoban:
Fixes #255.
Fixes #229.
Fixes #257.

Align more closely with `helm` options for chart references to support all scenarios supported by Helm including (a) getting latest by passing no version (b) using explicitly repo URLs (c) providing fully qualified URLs to charts. 

All of the following now work:

```javascript
// 1. Explicit repo/chart/version - not technically part of `helm` format but needed for back compat
repo: "stable",
chart: "nginx-lego",
version: "0.3.1",

// 2. Reference to "latest"
repo: "stable",
chart: "nginx-lego",

// 3. Helm "chart reference"
chart: "stable/nginx-lego",

// 4. Helm "chart reference" with explicit version
chart: "stable/nginx-lego",
version: "0.3.1"

// 5. Helm "chart reference and repo url" with explicit version
chart: "nginx-lego",
version: "0.3.1",
fetchOpts: {
  repo: "https://kubernetes-charts.storage.googleapis.com"
},

// 6. Helm "chart reference and repo url" 
chart: "nginx-lego",
fetchOpts: {
  repo: "https://kubernetes-charts.storage.googleapis.com"
},

// 7. Helm "absolute URL" 
chart: "https://kubernetes-charts.storage.googleapis.com/nginx-lego-0.3.1.tgz",
```

### DIFF
--- a/pkg/gen/nodejs-templates/helm.ts.mustache
+++ b/pkg/gen/nodejs-templates/helm.ts.mustache
@@ -19,7 +19,7 @@ export namespace v2 {
     export interface ChartOpts extends BaseChartOpts {
         repo: pulumi.Input<string>;
         chart: pulumi.Input<string>;
-        version: pulumi.Input<string>;
+        version?: pulumi.Input<string>;
 
         fetchOpts?: pulumi.Input<FetchOpts>;
     }
--- a/pkg/gen/nodejs-templates/helm.ts.mustache
+++ b/pkg/gen/nodejs-templates/helm.ts.mustache
@@ -11,15 +11,15 @@ import * as nodepath from "path";
 
 export namespace v2 {
     interface BaseChartOpts {
-        /** 
+        /**
          * The optional namespace to install chart resources into.
          */
         namespace?: pulumi.Input<string>;
-        /** 
-         * Overrides for chart values. 
+        /**
+         * Overrides for chart values.
          */
         values?: pulumi.Inputs;
-        /** 
+        /**
          * Optional array of transformations to apply to resources that will be created by this chart prior to
          * creation. Allows customization of the chart behaviour without directly modifying the chart itself.
          */
@@ -53,7 +53,7 @@ export namespace v2 {
     }
 
     export interface LocalChartOpts extends BaseChartOpts {
-        /** 
+        /**
          * The path to the chart directory which contains the `Chart.yaml` file.
          */
         path: string;
--- a/pkg/gen/nodejs-templates/helm.ts.mustache
+++ b/pkg/gen/nodejs-templates/helm.ts.mustache
@@ -11,25 +11,51 @@ import * as nodepath from "path";
 
 export namespace v2 {
     interface BaseChartOpts {
+        /** 
+         * The optional namespace to install chart resources into.
+         */
         namespace?: pulumi.Input<string>;
+        /** 
+         * Overrides for chart values. 
+         */
         values?: pulumi.Inputs;
+        /** 
+         * Optional array of transformations to apply to resources that will be created by this chart prior to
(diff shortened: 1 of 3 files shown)


pkg/gen/nodejs-templates/helm.ts.mustache
```diff
@@ -11,25 +11,51 @@ import * as nodepath from "path";
 
 export namespace v2 {
     interface BaseChartOpts {
+        /**
```
hausdorff: Lots of trailing whitespace in this commit -- mind if we get rid of it?

lukehoban:
Is that really where that chart lives?  I get the same from `helm`:

```
$ helm fetch --repo https://istio.io/charts istio
Error: Failed to fetch https://istio.io/istio-1.0.1.tgz : 404 Not Found
```

lukehoban:
Looks like there is a bug in `helm` here where a missing trailing slash makes `helm fetch --repo url` not work, but the same missing slash is okay in `helm repo add url`.  I guess we can work around that bug and make sure there is always a trailing slash?

lukehoban:
Opened https://github.com/helm/helm/issues/4954 to track the issue in `helm`. 

I'm not sure it makes sense to workaround this in our layer - the `fetchOpts` are semantically exactly flags to `helm`, so trying to "fix" them up feels wrong.  

Were we to have done something here where we allowed this - I could imagine us adding the workaround to it - but we haven't added support for this form at all (because it isn't really something `helm` supports):

```
repo: "https://istio.io/charts",
chart: "istio",
```

Given that, I'll go ahead and merge this in as is for now.

PR #319, opened 2018-12-13
await: add PersistentVolumeClaim event processing to apps_deployment

metral:
Begins to address https://github.com/pulumi/pulumi-kubernetes/issues/133

### DIFF
--- a/pkg/await/apps_deployment.go
+++ b/pkg/await/apps_deployment.go
@@ -3,6 +3,7 @@ package await
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -88,6 +89,7 @@ type deploymentInitAwaiter struct {
 	config                 updateAwaitConfig
 	deploymentAvailable    bool
 	replicaSetAvailable    bool
+	pvcsAvailable          bool
 	updatedReplicaSetReady bool
 	currentGeneration      string
 
@@ -96,6 +98,7 @@ type deploymentInitAwaiter struct {
 	deployment  *unstructured.Unstructured
 	replicaSets map[string]*unstructured.Unstructured
 	pods        map[string]*unstructured.Unstructured
+	pvcs        map[string]*unstructured.Unstructured
 }
 
 func makeDeploymentInitAwaiter(c updateAwaitConfig) *deploymentInitAwaiter {
@@ -112,6 +115,7 @@ func makeDeploymentInitAwaiter(c updateAwaitConfig) *deploymentInitAwaiter {
 		deployment:  c.currentOutputs,
 		pods:        map[string]*unstructured.Unstructured{},
 		replicaSets: map[string]*unstructured.Unstructured{},
+		pvcs:        map[string]*unstructured.Unstructured{},
 	}
 }
 
@@ -133,7 +137,7 @@ func (dia *deploymentInitAwaiter) Await() error {
 	//      corresponding ReplicaSet), and therefore there is no rollout to mark as "Progressing".
 	//
 
-	replicaSetClient, podClient, err := dia.makeClients()
+	replicaSetClient, podClient, pvcClient, err := dia.makeClients()
 	if err != nil {
 		return err
 	}
@@ -164,15 +168,24 @@ func (dia *deploymentInitAwaiter) Await() error {
 	}
 	defer podWatcher.Stop()
 
+	// Create PersistentVolumeClaims watcher.
+	pvcWatcher, err := pvcClient.Watch(metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrapf(err,
+			"Could not create watcher for PersistentVolumeClaims objects associated with Deployment %q",
+			dia.config.currentInputs.GetName())
+	}
+	defer pvcWatcher.Stop()
+
 	period := time.NewTicker(10 * time.Second)
 	defer period.Stop()
 
-	return dia.await(deploymentWatcher, replicaSetWatcher, podWatcher, time.After(5*time.Minute), period.C)
+	return dia.await(deploymentWatcher, replicaSetWatcher, podWatcher, pvcWatcher, time.After(5*time.Minute), period.C)
 }
 
 func (dia *deploymentInitAwaiter) Read() error {
 	// Get clients needed to retrieve live versions of relevant Deployments, ReplicaSets, and Pods.
-	replicaSetClient, podClient, err := dia.makeClients()
+	replicaSetClient, podClient, pvcClient, err := dia.makeClients()
 	if err != nil {
 		return err
 	}
@@ -206,12 +219,19 @@ func (dia *deploymentInitAwaiter) Read() error {
 		podList = &unstructured.UnstructuredList{Items: []unstructured.Unstructured{}}
 	}
 
+	pvcList, err := pvcClient.List(metav1.ListOptions{})
+	if err != nil {
+		glog.V(3).Infof("Error retrieving PersistentVolumeClaims list for Deployment %q: %v",
+			deployment.GetName(), err)
+		pvcList = &unstructured.UnstructuredList{Items: []unstructured.Unstructured{}}
+	}
+
 	return dia.read(deployment, rsList.(*unstructured.UnstructuredList),
-		podList.(*unstructured.UnstructuredList))
+		podList.(*unstructured.UnstructuredList), pvcList.(*unstructured.UnstructuredList))
 }
 
 func (dia *deploymentInitAwaiter) read(
-	deployment *unstructured.Unstructured, replicaSets, pods *unstructured.UnstructuredList,
+	deployment *unstructured.Unstructured, replicaSets, pods, pvcs *unstructured.UnstructuredList,
 ) error {
 	dia.processDeploymentEvent(watchAddedEvent(deployment))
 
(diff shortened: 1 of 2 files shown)


hausdorff:
This looks sick, thanks for the PR!

Earlier we talked about the problem that `Deployment` will time out if PVs don't get bound, but won't tell the user what went wrong. This patch does make significant progress on that problem. Couple of questions/thoughts which I'm curious to hear your perspective on:

* It's clear to me that the patch as it stands has value for users. This would otherwise be a totally invisible error, and even if we did report errors correlated to the PV/PVC lines, it would not be obvious at all that this is what caused the `Deployment` to fail.
* Which errors should be correlated with which resource? For example, data in [`v1/Event`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#event-v1-core) usually contains information for _why_ the PV didn't mount, but which resources should report that reason? The PV/PVC? The `Deployment`s/`ReplicaSet`s/`Pod`s that reference it? Both?
* (#133 is basically proposing, to surface this data _somewhere_, but one thing this code makes obvious to me is that it is at least worth discussing _where_.)

Overall I think the answers we determine here will be answers we use for _every_ Kubernetes resource type, so it's worth considering in that context.

pkg/await/apps_deployment.go
```diff
@@ -164,15 +167,24 @@ func (dia *deploymentInitAwaiter) Await() error {
 	}
 	defer podWatcher.Stop()
 
+	// Create PersistentVolumeClaims watcher.
+	pvcWatcher, err := pvcClient.Watch(metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrapf(err,
+			"Could not create watcher for PersistentVolumeClaims objects associated with Deployment '%s'",
```
lblackstone: FYI, you can use `%q` to produce a quoted string rather than using `'%s'`

pkg/await/apps_deployment.go
```diff
@@ -503,6 +541,31 @@ func (dia *deploymentInitAwaiter) changeTriggeredRollout() bool {
...
+	allPVCsReady := true // bool used to do a rolling AND operation across all PVCs statuses
+	for _, pvc := range dia.pvcs {
+		phase, hasConditions := openapi.Pluck(pvc.Object, "status", "phase")
+		if !hasConditions {
+			return
+		}
+
+		// Success occurs when all PersistentVolumeClaims are marked as Bound
+		if phase == statusBound {
+			allPVCsReady = allPVCsReady && true
```
lblackstone: This appears to be a no-op. Should be able to simplify the conditional logic to ignore this branch.

pkg/await/apps_deployment.go
```diff
@@ -503,6 +541,31 @@ func (dia *deploymentInitAwaiter) changeTriggeredRollout() bool {
...
+		} else {
+			allPVCsReady = allPVCsReady && false
```
lblackstone: Not sure I understand this. This appears to be setting `allPVCsReady` to true only in the case where it was false to start. Rewriting the logic a bit produces:
```go
if phase != statusBound {
    allPVCsReady = !allPVCsReady
    ...
}
```

My guess is that you meant something like

```go
if phase != statusBound {
    allPVCsReady = false
    // Log errors ...
}
```

pkg/await/apps_deployment.go
```diff
@@ -580,6 +663,10 @@ func (dia *deploymentInitAwaiter) errorMessages() []string {
 	}
 
 	if dia.currentGeneration == "1" {
+		if !dia.pvcsAvailable {
+			messages = append(messages,
+				"Attepted to bind all PersistentVolumeClaims, but not all bindings were attained")
```
lblackstone: How about "Failed to bind PersistentVolumeClaim(s)"?

It would also be nice if we could print a list of PVCs that failed to bind.

pkg/await/apps_deployment.go
metral: Done! Thanks for the helpful tip

pkg/await/apps_deployment.go
metral: Ya, I'm not sure what I was thinking. I heeded your advice and simplified the logic - it should be straight forward now.

pkg/await/apps_deployment.go
metral: Good idea! I've rephrased the error message, and now include the list of PVCs that failed to bind.

metral:
@hausdorff 

> * Which errors should be correlated with which resource? For example, data in [`v1/Event`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#event-v1-core) usually contains information for _why_ the PV didn't mount, but which resources should report that reason? The PV/PVC? The `Deployment`s/`ReplicaSet`s/`Pod`s that reference it? Both?

IMO the most top-level owner object in the dependency graph should be bubbling up this error. That is, if we have a Deployment that contains a PVC (Deploy -> PVC), the Deployment object should inform the user of the error.

Similarly, if the top-level owner object in the dep graph is a Pod with a PVC (Pod -> PVC), or a stand-alone PVC itself (just PVC in the graph), each top object should notify the user of all errors pertaining to the PVC.

Thoughts?

metral:
> How about "Failed to bind PersistentVolumeClaim(s)"? It would also be nice if we could print a list of PVCs that failed to bind.

@hausdorff @lblackstone All feedback has been addressed, and now the error also includes which PVC failed - see below:

image


I also added tests for a Deploy -> PVC scenario, which was missing from the original PR.

PTAL

pkg/await/apps_deployment.go
```diff
@@ -403,6 +441,7 @@ func (dia *deploymentInitAwaiter) processDeploymentEvent(event watch.Event) {
 	}
 
 	dia.checkReplicaSetStatus()
+	dia.checkPersistentVolumeClaimStatus()
```
hausdorff: If we start bubbling up errors from other types, calling these `*Status` methods in the right places will start to get confusing. @lblackstone any chance we can simplify this a bit in your upcoming refactoring?

pkg/await/apps_deployment.go
```diff
@@ -503,6 +542,30 @@ func (dia *deploymentInitAwaiter) changeTriggeredRollout() bool {
...
+		phase, hasConditions := openapi.Pluck(pvc.Object, "status", "phase")
+		if !hasConditions {
+			return
+		}
+
+		// Success only occurs when there are no PersistentVolumeClaims
+		// defined, or when all PVCs have a status of 'Bound'
+		if phase != statusBound {
+			allPVCsReady = false
+			message := fmt.Sprintf("PersistentVolumeClaim: [%s] is not ready. status.phase currently at: %s", pvc.GetName(), phase)
```
hausdorff: We've historically tried to break these up into steps to make it seem sequential (_e.g._, `"[1/2] Waiting for app ReplicaSet be marked available"`). I'm curious to hear your thoughts on whether we should be doing that here, too? Where might it fit in?

pkg/await/apps_deployment.go
```diff
@@ -527,6 +590,26 @@ func (dia *deploymentInitAwaiter) processPodEvent(event watch.Event) {
...
+	if !isUnstructured {
+		glog.V(3).Infof("PersistentVolumeClaim watch received unknown object type %q",
+			reflect.TypeOf(pvc))
+		return
+	}
+
+	glog.V(3).Infof("Received update for PersistentVolumeClaim %q", pvc.GetName())
+
+	// If Pod was deleted, remove it from our aggregated checkers.
+	uid := string(pvc.GetUID())
```
hausdorff: Ha, I had forgotten about this. Historically we've been using `[namespace/]<name>` as the canonical ID, but maybe we should be using this instead everywhere. cc @lblackstone

pkg/await/apps_deployment.go
```diff
@@ -298,11 +329,18 @@ func (dia *deploymentInitAwaiter) await(
 func (dia *deploymentInitAwaiter) checkAndLogStatus() bool {
 	if dia.currentGeneration == "1" {
 		if dia.deploymentAvailable && dia.updatedReplicaSetReady {
+			if len(dia.pvcs) > 0 && !dia.pvcsAvailable {
```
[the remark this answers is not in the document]
hausdorff: Here and below in the other method.

pkg/await/apps_deployment.go
lblackstone: Yes, this is one of the main things I'd like to tackle next. Won't be in the client-go refactor, but that will enable a more state-machine based approach.

pkg/await/apps_deployment.go
lblackstone: Yeah, UID seems nice. Probably should open an issue to track.

pkg/await/apps_deployment.go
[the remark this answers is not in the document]
metral: I've rewritten the logic to be simpler and leverage reusable pieces. PTAL

pkg/await/apps_deployment.go
metral: Per @lblackstone's comment, it seems like this is a branch of work in and of itself. I'll table this feedback for the sake of this PR

hausdorff: Ah, to be clear, I'd meant mainly to bring this to Levi's attention rather than to ask you to do it. :)

pkg/await/apps_deployment.go
metral: Hmm great point. Thinking about this more, many other API resources can be embedded within a Deployment other than a PVC: e.g. a ConfigMap, Secret etc. This tells me we should go back to the drawing board to think on how to know of all dependent pieces of a Deployment up front, so that we can track each object required, as well as programmatically form the progress message (e.g. `[1/2]`...`[2/5]` etc.) accordingly to the total number of objects we're waiting on.

Given this, it seems out of scope to hard-code a number of objects for PVCs in this PR, so I'll file this as a separate issue to track.

pkg/await/apps_deployment.go
metral: Ya, I went with UID to avoid any unforeseen `namespace/name` conflicts

metral:
Squashed commits

PR #322, opened 2018-12-18
Add CHANGELOG.md

hausdorff:


ellismg:
Looks like a log of changes to me.  If there are interesting things we forgot to add, we can always go back and add them!

PR #324, opened 2018-12-19
Fix replacement logic in `Job`

### LINKED ISSUE #316: Field is immutable error when Helm updates jobs
I'm deploying Consul and Wordpress via helm, both of these charts use jobs to test the deployment. When running a new deployment Pulumi is issuing an update to the job definition and k8s responds with a field is immutable warning. This should be fixable by using a replace.

hausdorff:
Fixes #316.

### DIFF
--- a/pkg/provider/diff.go
+++ b/pkg/provider/diff.go
@@ -62,6 +62,11 @@ var forceNew = groups{
 			},
 		},
 	},
+	"batch": versions{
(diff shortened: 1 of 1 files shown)


hausdorff:
I'm going to go ahead and merge this because it's so simple and the issue it addresses is P1. @lukehoban let's follow up if you have issues.

PR #328, opened 2018-12-20
Allow `yaml` package to understand URLs

### LINKED ISSUE #323: apiextensions.CustomResource#get is missing

hausdorff:


sdk/nodejs/provider.ts
```diff
@@ -2222,7 +2247,14 @@ export namespace yaml {
     export class ConfigFile extends CollectionComponentResource {
         constructor(name: string, config?: ConfigFileOpts, opts?: pulumi.ComponentResourceOptions) {
             super("kubernetes:yaml:ConfigFile", name, config, opts);
-            const text = fs.readFileSync(config && config.file || name).toString();
+            const fileId = config && config.file || name
+            let text = "";
+            if (fileId.startsWith("http://") || fileId.startsWith("https://")) {
```
lukehoban: What is the full set of rules that `kubectl` uses for interpreting an argument to `-f`?  Are we doing the same?

sdk/nodejs/provider.ts
```diff
@@ -2222,7 +2247,14 @@ export namespace yaml {
...
+                text = syncreq.default("GET", fileId).getBody().toString()
```
lukehoban: I *think* we can pretty easily incorporate async IO here, which would be better if it doesn't negatively impact usabilty.  Specifically - can't you do something like this:

```
export class ConfigFile extends CollectionComponentResource {
        constructor(name: string, config?: ConfigFileOpts, opts?: pulumi.ComponentResourceOptions) {
            super("kubernetes:yaml:ConfigFile", name, config, opts);
            const textP = fs.readFilePromise(config && config.file || name);
            const objsP = textP.then(text => jsyaml.safeLoadAll(text.toString()));
            this.resources = pulumi.output(objs).apply(objs => parseYamlDocument({
                objs: objs,
                transformations: config && config.transformations || []
            }, {parent: this}));
        }
    }
```

pkg/gen/nodejs-templates/provider.ts.mustache
```diff
@@ -174,10 +197,16 @@ export namespace yaml {
...
-            const text = fs.readFileSync(config && config.file || name).toString();
-            const objs = jsyaml.safeLoadAll(text);
-            this.resources = pulumi.output(parseYamlDocument({
-                objs: objs,
+            const fileId = config && config.file || name
+            let text: Promise<string>;
+            if (fileId.startsWith("http://") || fileId.startsWith("https://")) {
+                text = got(fileId).then(r => r.body)
+            } else {
+                text = new Promise<string>(resolve => resolve(fs.readFileSync(fileId).toString()))
```
lukehoban: This is fine - but:
1. You could do the same thing with just `Promise.resolve(fs.readFileSync(fileId).toString())`.
2. You sort of may as well do async file I/O here too at this point

PR #331, opened 2018-12-28
Add container and restartPolicy properties to Job diff logic

### LINKED ISSUE #330: Replace logic on `Job` is incorrect
_cf_.

```
Job.batch "api-migration-job-zdq5ov6o" is invalid: spec.template: Invalid value: [job spec]: field is immutable
```

iagomelanias:
Closes #330. Following [this comment](https://github.com/kubernetes/kubernetes/issues/48388#issuecomment-320975927) and a lot of local tests, both `containers` and `restartPolicy` cannot be changed after it's creation. Instead, Pulumi should replace them.

This PR fixes this by adding two new rules:
* `.spec.template.spec.containers[*]`
* `.spec.template.spec.restartPolicy`

### DIFF
--- a/pkg/provider/diff.go
+++ b/pkg/provider/diff.go
@@ -106,6 +106,8 @@ var deployment = properties{
 
 var job = properties{
 	".spec.selector",
+	".spec.template.spec.containers[*]",
+	".spec.template.spec.restartPolicy",
 }
 
 var statefulSet = properties{


iagomelanias:
Some notes:
* I dind't find any documentation about which fields are immutable and which are not. 🤨I'm pretty sure there are more immutable fields, so we should test every field manually or look at Kubernetes codebase in the future.
* I'm don't know why the tests are failing.

hausdorff:
The CI will fail because we use secrets to run tests, and you travis only gives you access to secrets if you are pushing to the repository. We should turn off integration tests for external PRs -- sorry about that, I'll do it when I get back.

Have you tested this on real objects? If so I'm happy to merge it anyway, since it should not have impact on any other resource types.

hausdorff:
And yes, unfortunately, we so far have had to depend on experimentation and intuition re: which fields are immutable. :) I'm working with upstream on this.

iagomelanias:
That makes sense. I wasn't able to build/test it locally, some commands fail (probably because i'm not using Linux).

iagomelanias:
Is there any test that can potentially fail after this change? If there are tests covering the diff checker, we can merge to a temporary PR before merging to master.

I found [some tests](https://github.com/pulumi/pulumi-kubernetes/blob/master/pkg/provider/diff_test.go), but i'm not sure if this change is going to be covered.

hausdorff:
@iagomelanias No, there should be no tests that can fail as a result of this, no. We don't have tests asserting which properties are immutable because there is no recorded ground truth in the OpenAPI spec -- it's basically a list of accumulated operational knowledge.

If you can't build it and test locally I'll do it myself when I have time -- it may have to be after Jan 2nd, which is when I return to work. Apologies for the delay.

iagomelanias:
No problem. I'll try to setup a Docker image to build/test locally. 
Sorry for taking your time on vacation. 😐

hausdorff:
@iagomelanias We haven't quite nailed down the PR flow for outside contributors yet, so for now, I'll re-submit the PR from a local branch, but you'll still get credit for the commit. :) Thanks again for the contrib.

Issue #332, opened 2019-01-03
Allow external contributors


hausdorff:
For security reasons Travis does not expose secrets to PRs from orgs and accounts that are not `pulumi-kubernetes`. This means that CI will fail for all external PRs, _cf_., #331. We should simply disable integration tests for external PRs.

PR #333, opened 2019-01-03
Fix replacement logic in `Job`

### LINKED ISSUE #330: Replace logic on `Job` is incorrect

hausdorff:
Fixes #330.

Submitted on behalf of @iagomelanias because our CI system currently does not support external contributors (#332). Previously this PR was #331.

### DIFF
--- a/pkg/provider/diff.go
+++ b/pkg/provider/diff.go
@@ -106,6 +106,7 @@ var deployment = properties{
 
 var job = properties{
 	".spec.selector",
+	".spec.template",
 }
 
 var statefulSet = properties{


hausdorff:
Per offline conversation with @iagomelanias, we've decided to change the original PR to make `.spec.template` a force-new property, based on our understanding of the spec, and based on this error we get when you change a property somewhere deep in the `Job`, as it seems to be saying that `.spec.template` is an immutable field:

```
The Job "pi" is invalid: spec.template: Invalid value: core.PodTemplateSpec{...}: field is immutable
```

PR #469, opened 2019-03-05
flow downstream

CyrusNajmabadi:


CyrusNajmabadi:
The failure is ok.  It's just the nature of needing to flow these changes down.

CyrusNajmabadi:
@hausdorff Rerunning, i now see:

```
[                          step1 ]     pulumi:pulumi:Stack get-tests-p-it-travis-job-step1-992b8e49  	/home/travis/gopath/src/github.com/pulumi/pulumi-kubernetes/vendor/google.golang.org/grpc/server.go:683 +0xa1
[                          step1 ]  
[                          step1 ] Diagnostics:
[                          step1 ]   pulumi:pulumi:Stack (get-tests-p-it-travis-job-step1-992b8e49):
[                          step1 ]     panic: interface conversion: runtime.Object is *unstructured.UnstructuredList, not *unstructured.Unstructured
[                          step1 ]     goroutine 37 [running]:
[                          step1 ]     github.com/pulumi/pulumi-kubernetes/vendor/k8s.io/client-go/dynamic.(*dynamicResourceClient).Get(0xc0005ef2c0, 0xc0000b87b8, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
[                          step1 ]     	/home/travis/gopath/src/github.com/pulumi/pulumi-kubernetes/vendor/k8s.io/client-go/dynamic/simple.go:231 +0x905
[                          step1 ]     github.com/pulumi/pulumi-kubernetes/pkg/await.Read(0x16ff2a0, 0xc000556900, 0xc0005e7ec0, 0xc00004a980, 0x76, 0xc0003eb950, 0xc00000e048, 0xc0000b87b8, 0x0, 0x0, ...)
[                          step1 ]     	/home/travis/gopath/src/github.com/pulumi/pulumi-kubernetes/pkg/await/await.go:169 +0x159
[                          step1 ]     github.com/pulumi/pulumi-kubernetes/pkg/provider.
```

This seems to be a crash in something related to 'await.go'.  Do you know what's up here?

hausdorff:
@CyrusNajmabadi Strangely, this error seems to be coming from _inside_ kubernetes/client-go. No combination of arguments should cause `ResourceInterface#Get` to panic, so I'm guessing the "solution" is going to be to upgrade the package.

CyrusNajmabadi:
@hausdorff Is that something you could do in another PR?  Thanks!

hausdorff:
Hmm, I can't reproduce this locally from your PR. I'm not sure what's going on yet.

CyrusNajmabadi:
No worries.!

hausdorff:
The problem is that in stage2, this:

```
    id: `${idNamespace}/my-new-cron-object`,
```

needs to become this:

```
    id: idNamespace.apply(ns => `${ns}/my-new-cron-object`),
```

hausdorff:
Or, more specifically, this `idNamespace` in the code snippet below:

```
`${idNamespace}/my-new-cron-object`
```

will evaluate to an error message, which means that this string will evaluate to the following. Notice the `/my-new-cron-object` at the end of this error -- that's the name of the object that's supposed to go after `idNamespace` above.

```
Calling [toString] on an [Output<T>] is not supported.
    To get the value of an Output<T> as an Output<string> consider either:
    1: o.apply(v => `prefix${v}suffix`)
    2: pulumi.interpolate `prefix${v}suffix`
    See https://pulumi.io/help/outputs for more details.
    This function may throw in a future version of @pulumi/pulumi./my-new-cron-object
```

Apparently when we pass this beauty of a "namespace" to the client, it causes client-go to totally flip out and panic because it doesn't know what to do with that string, and it generates an `*unstructured.UnstructuredList` instead of `*unstructured.Unstructured`. Eek!

cc @lblackstone I won't get to it tonight, but we'll probably need to try upgrading client-go and then if that doesn't work, file a bug against them. I'll start by filing an issue.

hausdorff:
Alright I've filed the bug: #470. In the mean time, @CyrusNajmabadi, you should change that line, and that should make everything good to go.

Issue #494, opened 2019-03-21
Method handler checkConfig for /pulumirpc.ResourceProvider/CheckConfig expected but not provided


swgillespie:
Seen occasionally after previews:

```
Diagnostics:
  pulumi:pulumi:Stack (aws-ts-eks-stream):
    Method handler checkConfig for /pulumirpc.ResourceProvider/CheckConfig expected but not provided
    Method handler diffConfig for /pulumirpc.ResourceProvider/DiffConfig expected but not provided
```

PR #495, opened 2019-03-21
Stub out DiffConfig and CheckConfig

swgillespie:
This commit stubs out two new RPC endpoints expected by the engine. The
new stubs are functionally equivalent to the previous behavior where the
endpoints were not provided, except that gRPC will no longer warn when
the endpoints are used.

This commit also updates the reference to pulumi/pulumi to take the new
gRPC interface changes.


swgillespie:
Fixes https://github.com/pulumi/pulumi-kubernetes/issues/494.

PR #580, opened 2019-06-03
Remove `getInputs` from Kubernetes resources

hausdorff:


hausdorff:
It's related to #566 but it does not come close to fixing it.

PR #634, opened 2019-07-13
Don't render emoji on Windows.

### LINKED ISSUE #159: Checkmarks during Updates don't render on Windows
When you run `pulumi up` on Windows, the emjois displayed in the console as part of the status rich updates don't render correctly on a default windows install (you end up with question-mark boxes)

We should just not print emjoi's on windows.  We already do this elsewhere in the CLI, but I assume these messages are coming from the kubernetes provider and we'll need a change here as well.

Here's a screenshot:

image

hausdorff:
Fixes #159.

### DIFF
--- a/pkg/await/apps_deployment.go
+++ b/pkg/await/apps_deployment.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pulumi/pulumi/pkg/util/cmdutil"
+
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 	"github.com/pulumi/pulumi-kubernetes/pkg/await/states"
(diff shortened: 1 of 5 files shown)


pkg/await/apps_deployment.go
```diff
@@ -342,7 +344,8 @@ func (dia *deploymentInitAwaiter) checkAndLogStatus() bool {
 				return false
 			}
 
-			dia.config.logStatus(diag.Info, "✅ Deployment initialization complete")
+			dia.config.logStatus(diag.Info,
+				fmt.Sprintf("%s Deployment initialization complete", cmdutil.EmojiOr("✅", "")))
```
pgavlin:
```suggestion
				fmt.Sprintf("%sDeployment initialization complete", cmdutil.EmojiOr("✅ ", "")))
```

Same feedback throughout

pgavlin: (assuming that works and the format string parser doesn't choke on `%sDeployment`)

PR #636, opened 2019-07-13
Emit useful error when `kubernetes:config:kubeconfig` fails to parse

### LINKED ISSUE #209: Problem parsing kubeconfig when passed via "kubernetes:kubeconfig"

hausdorff:
Fixes #209.

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -27,6 +27,7 @@ import (
 	"github.com/golang/glog"
 	pbempty "github.com/golang/protobuf/ptypes/empty"
 	structpb "github.com/golang/protobuf/ptypes/struct"
+	"github.com/pkg/errors"
 	pkgerrors "github.com/pkg/errors"
 	"github.com/pulumi/pulumi-kubernetes/pkg/await"
 	"github.com/pulumi/pulumi-kubernetes/pkg/clients"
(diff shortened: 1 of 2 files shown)


pkg/provider/provider.go
```diff
@@ -234,7 +234,9 @@ func (k *kubeProvider) Configure(_ context.Context, req *pulumirpc.ConfigureRequ
 	if configJSON, ok := vars["kubernetes:config:kubeconfig"]; ok {
 		config, err := clientcmd.Load([]byte(configJSON))
 		if err != nil {
-			return nil, fmt.Errorf("failed to parse kubeconfig: %v", err)
+			return nil, pkgerrors.Wrap(err, "failed to parse kubeconfig data in "+
```
pgavlin: Does a YAML literal work here as well as JSON? Also, the wording is a bit informal; I'd kill "and not (say) a filename".

hausdorff: Updated the wording, but I think it's fine to not bother to clarify that JSON is a strict subset of YAML because (1) "YAML" in the k8s ecosystem is used, basically, as shorthand for "JSON or YAML", (20 I've actually never seen someone write kubeconfig in JSON (3) if you do, `kubectl` auto-converts it to YAML the first time it changes it for any reason, and (4) the ecosystem, IME, largely understands that JSON is a subset of YAML anyway.

PR #639, opened 2019-07-13
Codegen fixes for the Python SDK

### LINKED ISSUE #602: Change `__opts__` to `opts` to align with tfgen providers
In Python, the `ResourceOptions` parameter name on resource constructors for tfgen-backed providers has been `opts` for a while (though `__opts__` is still supported for backwards compatibility). We should roll this change through to the Kubernetes provider also.

hausdorff:
Fixes #602.

### DIFF
--- a/pkg/gen/python-templates/kind.py.mustache
+++ b/pkg/gen/python-templates/kind.py.mustache
@@ -3,18 +3,25 @@
 
 import pulumi
 import pulumi.runtime
+import warnings
 
 from ... import tables, version
 
 
 class {{Kind}}(pulumi.CustomResource):
     {{{Comment}}}
-    def __init__(self, __name__, __opts__=None{{#Properties}}, {{LanguageName}}=None{{/Properties}}):
-        if not __name__:
+    def __init__(self, resource_name, opts=None{{#Properties}}, {{LanguageName}}=None{{/Properties}}, __name__=None, __opts__=None):
+        if __name__ is not None:
+            warnings.warn("explicit use of __name__ is deprecated", DeprecationWarning)
+            resource_name = __name__
+        if __opts__ is not None:
+            warnings.warn("explicit use of __opts__ is deprecated, use 'opts' instead", DeprecationWarning)
+            opts = __opts__
+        if not resource_name:
             raise TypeError('Missing resource name argument (for URN creation)')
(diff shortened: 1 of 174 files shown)


pkg/gen/python-templates/kind.py.mustache
```diff
@@ -9,12 +9,15 @@ from ... import tables, version
...
     {{{Comment}}}
-    def __init__(self, __name__, __opts__=None{{#Properties}}, {{LanguageName}}=None{{/Properties}}):
+    def __init__(self, __name__, opts=None{{#Properties}}, {{LanguageName}}=None{{/Properties}}, __opts__=None):
         if not __name__:
             raise TypeError('Missing resource name argument (for URN creation)')
         if not isinstance(__name__, str):
             raise TypeError('Expected resource name to be a string')
-        if __opts__ and not isinstance(__opts__, pulumi.ResourceOptions):
+        if __opts__ is not None:
+            warnings.warn("explicit use of __opts__ is deprecated, use 'opts' instead", DeprecationWarning)
```
pgavlin: Do we actually render these deprecation warnings in the CLI? I thought we had some issue around this...

hausdorff: I just copied what the TF bridge providers do because it seemed weirder to not do what they did in the same situation. We can do something else if it makes more sense.

pgavlin: This seems great, then :+1:

lukehoban: If we are going to match the TF providers, we should also consider renaming `__name__` to `resource_name`:

https://github.com/pulumi/pulumi-aws/blob/master/sdk/python/pulumi_aws/athena/database.py#L28

hausdorff: We didn't make this other breaking change because it wasn't listed in the original issue, and we therefore didn't know about it.

Is this all the changes we need to make here?

hausdorff: @lukehoban @pgavlin I've updated to include the breaking `__name__` -> `resource_name` change as well. Please let me know if there are other things we need to put in here ASAP, as we're planning on merging soon.

PR #647, opened 2019-07-23
Use Python Codegen package from pulumi/pulumi

stack72:


stack72:
@hausdorff yes, I ran it and no code changes here - travis build going green has verified this as well!

PR #658, opened 2019-07-26
Return to diffing inputs for resources w/o dry run.

### LINKED ISSUE #653: Setting `.stringData` in `v1/Secret` results in Pulumi triggering a replace on every update
The following Pulumi program will faithfully report on every single `pulumi up` that the `stringData` member has been added to this object, and therefore the `Secret` has to be replaced.

```typescript
const testSecrets = new k8s.core.v1.Secret("test", {
    type: "Opaque",
    stringData: {
        "client-secret": "hello",
    },
});

```

Checking the live state of this, we see that `stringData` is missing. This is because the API server takes `stringData` and does the base64 conversion automatically, and uses that to populate the `data` field:

```yaml
apiVersion: v1
data:
  client-secret: aGVsbG8=
kind: Secret
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Secret","metadata":{"annotations":{"pulumi.com/autonamed":"true"},"labels":{"app.kubernetes.io/managed-by":"pulumi"},"name":"test-pyooxt59"},"stringData":{"client-secret":"hello"},"type":"Opaque"}
    pulumi.com/autonamed: "true"
  creationTimestamp: "2019-07-25T05:55:04Z"
  labels:
    app.kubernetes.io/managed-by: pulumi
  name: test-pyooxt59
  namespace: default
  resourceVersion: "366196"
  selfLink: /api/v1/namespaces/default/secrets/test-pyooxt59
  uid: bff1d5f7-aea0-11e9-aff3-42010a800013
type: Opaque
```

So this seems like a problem with the new diff logic generally.

pgavlin:
Due to normalization performed by the API server, the recent changes to
use a kubectl-style merge patch approach when diffing resources often
results in spurious diffs. This interacts particularly badly with
resources like `Secret` where the spurious diffs cause unnecessary
resource replacements. These changes return to the earlier input-based
diff logic for resources and/or API servers that do not support dry-run.

Fixes #653.

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -45,9 +45,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/client-go/tools/clientcmd"
 	clientapi "k8s.io/client-go/tools/clientcmd/api"
 )
@@ -522,7 +520,7 @@ func (k *kubeProvider) Diff(
 	if err != nil {
 		return nil, err
 	}
-	oldInputs, oldLiveState := parseCheckpointObject(oldState)
+	oldInputs, _ := parseCheckpointObject(oldState)
 
 	// Get new resource inputs. The user is submitting these as an update.
 	newResInputs, err := plugin.UnmarshalProperties(req.GetNews(), plugin.MarshalOptions{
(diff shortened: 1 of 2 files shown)


pkg/provider/provider.go
```diff
@@ -586,25 +571,28 @@ func (k *kubeProvider) Diff(
...
-			oldLiveState = live
+			isInputPatch = true
+			patch, err = k.inputPatch(oldInputs, newInputs)
+			patchBase = oldInputs
 		}
 	} else {
-		patch, err = k.localPatch(oldInputs, newInputs, oldLiveState)
+		isInputPatch = true
+		patch, err = k.inputPatch(oldInputs, newInputs)
+		patchBase = oldInputs
```
hausdorff: Just to protect future, tiny-brained Alex, can we please add a `.V(1)` log in the appropriate spot after this block to indicating which patch type was used to calculate the diff? I'd rather have that information than not when users file issues.

pkg/provider/provider.go
```diff
@@ -565,19 +563,6 @@ func (k *kubeProvider) Diff(
...
-	if oldAnnotations := oldLiveState.GetAnnotations(); oldAnnotations != nil {
-		if lastAppliedConfig, ok := oldAnnotations[lastAppliedConfigKey]; ok {
-			annotations := newInputs.GetAnnotations()
-			if annotations == nil {
-				annotations = make(map[string]string)
-			}
-			annotations[lastAppliedConfigKey] = lastAppliedConfig
-			newInputs.SetAnnotations(annotations)
-		}
-	}
```
hausdorff: Is this not necessary because of this patch, or because of a previous change? If the first, why is it not needed anymore?

pkg/provider/provider.go
pgavlin: Yes.

pkg/provider/provider.go
pgavlin: This was necessary when we were calculating the patch locally using strategic or three-way JSON merge. This code was used to prevent diffs in the last-applied-config annotation. The new code does not require this because:
- In the case of dry run, the server will handle this itself
- In the case of input diffing, neither the old or new inputs contain this annotation, so adding it creates rather than prevents a diff

pkg/provider/provider.go
pgavlin: Done.

PR #662, opened 2019-07-29
Update pulumi-kubernetes to use Pulumi fork of glog

jen20:
This fixes pulumi/pulumi#2989, and removes any module replacements which are not directly related to Kubernetes.

Fixes pulumi/pulumi#2989.


jen20:
I don't think we actually need to address this yet, and should come up with a better solution for when we do.

PR #841, opened 2019-10-11
Examples: Change the examples to run as individual tests

stack72:


stack72:
Sorry @lblackstone, I thought .gitignore would have stopped this - my bad for not checking!

Issue #863, opened 2019-10-25
.NET support for Kubernetes


lukehoban:
We are adding support for .NET to [Pulumi core](https://github.com/pulumi/pulumi/pull/3399) and to [pulumi-terraform-bridge](https://github.com/pulumi/pulumi-terraform-bridge/pull/54).  We should also add support for .NET to the Kubernetes provider.

- [x] Core Provider Code Generation (https://github.com/pulumi/pulumi-kubernetes/pull/885)
- [ ] YAML/Helm support (#957)

lukehoban:
Here's an example based on the in-progress PR: https://github.com/pulumi/examples/pull/477

kenlefeb:
Is the `version.txt` file supposed to be omitted? Do you have a sample of the contents, in case I want to build this package locally?

lukehoban:
> in case I want to build this package locally?

`make build install` will build the package locally.  The `version.txt` file is written into the source package as part of the build.

mikhailshilkov:
I'll close this issue to track Helm support in #957

PR #885, opened 2019-11-19
Add .NET code generation

lukehoban:
Adds core code-generation support for .NET to the Kubernetes provider.

```cs
var pod = new Pulumi.Kubernetes.Core.V1.Pod("pod", new Pod
{
    Spec = new PodSpec
    {
        Containers = 
        {
            new Container
            {
                Name = "nginx",
                Image = "nginx",
            },
        },
    },
});
```

Part of https://github.com/pulumi/pulumi-kubernetes/issues/863.


pkg/gen/dotnet-templates/typesInput.cs.mustache
```diff
@@ -0,0 +1,63 @@
+// *** WARNING: this file was generated by the Pulumi Kubernetes client generation tool. ***
+// *** Do not edit by hand unless you're certain you know what you are doing! ***
+
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Pulumi.Serialization;
+
+{{#Groups}}
+namespace Pulumi.Kubernetes.Types.Inputs.{{Group}} {
```
lukehoban: This doesn't quite feel right - but we've diverged a fair bit on where we place these types across the .NET/TypeScript and TF/Kubernetes providers - so I'm not sure what the "right" answer is here.

Today - we have:
* TS+TF: `aws.types.input.ec2.InstanceEbsBlockDevice`
* CS+TF: `Pulumi.Aws.Ec2.Inputs.InstanceEbsBlockDevicesArgs`
* TS+Kubernetes: `kubernetes.types.input.core.v1.Pod`

By triangulating those - I think you end up with: `Pulumi.Kubernetes.Core.V1.Inputs.PodArgs` instead of the current `Pulumi.Kubernetes.Types.Input.Core.V1.PodSpec`. 

Does that wound right?  Is there a reason we were inconsistent with TypeScript on where these types live in our .NET codegen for TF?

cc @pgavlin @lblackstone @mikhailshilkov

lukehoban:
Current build failures are due to the issue in the integration test framework being fixed in https://github.com/pulumi/pulumi/pull/3527.

pkg/gen/dotnet-templates/typesInput.cs.mustache
lukehoban: I've landed on `Pulumi.Kubernetes.Types.Input.Core.V1.PodSpecArgs` for now.  Open to feedback.

lukehoban:
@lblackstone @CyrusNajmabadi @mikhailshilkov @hausdorff This now ready for review.  Tests are passing for the [Guestbook](https://github.com/pulumi/pulumi-kubernetes/pull/885/files#diff-526e0c0ad738a0a7d1292c6e66403cb3) example.

pkg/gen/dotnet-templates/typesInput.cs.mustache
lblackstone: That looks right to me. The CS+TF example appears to be inconsistent with the existing layout (k8s.types.[input,output].[group].[version].[kind])

pkg/gen/dotnet-templates/kind.cs.mustache
```diff
@@ -0,0 +1,68 @@
...
+        ///
+        /// <param name="name">The unique name of the resource</param>
+        /// <param name="args">The arguments used to populate this resource's properties</param>
+        /// <param name="options">A bag of options that control this resource's behavior</param>
+        public {{Kind}}(string name, Types.Inputs.{{Group}}.{{Version}}.{{Kind}}Args? args = null, CustomResourceOptions? options = null)
+            : base("kubernetes:{{URNAPIVersion}}:{{Kind}}", name, SetAPIKindAndVersion(args), MakeResourceOptions(options, ""))
+        {
+        }
+
+        private static ResourceArgs? SetAPIKindAndVersion(Types.Inputs.{{Group}}.{{Version}}.{{Kind}}Args? args) {
```
mikhailshilkov: I believe we should never send `null` to the base class arguments. Use `Empty` for the default case.

pkg/gen/dotnet-templates/kind.cs.mustache
```diff
@@ -0,0 +1,68 @@
+// *** WARNING: this file was generated by the Pulumi Kubernetes codegen tool. ***
+// *** Do not edit by hand unless you're certain you know what you are doing! ***
+
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Pulumi.Serialization;
+
+namespace Pulumi.Kubernetes.{{Group}}.{{Version}} {
```
mikhailshilkov: I would put all opening curly braces on a new line according to C# convensions

pkg/gen/dotnet-templates/kind.cs.mustache
```diff
@@ -0,0 +1,68 @@
...
+        {{{Comment}}}
+        [Output("{{Name}}")]
+        public {{{PropType}}} {{LanguageName}} { get; private set; } = null!;
+
+        {{/Properties}}
+
+        /// <summary>
+        /// Create a {{Kind}} resource with the given unique name, arguments, and options.
+        /// </summary>
+        ///
```
mikhailshilkov: Nit: no need to add an empty line between summary and params, here and in other places

tests/integration/dotnet/basic/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: basic_dotnet_kubernetes
+description: A simple .NET Kubernetes program.
```
mikhailshilkov: Is that a basic, simple K8s program? 😄

cmd/pulumi-gen-kubernetes/main.go
```diff
@@ -216,3 +218,54 @@ func writePythonClient(data map[string]interface{}, outdir, templateDir string)
...
+		panic(err)
+	}
+
+	typesDir := fmt.Sprintf("%s/Types", outdir)
+	err = os.MkdirAll(typesDir, 0700)
+	if err != nil {
+		panic(err)
+	}
+
+	err = ioutil.WriteFile(fmt.Sprintf("%s/Input.cs", typesDir), []byte(inputAPIcs), 0777)
```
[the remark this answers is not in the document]
lukehoban: This aligns with how we do code generation for TS and Py.  It's just as bad for those two - so I think I'll leave as-is for now and open a separate follow-up to clean up this codegen across the board.

pkg/gen/dotnet-templates/kind.cs.mustache
lukehoban: Should we fix the base class constructor to not accept null?

cmd/pulumi-gen-kubernetes/main.go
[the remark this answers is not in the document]
lukehoban: https://github.com/pulumi/pulumi-kubernetes/issues/888

sdk/dotnet/AdmissionRegistration/V1/MutatingWebhookConfiguration.cs
```diff
@@ -0,0 +1,94 @@
...
+        public MutatingWebhookConfiguration(string name, Types.Inputs.AdmissionRegistration.V1.MutatingWebhookConfiguration? args = null, CustomResourceOptions? options = null)
+            : base("kubernetes:admissionregistration.k8s.io/v1:MutatingWebhookConfiguration", name, SetAPIKindAndVersion(args), MakeResourceOptions(options, ""))
+        {
+        }
+
+        private static ResourceArgs? SetAPIKindAndVersion(Types.Inputs.AdmissionRegistration.V1.MutatingWebhookConfiguration? args) {
+            if (args != null) {
+                args.ApiVersion = "admissionregistration.k8s.io/v1";
+                args.Kind = "MutatingWebhookConfiguration";
+            }
```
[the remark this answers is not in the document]
CyrusNajmabadi: i don't mind 'heavyweght' extra work we do that the user won't see.  it's not like they'd ever be aware of this or feel any perf hit here (compared to the IO and costs of the actual K work).  So i'm ok with just generating internally and doing the copies.

pkg/gen/dotnet-templates/Pulumi.Kubernetes.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <PackageProjectUrl>https://pulumi.com</PackageProjectUrl>
+    <RepositoryUrl>https://github.com/pulumi/pulumi-kubernetes</RepositoryUrl>
+
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn>1701;1702;1591;8604;8625</NoWarn>
```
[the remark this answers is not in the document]
CyrusNajmabadi: 👍

CyrusNajmabadi:
Skimmed over this and nothing looks out of place or concerning.  If you have any particular areas you'd like me to look more deeply at, just call them out.  But, otherwise, if this works then this works :)

pkg/gen/typegen.go
```diff
@@ -530,14 +560,189 @@ func makePythonType(resourceType, propName string, prop map[string]interface{},
...
+
+	isSimpleRef := true
+	switch ref {
+	case quantity:
+		ref = "string"
+	case intOrString:
+		return oneOf("int", "string")
+	case v1Fields, v1FieldsV1, rawExtension:
+		// TODO: This is quite possibly wrong - these are actually JSON objects
+		ref = "string"
```
lukehoban: @CyrusNajmabadi This and a little further below are two places where the type really should be `object` - but we can't represent this.  The JSON Schema for these allows any JSON object whatsoever.  Is it possible to represent this in our .NET projection?

pkg/gen/dotnet-templates/kind.cs.mustache
mikhailshilkov: It looks like I'm wrong and it's only `Resource` that doesn't accept them https://github.com/pulumi/pulumi/blob/394c91d7f6ab7a4096f4454827690a460f665433/sdk/dotnet/Pulumi/Resources/CustomResource.cs#L34

sdk/dotnet/Apps/V1/Deployment.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// <summary>
+    /// Deployment enables declarative updates for Pods and ReplicaSets.
+    /// 
+    /// This resource waits until its status is ready before registering success
+    /// for create/update, and populating output properties from the current state of the resource.
+    /// The following conditions are used to determine whether the resource creation has
+    /// succeeded or failed:
+    /// 
+    /// 1. The Deployment has begun to be updated by the Deployment controller. If the current
+    ///    generation of the Deployment is &amp;gt; 1, then this means that the current generation
```
lblackstone: Is `&amp;gt;` correct here?

sdk/dotnet/Apps/V1/Deployment.cs
```diff
@@ -0,0 +1,122 @@
...
+
+
+        /// <summary>
+        /// Create a Deployment resource with the given unique name, arguments, and options.
+        /// </summary>
+        /// <param name="name">The unique name of the resource</param>
+        /// <param name="args">The arguments used to populate this resource's properties</param>
+        /// <param name="options">A bag of options that control this resource's behavior</param>
+        public Deployment(string name, Types.Inputs.Apps.V1.DeploymentArgs? args = null, CustomResourceOptions? options = null)
+            : base("kubernetes:apps/v1:Deployment", name, SetAPIKindAndVersion(args), MakeResourceOptions(options, ""))
```
lblackstone: It might not matter at this point, but for TS, Alex split the type string out into a separate variable for use with query features.

sdk/nodejs/yaml/yaml.ts
```diff
@@ -590,72 +590,72 @@ import * as outputs from "../types/output";
          * For example:
          *     getResourceProperty("v1/Service", "nginx", "spec")
          */
-        public getResourceProperty(groupVersionKind: "admissionregistration.k8s.io/v1/MutatingWebhookConfiguration", name: string, property: "api_version"): pulumi.Output<"admissionregistration.k8s.io/v1">;
-        public getResourceProperty(groupVersionKind: "admissionregistration.k8s.io/v1/MutatingWebhookConfiguration", namespace: string, name: string, property: "api_version"): pulumi.Output<"admissionregistration.k8s.io/v1">;
+        public getResourceProperty(groupVersionKind: "admissionregistration.k8s.io/v1/MutatingWebhookConfiguration", name: string, property: "apiVersion"): pulumi.Output<"admissionregistration.k8s.io/v1">;
```
lblackstone: Was this supposed to change? I think the change looks correct, so I'm guessing this was a bug you uncovered while updating the generator code?

tests/integration/dotnet/basic/Program.cs
```diff
@@ -0,0 +1,317 @@
+﻿// Copyright 2016-2019, Pulumi Corporation.  All rights reserved.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Pulumi;
+using Pulumi.Kubernetes.Core.V1;
+using Pulumi.Kubernetes.Types.Inputs.Core.V1;
```
lblackstone: Is there any way to avoid having to import all of these individually?

sdk/dotnet/Apps/V1/Deployment.cs
lukehoban: Yes.
Screen Shot 2019-11-22 at 3 41 47 PM

sdk/dotnet/Apps/V1/Deployment.cs
lukehoban: Yeah - we didn't yet match that in the TF .NET SDK - so I've aligned with other .NET projections here for now.  We can follow up on this if/when needed and will need to do so for .NET as well.  Better not to implement now until we know how we'll want to use this across .NET.

sdk/nodejs/yaml/yaml.ts
lukehoban: Yes - I believe this was a bug previously that was fixed incidentally as part of other fixes in `typegen.go`.

tests/integration/dotnet/basic/Program.cs
lukehoban: No - at least not without typing them out explicitly when used.

sdk/dotnet/AdmissionRegistration/V1/MutatingWebhookConfiguration.cs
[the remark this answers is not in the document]
lukehoban: I actually was slightly wrong about this.  In TypeScript, we do expose these properties on the input interfaces, but we then also always replace whatever values are set with fixed values.  So it's mostly the same as what we currently do in C#.  It's probably slightly "wrong" in both, but I think I'll leave as-is for now.  

The one thing that's slightly different is that TS copies the fields into a new object instead of modifying the args object passed in by the user.  I'm okay doing that for now here - though it's something we could potentially improve on in the future without a breaking change here.

pkg/gen/typegen.go
CyrusNajmabadi: Can we json stringify this so it is a string in the .net side?  Alternatively, we could define this in .net as an actual Json type and the marshal properly to that.

lukehoban: Opened #889 to track this - and going to go ahead and get this PR merged without this first (since it already works for most resources).  Will add your note over on that issue.

PR #897, opened 2019-11-27
Kubernetes provider

mikhailshilkov:
Add a provider resource to the .NET SDK


lukehoban:
Test failure appears to be due to some flakiness with CRDs and refresh that I've seen a few times (but is not consistent).  I pushed https://github.com/pulumi/pulumi-kubernetes/commit/2fe9d0f01a46a552f7789cae5739c8d81641f3ee to try to address this.

PR #899, opened 2019-11-29
Prepare for v1.3.3 release

stack72:


stack72:
I *really* want to get the Provider.cs class for DotNet SDK released so that I can finish off the porting of K8s examples to CSharp

PR #1001, opened 2020-02-20
Custom resource options for provider

mikhailshilkov:
`ResourceOptions` becomes an abstract class in pulumi/pulumi#3857.
Pass `CustomResourceOptions` to the provider.


mikhailshilkov:
This may cause a compilation error once they upgrade the dependency versions. Once it's fixed, the deployment won't produce any differences.

Issue #1009, opened 2020-02-28
Kubernetes Python package does not fail gracefully when `pulumi` isn't on the path


joeduffy:
For our other packages, like AWS, etc, we fail gracefully when `pulumi` isn't on the path:

```
class InstallPluginCommand(install):
    def run(self):
        install.run(self)
        try:
            check_call(['pulumi', 'plugin', 'install', 'resource', 'aws', '${PLUGIN_VERSION}'])
        except OSError as error:
            if error.errno == errno.ENOENT:
                print("""
                There was an error installing the aws resource provider plugin.
                It looks like `pulumi` is not installed on your system.
                Please visit https://pulumi.com/ to install the Pulumi CLI.
                You may try manually installing the plugin by running
                `pulumi plugin install resource aws ${PLUGIN_VERSION}`
                """)
            else:
                raise
```

For the Kubernetes package, we do not:

```
class InstallPluginCommand(install):
    def run(self):
        install.run(self)
        check_call(['pulumi', 'plugin', 'install', 'resource', 'kubernetes', '${PLUGIN_VERSION}'])
```

We should fix the Kubernetes package, as it's very reasonable for end users to want to install packages manually ahead of time, when `pulumi` isn't available and/or Internet access is unavailable.

PR #1010, opened 2020-02-28
Display appropriate error when pulumi plugin install fails

stack72:
Fixes: #1009

This ensures we don't error but instead display an appropriate
error message to ensure that Pulumi is available on the users system


stack72:
@lukehoban that is awesome to know because I spent some time trying to work out where the code was autogenerated from in this package :D

PR #1073, opened 2020-04-16
Upgrade to v2.0.0 of Pulumi

stack72:


sdk/dotnet/Pulumi.Kubernetes.csproj
```diff
@@ -25,7 +25,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Pulumi" Version="1.12.1-preview" />
+    <PackageReference Include="Pulumi" Version="2.0.0" />
```
mikhailshilkov, line 28: Is that intentionally a specific version and not `2.*`?

stack72: It is not - changing now!

PR #1141, opened 2020-05-21
Publish plugins to s3://get.pulumi.com, too.

praneetloke:
Part of pulumi/home#845.

This PR updates the publish-plugin.sh script to copy built plugins to s3://get.pulumi.com
in addition to s3://rel.pulumi.com. This is so that we can migrate to serving plugins via
CDN on https://get.pulumi.com, instead of https://api.pulumi.com like we do today.

For more information about the overall effort, see [this comment](https://github.com/pulumi/pulumi-service/issues/3068#issuecomment-626878731).

In addition, immediately before this code gets merged, I will also update the AWS access key
used for this repo on:

- Travis CI: https://travis-ci.com/github/pulumi/pulumi-kubernetes/settings
- GitHub Actions: https://github.com/pulumi/pulumi-kubernetes/settings/secrets (if applicable)

The access key will be for a different IAM User. This is part of https://github.com/pulumi/home/issues/708,
and managing Pulumi's internal AWS access using Pulumi (and migrating off hand-crafted IAM permissions we set up in 2017).

> I'm assuming that this repository doesn't need to create AWS resources, so
> I'll use an access key that only has permissions to upload resources and
> publish plugins but nothing else. I'll discover if this is correct when
> testing out the release path post-merge. (And can update Travis out of band.)

## Next Steps
After this review gets the LGTM, the next steps are:
1. Update the AWS credentials on Travis, GitHub (if applicable)
2. Merge the PR. Wait for master to be green. Fix any problems associated with this PR.

### DIFF
--- a/scripts/publish-plugin.sh
+++ b/scripts/publish-plugin.sh
@@ -3,10 +3,13 @@
 # s3://rel.pulumi.com/releases/plugins.
 set -o nounset -o errexit -o pipefail
 
+# Update this with the name of the provider
+PROVIDER_NAME="kubernetes"
+
 ROOT=$(dirname $0)/..
 WORK_PATH=$(mktemp -d)
 VERSION=$(jq -r '.version' < "${ROOT}/sdk/nodejs/bin/package.json")
-PLUGIN_PACKAGE_NAME="pulumi-resource-kubernetes-${VERSION}-$(go env GOOS)-$(go env GOARCH).tar.gz"
+PLUGIN_PACKAGE_NAME="pulumi-resource-${PROVIDER_NAME}-${VERSION}-$(go env GOOS)-$(go env GOARCH).tar.gz"
 PLUGIN_PACKAGE_DIR="$(mktemp -d)"
 PLUGIN_PACKAGE_PATH="${PLUGIN_PACKAGE_DIR}/${PLUGIN_PACKAGE_NAME}"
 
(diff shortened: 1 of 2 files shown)


scripts/publish-plugin.sh
```diff
@@ -16,26 +19,70 @@ if [ "$(go env GOOS)" = "windows" ]; then
     BIN_SUFFIX=".exe"
 fi
 
-(cd "${ROOT}/provider" && go build \
-   -ldflags "-X github.com/pulumi/pulumi-kubernetes/provider/v2/pkg/version.Version=${VERSION}" \
-   -o "${WORK_PATH}/pulumi-resource-kubernetes${BIN_SUFFIX}" \
-   "${ROOT}/cmd/pulumi-resource-kubernetes")
+(cd provider && go build \
```
praneetloke, line 22: I am not sure if the change in the line is ok. @metral

metral: For ref we do this in other PRs: 

- https://github.com/pulumi/pulumi-auth0/pull/2/files#diff-682a1507e4f9c0dae17cadd4bec21056L22
- https://github.com/pulumi/pulumi-aws/pull/973/files#diff-682a1507e4f9c0dae17cadd4bec21056R22

praneetloke: Yes, I used `pulumi-digitalocean` for reference and it, too, was like `pulumi-auth0` from your link. I guess I should have clarified that I just wasn't sure if the way that this script is invoked in _this_ repo necessitates the `${ROOT}` prefix in the command. But it sounds like it is not so the change should be ok?

metral: Ah gotcha. IIUC `${ROOT}` does not seem to be needed @chrsmith ?

praneetloke:
Updated AWS creds in Travis. There are no GHA secrets for this repo.

PR #1249, opened 2020-08-12
[crd2pulumi] Handle multiple CRDs at once

albert-zhong:
Here's my fix to handle multiple CRDs in one YAML file. This should resolve issues #1246 and #1245 (at least for the `cert-manager.crds.yaml` file @lblackstone had, which I verified). Soon, I want the CLI to handle reading multiple YAML files and remote YAML files as well.

I also did some refactoring and I managed to just use the Go codegen's `GeneratePackage()` function directly.

Notice: the codegen groups Custom Resources with the same `group` (e.g acme.cert-manager-io, cert-manager-io) under the same folder. Some of these `group` names have dots in them, so I just swap them out for hyphens (otherwise the NodeJS imports would break).


provider/cmd/crd2pulumi/gen/generate.go
```diff
@@ -101,25 +204,29 @@ func NewCustomResourceGenerator(language, yamlPath, outputDir string) (CustomRes
 		}
 	}
 
-	kind, _, _ := unstruct.NestedString(crd.Object, "spec", "names", "kind")
-	plural, _, _ := unstruct.NestedString(crd.Object, "spec", "names", "plural")
-	group, _, _ := unstruct.NestedString(crd.Object, "spec", "group")
+	kind, foundKind, _ := unstruct.NestedString(crd.Object, "spec", "names", "kind")
+	if !foundKind {
+		return CustomResourceGenerator{}, errors.Errorf("could not find `spec.names.kind` field in the CRD")
```
justinvp, line 209: Nit:

```suggestion
		return CustomResourceGenerator{}, errors.New("could not find `spec.names.kind` field in the CRD")
```

provider/cmd/crd2pulumi/gen/generate.go
```diff
@@ -101,25 +204,29 @@ func NewCustomResourceGenerator(language, yamlPath, outputDir string) (CustomRes
...
+	}
+	plural, foundPlural, _ := unstruct.NestedString(crd.Object, "spec", "names", "plural")
+	if !foundPlural {
+		return CustomResourceGenerator{}, errors.Errorf("could not find `spec.names.plural` field in the CRD")
```
justinvp, line 213:
```suggestion
		return CustomResourceGenerator{}, errors.New("could not find `spec.names.plural` field in the CRD")
```

provider/cmd/crd2pulumi/gen/generate.go
```diff
@@ -101,25 +204,29 @@ func NewCustomResourceGenerator(language, yamlPath, outputDir string) (CustomRes
...
+	}
+	group, foundGroup, _ := unstruct.NestedString(crd.Object, "spec", "group")
+	if !foundGroup {
+		return CustomResourceGenerator{}, errors.Errorf("could not find `spec.group` field in the CRD")
```
justinvp, line 217:
```suggestion
		return CustomResourceGenerator{}, errors.New("could not find `spec.group` field in the CRD")
```

provider/cmd/crd2pulumi/gen/schema.go
```diff
@@ -187,27 +177,6 @@ func addType(schema map[string]interface{}, name string, types map[string]pschem
 	}
 }
 
-// UnderscoreFields replaces the hyphens in the x-kubernetes-... fields with
-// underscores
-func UnderscoreFields(schema map[string]interface{}) {
```
justinvp, line 192: This is no longer needed?

albert-zhong: For this commit I temporarily stopped generating `new CustomResourceDefinition()...` helper classes in TypeScript, so this function wasn't needed. I'll add back in again, but once `kube2pulumi` fixes CRDs, this will be removed permanently.

Issue #1261, opened 2020-08-19
Auto-generate libraries for all operators/CRDs in community-managed operator-framework/community-operators


lukehoban:
We recently added the `crd2pulumi` tools, which is a nice way to generate a strongly typed wrapper around a CRD.  However, users need to find the tool, download it, run it, check in the generated code and own it, re-run it on every future update to the CRD version, etc.  Moreover, while it helps provide wrappers over the CRD and the CRs, it does not help make it easier to install the Operators themselves that power these - so users still have to find bespoke instructions for that for each operator (Helm chart, etc.).

An option for addressing these remaining issues in a scalable way is to auto-generate the CRs, CRDs, and possibly even the deployments of the Operators based on the metadata that is captured in https://github.com/operator-framework/community-operators (or possibly another open registry if there are others out there that are now, or in the future become, more complete).

https://github.com/operator-framework/community-operators powers https://operatorhub.io/ today, and has 148 operators managed there, with versioning for all operators, active community contribution, and schematization of both the CRD metadata needed for `crd2pulumi` - but also the information needed to install the corresponding Operator that powers the CRD if we want to use it (e.g. https://github.com/operator-framework/community-operators/blob/master/upstream-community-operators/cockroachdb/3.0.7/cockroachdb.v3.0.7.clusterserviceversion.yaml#L225-L347).  

In theory, we could auto-generate a Pulumi library from this that provided easy access to all of these (without the need for unreliable Helm charts, etc.) including all versions and strongly typed APIs in all Pulumi languages.

justinvp:
We've generated these at https://github.com/pulumi/pulumi-kubernetes-crds

PR #1282, opened 2020-08-28
[crd2pulumi] Initial commit for .NET and Python

albert-zhong:
Here's my initial attempt at generating .NET crds. It works (at least the simple CRDs I tried), although unfortunately I had to do some ugly search-and-replace to fix 2 issues:

- **Issue 1**: By default the codegen references `Pulumi.Kubernetes.Meta.V1.Inputs.ObjectMeta` and `Pulumi.Kubernetes.Meta.V1.Outputs.ObjectMeta` for the input/output types respectively for the ObjectMeta type. However this doesn't match up with the actual ObjectMeta types in this repo, which have the namespaces `Pulumi.Kubernetes.Types.Inputs.Meta.V1` and `Pulumi.Kubernetes.Types.Outputs.Meta.V1` respectively. So it ends up not being imported at all.
- Currently I just manually insert the import statements, and search/replace all instances of the old ObjectMeta references to the new one in all CustomResource classes. However I feel this is quite inefficient, since it means I have to copy over the buffer for the code 3 times, for every single Custom Resource class. I feel like a better way would be to generate a namespace at the `Pulumi.Kubernetes.Meta.V1.Inputs.ObjectMeta` location where the codegen thinks ObjectMeta should be at, and just "export" a reference to the actual ObjectMeta class. @justinvp and @lblackstone , is there a way to do this in C#?

- **Issue 2**: The generated `Utilities.cs` class originally causes this error:
```
error: Running program '/Users/albert/go/src/github.com/pulumi/pulumi-kubernetes/provider/cmd/crd2pulumi/examples/dotnet/bin/Debug/netcoreapp3.1/dotnet-crds.dll' failed with an unhandled exception:
    System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
     ---> System.TypeInitializationException: The type initializer for 'Pulumi.Kubernetes.Utilities' threw an exception.
     ---> System.NotSupportedException: Missing embedded version.txt file
```
As a easy fix I just generated my own copy of `Utilities.cs`, but with the entire function that looks for `version.txt` removed. However I'm not sure if this is actually needed sometimes, so is there another fix to this?

EDIT: I also added Python support. It was pretty simple to do--the only thing I had to modify was the `_utilities.py.` Running `pulumi up` didn't work due to some dependency bug, so I created a `_utilities.py` file that just imported the same file from `pulumi_kubernetes`, and just called all of its functions.


provider/cmd/crd2pulumi/gen/python.go
```diff
@@ -0,0 +1,80 @@
...
+	"bytes"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"github.com/pulumi/pulumi/pkg/v2/codegen/python"
+	pschema "github.com/pulumi/pulumi/pkg/v2/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
+)
+
+func (pg *PackageGenerator) genPython(types map[string]pschema.ObjectTypeSpec, baseRefs []string) (map[string]*bytes.Buffer, error) {
```
justinvp, line 27: Note: In order to get the latest codegen, you'll need to update `go.mod` to refer to a more recent tag/commit. v2.9.2 was recently tagged, so that would work.

In go.mod:

```
	github.com/pulumi/pulumi/pkg/v2 v2.9.2
	github.com/pulumi/pulumi/sdk/v2 v2.9.2
```

provider/cmd/crd2pulumi/gen/nodejs.go
```diff
@@ -34,34 +32,14 @@ const metaFile = `import * as k8s from "@pulumi/kubernetes";
...
-
-// un-capitalizes the first character of a string
-func toLowerFirst(input string) string {
-	if input == "" {
-		return ""
-	}
-	return string(unicode.ToLower(rune(input[0]))) + input[1:]
-}
-
 func (pg *PackageGenerator) genNodeJS(types map[string]pschema.ObjectTypeSpec, baseRefs []string) (map[string]*bytes.Buffer, error) {
```
justinvp, line 35: FYI, I am running into the following assert when running this locally:

```
$ go run . nodejs resourcedefinition.yaml out/nodejs
panic: fatal: An assertion has failed

goroutine 1 [running]:
github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.failfast(...)
	/Users/justin/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.9.2/go/common/util/contract/failfast.go:23
github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.Assert(...)
	/Users/justin/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.9.2/go/common/util/contract/assert.go:26
github.com/pulumi/pulumi-kubernetes/provider/v2/cmd/crd2pulumi/gen.(*PackageGenerator).genNodeJS(0xc000121c90, 0xc000146960, 0xc000612930, 0x1, 0x1, 0x1, 0x1, 0xc0006128e0)
	/Users/justin/go/src/github.com/pulumi/pulumi-kubernetes/provider/cmd/crd2pulumi/gen/nodejs.go:69 +0xbb0
github.com/pulumi/pulumi-kubernetes/provider/v2/cmd/crd2pulumi/gen.(*PackageGenerator).generateFiles(0xc000121c90, 0x6, 0x7ffeefbff9ae, 0x17)
	/Users/justin/go/src/github.com/pulumi/pulumi-kubernetes/provider/cmd/crd2pulumi/gen/generate.go:168 +0x25c
github.com/pulumi/pulumi-kubernetes/provider/v2/cmd/crd2pulumi/gen.Generate(0x7ffeefbff9a7, 0x6, 0x7ffeefbff9ae, 0x17, 0x7ffeefbff9c6, 0xa, 0xc000121d00, 0x1c03ce0, 0xc0007050a0)
	/Users/justin/go/src/github.com/pulumi/pulumi-kubernetes/provider/cmd/crd2pulumi/gen/generate.go:59 +0x159
github.com/pulumi/pulumi-kubernetes/provider/v2/cmd/crd2pulumi/cmd.glob..func2(0x26b7ae0, 0xc000133a70, 0x3, 0x3)
	/Users/justin/go/src/github.com/pulumi/pulumi-kubernetes/provider/cmd/crd2pulumi/cmd/root.go:54 +0x105
github.com/spf13/cobra.(*Command).execute(0x26b7ae0, 0xc0001ac010, 0x3, 0x3, 0x26b7ae0, 0xc0001ac010)
	/Users/justin/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:846 +0x29d
github.com/spf13/cobra.(*Command).ExecuteC(0x26b7ae0, 0x0, 0x0, 0xc000180058)
	/Users/justin/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x349
github.com/spf13/cobra.(*Command).Execute(...)
	/Users/justin/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
github.com/pulumi/pulumi-kubernetes/provider/v2/cmd/crd2pulumi/cmd.Execute(...)
	/Users/justin/go/src/github.com/pulumi/pulumi-kubernetes/provider/cmd/crd2pulumi/cmd/root.go:68
main.main()
	/Users/justin/go/src/github.com/pulumi/pulumi-kubernetes/provider/cmd/crd2pulumi/main.go:22 +0x2e
exit status 2

```

Though, I think you've force pushed some changes since I've run this, so not sure if it's still an issue.

provider/cmd/crd2pulumi/gen/dotnet.go
```diff
@@ -0,0 +1,203 @@
...
+        {
+        }
+    }
+}
+`
+
+// For some reason, we get a `Missing embedded version.txt file` error if we
+// tried running `pulumi up` with the normal `Utilities.cs` file.
+// As a temporary fix, this modified `Utilities.cs` file just removes the
+// `static Utilities()` method.
```
justinvp, line 136: I think we're going to need to come up with a better solution, because creating the resource sets the `CustomResourceOptions.Version` to the value returned from `Utilities.Version`.

```c#
        public CronTab(string name, Pulumi.Kubernetes.Types.Inputs.Stable.V1.CronTabArgs? args = null, CustomResourceOptions? options = null)
            : base("kubernetes:stable.example.com/v1:CronTab", name, MakeArgs(args), MakeResourceOptions(options, ""))
        {
        }

        private static CustomResourceOptions MakeResourceOptions(CustomResourceOptions? options, Input<string>? id)
        {
            var defaultOptions = new CustomResourceOptions
            {
                Version = Utilities.Version,
            };
            var merged = CustomResourceOptions.Merge(defaultOptions, options);
            // Override the ID if one was specified for consistency with other language SDKs.
            merged.Id = id ?? merged.Id;
            return merged;
        }
```

We might need to update the implementation of Utilities that we emit here to have `Utilities.Version` call the `Utilities.Version` inside the Kubernetes assembly using private reflection. Alternatively, we may need to add a new flag to pass to the codegen in `pulumi/pulumi` for .NET to not set the `CustomResourceOptions.Version` at all in any emitted resources, if it's OK not to set it (not sure why setting it is necessary). Curious if @mikhailshilkov has an opinion here.

I think we may also be able to pare down this copy of the class -- I don't think `GetEnv*` and `WithVersion` are needed here.

provider/cmd/crd2pulumi/gen/dotnet.go
```diff
@@ -0,0 +1,203 @@
...
+	// class defined in the .NET SDK is located at
+	// `Pulumi.Kubernetes.Types.Outputs.Meta.V1.ObjectMeta`. This path would
+	// only get generated properly if `compatibility` was `kubernetes20`.
+	pkg.Language["csharp"] = rawMessage(map[string]interface{}{
+		"compatibility":          "kubernetes20",
+		"dictionaryConstructors": true,
+		"namespaces":             namespaces,
+	})
+
+	files, err := dotnet.GeneratePackage(tool, pkg, nil)
```
justinvp, line 70: We probably also want to customize the name of the resulting `Pulumi.Kubernetes.csproj` file that is emitted. Ultimately, for https://github.com/pulumi/pulumi-kubernetes/issues/1261, we're going to want to have a package name based on each community operator name. We need to come up with a naming scheme for these. For .NET, maybe these will be something like `Pulumi.Kubernetes.Crds.<operator>` (and similar for other languages like `@pulumi/kubernetes-crds-<operator>` for Node.js, `pulumi_kubernetes_crds_<operator>` for Python, etc.).

I see for Python you're using `pulumi_crds`. Perhaps for now this file should be `Pulumi.Crds.csproj`?

provider/cmd/crd2pulumi/gen/python.go
```diff
@@ -0,0 +1,80 @@
...
+def get_env_bool(*args):
+    return _utilities.get_env_bool(*args)
+
+
+def get_env_int(*args):
+    return _utilities.get_env_int(*args)
+
+
+def get_env_float(*args):
+    return _utilities.get_env_float(*args)
```
justinvp, line 75: I don't think any of the `get_env_*` are needed here.

provider/cmd/crd2pulumi/gen/python.go
```diff
@@ -0,0 +1,80 @@
...
+		"compatibility":       "kubernetes20",
+		"moduleNameOverrides": pg.moduleToPackage(),
+		"requires": map[string]string{
+			"pulumi":   "\u003e=2.0.0,\u003c3.0.0",
+			"pyyaml":   "\u003e=5.1,\u003c5.2",
+			"requests": "\u003e=2.21.0,\u003c2.22.0",
+		},
+	})
+
+	files, err := python.GeneratePackage(tool, pkg, nil)
```
justinvp, line 42: We can skip emitting `provider.py` and the empty `README.md`.

I think we're going to want to be able to customize the generated `setup.py`. First, it has an `InstallPluginCommand` that isn't relevant here that we need to strip.

Second, it has `${VERSION}` which is intended to be replaced as part of the build process. For https://github.com/pulumi/pulumi-kubernetes/issues/1261, we'll prob need it for each package we generate, but for people just running the tool on their own, it's probably not needed. For now, it's probably fine to leave it as-is, we'll just need to decide what we want to do here. We can chat about it in our 1:1.

provider/cmd/crd2pulumi/gen/dotnet.go
albert-zhong: I set the package name to be Kubernetes across all languages, so it should be `pulumi_kubernetes` for Python now.

provider/cmd/crd2pulumi/gen/python.go
albert-zhong: I got a missing resource bug when I used the generated `_utilities.py` file, so to be safe I just made this boilerplate script that just calls the imported SDK's own `_utilities.py`functions.

provider/cmd/crd2pulumi/gen/dotnet.go
mikhailshilkov: > Kubernetes assembly using private reflection

Assembly version isn't always equal to the provider version, unfortunately. I know they are different for alpha-generated versions, when a provider version isn't a legal assembly version.

Anyway, it looks like we need another flag for .NET codegen of CRDs:
- Don't generate `Utilities`, `MakeResourceOptions`, etc.
- Don't use `KubernetesResource` base class
- Maybe, use the standard namespacing approach but be aware of k8s-specific namespacing for existing classes like metadata
- Rename `csproj`
- ...

Ultimately, custom resources are different from our own resources, and it looks like we can't ignore that.

Was this not necessary for other languages?

mikhailshilkov (approved):
Agreed with @justinvp that we can take this and iterate from here. The workarounds are ugly, but I guess they are close to the best we can do without changing the main codegen. We should probably invest into a CRDT-aware `pulumi/pulumi` to clean up as per another comment thread above.

provider/cmd/crd2pulumi/gen/dotnet.go
albert-zhong: Although TypeScript used the generated `Utilities` file, it never ran into an error somehow. For Go, it seemed like the Utilities file was never even referenced by the CustomResources in the first place. For Python I got an error with Utilities, but since the Utilities file just contained functions, I was able to import the same utilities file from the actual SDK and have the generated utilities file just call those same functions instead (not sure if this is an actual solution though).

provider/cmd/crd2pulumi/gen/dotnet.go
```diff
@@ -0,0 +1,203 @@
...
+        }
+    }
+}
+`
+
+// For some reason, we get a `Missing embedded version.txt file` error if we
+// tried running `pulumi up` with the normal `Utilities.cs` file.
+// As a temporary fix, this modified `Utilities.cs` file just removes the
+// `static Utilities()` method.
+const dotNetUtilities = `// *** WARNING: this file was generated by crd2pulumi. ***
```
justinvp, line 137: For the time being, here's the `Utilities.cs` we could generate for the crds package, which uses reflection to get the version from the `Pulumi.Kubernetes` assembly:

```c#
// *** WARNING: this file was generated by crd2pulumi. ***
// *** Do not edit by hand unless you're certain you know what you are doing! ***

using System;
using System.Reflection;

namespace Pulumi.Kubernetes
{
    static class Utilities
    {
        public static string Version { get; } = GetVersion();

        private static string GetVersion()
        {
            const string UtilitiesType = "Pulumi.Kubernetes.Utilities";
            const string VersionProperty = "Version";

            Type? type = Assembly.GetAssembly(typeof(Provider))?.GetType(UtilitiesType);
            if (type is null)
            {
                throw new InvalidOperationException($"{UtilitiesType} type could not be obtained.");
            }

            PropertyInfo? prop = type.GetProperty(VersionProperty, BindingFlags.Static | BindingFlags.Public);
            if (prop is null)
            {
                throw new InvalidOperationException($"{UtilitiesType}.{VersionProperty} property could not be obtained.");
            }

            var result = prop.GetValue(type, null) as string;
            if (result is null)
            {
                throw new InvalidOperationException($"Expected {UtilitiesType}.{VersionProperty} to return a non-null string.");
            }
            return result;
        }
    }
}
```

PR #1292, opened 2020-09-05
[sdk/python] Add ObjectMetaArgs type to CustomResource

albert-zhong:
Updated `metadata: Optional[pulumi.Input[Any]]` to `metadata: Optional[pulumi.Input[pulumi.InputType['_meta.v1.ObjectMetaArgs']]]`


justinvp (commented):
This needs to be changed in `provider/pkg/gen/python-templates/apiextensions/CustomResource.py`, and then `make build` will copy the changes to `sdk/python/pulumi_kubernetes/apiextensions/CustomResource.py`.

Aside from that, the changes look OK to me, but adding @lblackstone to double-check.

Issue #1302, opened 2020-09-09
crd2pulumi panics on CRD without OpenApi schema


hazsetata:
### Problem description

When trying to run crd2pulumi on a CRD that doesn't contain an OpenApi schema (or on a YAML containing a set of CRDs, and at least one of them doesn't have an OpenApi schema), the tool panics. I compiled it from source, as the released binary had a different problem on the file (`error: invalid apiVersion`).

An example from Ambassador API Gateway I ran into:

```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  annotations: {}
  labels:
    app.kubernetes.io/name: ambassador
    product: aes
  name: filterpolicies.getambassador.io
spec:
  group: getambassador.io
  names:
    categories:
      - ambassador-crds
    kind: FilterPolicy
    plural: filterpolicies
    shortNames:
      - fp
    singular: filterpolicy
  scope: Namespaced
  version: null
  versions:
    - name: v2
      served: true
      storage: true
    - name: v1beta2
      served: true
      storage: false
    - name: v1beta1
      served: true
      storage: false

```

### Errors & Logs

```
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/pulumi/pulumi-kubernetes/provider/v2/cmd/crd2pulumi/gen.(*PackageGenerator).GetTypes(0xc0005239e8, 0xc00063b3b0)
        /Development/go/src/github.com/pulumi/pulumi-kubernetes/provider/cmd/crd2pulumi/gen/schema.go:62 +0x32f
github.com/pulumi/pulumi-kubernetes/provider/v2/cmd/crd2pulumi/gen.NewPackageGenerator(0xc0002db680, 0x1, 0x4, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /Development/go/src/github.com/pulumi/pulumi-kubernetes/provider/cmd/crd2pulumi/gen/generate.go:200 +0x8c7
github.com/pulumi/pulumi-kubernetes/provider/v2/cmd/crd2pulumi/gen.Generate(0xc000528800, 0x0, 0x0, 0x0, 0xc0002db680, 0x1, 0x4, 0xc000523d00, 0x1ae35d6, 0x2506260)
        /Development/go/src/github.com/pulumi/pulumi-kubernetes/provider/cmd/crd2pulumi/gen/generate.go:89 +0x6a
[...]

PR #1305, opened 2020-09-09
Possible fix for #1302

hazsetata:
### Proposed changes

Changes the `AddType` function to return boolean - true if it found the necessary properties in the OpenApi schema, false otherwise. This new return value is used to avoid trying to add properties to non-existing type objects in `GetTypes()` function.

### Related issues

#1302


albert-zhong:
Thanks for fixing this! `crd2pulumi` has moved to a stand-alone repo in [pulumi/crd2pulumi](https://github.com/pulumi/crd2pulumi), so I committed your solution over there.

Issue #1341, opened 2020-10-05
Implement PR workflow that allows OSS usage


stack72:
We want to go with a repository dispatch like we have in examples

PR #1345, opened 2020-10-06
Moving to a command dispatch workflow for PRs

stack72:
Fixes: #1341

This will mean that when a PR is opened, linting and an initial
build of the binaries will be performed to ensure there are no
obvious error.

The PR will have a comment added to it suggesting that the user needs
a maintainer to run the acceptance tests. The maintainer will run
the command `/run-acceptance-tests` and will wait for the results

This will trigger a full end-to-end set of tests including spinning
up the test infrastructure.

It is super important that the PR is checked that no secrets etc are
leaked before running the test command

Please note, as this is a command dispatch, this cannot be tested from
a PR - it must be in master and we can interate there!


stack72:
@lblackstone we need to update the review checks to remove the test-language-sdks. As these will happen out of band, we can't report them as part of this PR... that's the only downside here BUT it will allow anyone to send PRs

lblackstone (approved):

PR #1381, opened 2020-11-16
Moving away from deprecated GHA add-path and set-env

stack72:


stack72:
Self merging as we need this to be in master due to the way that command dispatches work - this has worked in all other repositories

PR #1387, opened 2020-11-19
Replace workload resources on all selector changes

csssuf:
### Proposed changes

Fix forced replacement for selector fields on PVC objects, and also add it to the workload resources.


lblackstone (approved):
LGTM -- Thanks for the PR, @csssuf!

PR #1443, opened 2021-01-22
Minor improvements to the .NET Helm/Yaml code

justinvp:
Addresses minor nitpicks I noticed:

- Specify `StringComparison.Ordinal` for string comparisons (which will be faster and is more correct in these cases, per [string best practices](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings)).
- Avoid unnecessary string, array, and dictionary allocations.
- Fix compiler warning about an issue in an XML doc comment.

### DIFF
--- a/provider/pkg/gen/dotnet-templates/helm/ChartBase.cs
+++ b/provider/pkg/gen/dotnet-templates/helm/ChartBase.cs
@@ -84,7 +84,7 @@ namespace Pulumi.Kubernetes.Helm
                         fetchOptions.Version = cfg.Version;
                         Fetch(chartToFetch, fetchOptions);
                         // Sort the directories into alphabetical order, and choose the first
-                        var fetchedChart = chartDirectory.GetDirectories().OrderBy(x => x.Name).ToArray()[0];
+                        var fetchedChart = chartDirectory.GetDirectories().OrderBy(x => x.Name).First();
                         var fetchedChartName = fetchedChart.Name;
                         chart = fetchedChart.FullName;
                         defaultValues = Path.Join(chartDirectoryName, fetchedChartName, "values.yaml");
@@ -110,13 +110,13 @@ namespace Pulumi.Kubernetes.Helm
                     // > looked up or retrieved in-cluster will be faked locally. Additionally, none
                     // > of the server-side testing of chart validity (e.g. whether an API is supported)
                     // > is done.
-                    var flags = new List<string>(new[]
+                    var flags = new List<string>
                     {
                         "template", chart,
                         "--name-template", releaseName,
                         "--values", defaultValues,
                         "--values", overrides
-                    });
+                    };
                     if (cfgBase.ApiVersions.Length > 0)
                     {
                         foreach (string version in cfgBase.ApiVersions)
@@ -136,7 +136,7 @@ namespace Pulumi.Kubernetes.Helm
                         flags.Add("--include-crds");
                     }
 
-                    var yaml = ExecuteCommand("helm", flags.ToArray(), new Dictionary<string, string>());
+                    var yaml = ExecuteCommand("helm", flags);
                     return ParseTemplate(
                         yaml, cfgBase.Transformations, cfgBase.ResourcePrefix, dependencies, cfgBase.Namespace);
                 }
@@ -161,7 +161,6 @@ namespace Pulumi.Kubernetes.Helm
 
         private static bool IsHelmV3()
         {
-            var env = new Dictionary<string, string>();
             string[] flags = { "version", "--short" };
 
             // Helm v2 returns version like this:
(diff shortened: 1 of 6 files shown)


provider/pkg/gen/dotnet-templates/helm/ChartBase.cs
```diff
@@ -84,7 +84,7 @@ protected ChartBase(string releaseName, Union<ChartArgs, LocalChartArgs> args, C
                         fetchOptions.Version = cfg.Version;
                         Fetch(chartToFetch, fetchOptions);
                         // Sort the directories into alphabetical order, and choose the first
-                        var fetchedChart = chartDirectory.GetDirectories().OrderBy(x => x.Name).ToArray()[0];
+                        var fetchedChart = chartDirectory.GetDirectories().OrderBy(x => x.Name).First();
```
justinvp, line 87: We can just use Linq's `First()` to get the first item, instead of allocating an array to get it.

provider/pkg/gen/dotnet-templates/helm/ChartBase.cs
```diff
@@ -110,13 +110,13 @@ protected ChartBase(string releaseName, Union<ChartArgs, LocalChartArgs> args, C
                     // > looked up or retrieved in-cluster will be faked locally. Additionally, none
                     // > of the server-side testing of chart validity (e.g. whether an API is supported)
                     // > is done.
-                    var flags = new List<string>(new[]
+                    var flags = new List<string>
```
justinvp, line 113: No need to create an intermediate array to pass to the list -- we can pass the initial items directly using the collection initializer syntax.

provider/pkg/gen/dotnet-templates/helm/ChartBase.cs
```diff
@@ -277,7 +277,7 @@ private void Fetch(string chart, ChartFetchArgsUnwrap opts)
                 });
         }
 
-        private static string ExecuteCommand(string command, string[] flags, IDictionary<string, string> env)
+        private static string ExecuteCommand(string command, IEnumerable<string> flags, Dictionary<string, string>? env = null)
```
justinvp, line 280: - Changed `flags` to accept `IEnumerable<string>` so we can pass a `List<string>` directly, without having to allocate and copy to an array.
- Made `evn` optional, so a new dictionary doesn't have to be allocated when not used. Also changed the type from the interface to the concrete type, which avoids the enumerator allocation when enumerating the dictionary.

provider/pkg/gen/dotnet-templates/helm/ChartBase.cs
```diff
@@ -309,11 +312,11 @@ private static string ExecuteCommand(string command, string[] flags, IDictionary
         /// <summary>
         /// Convert an argument array to an argument string for using with Process.StartInfo.Arguments.
         /// </summary>
-        private static string EscapeArguments(params string[] args)
-            => string.Join(" ", args.Select(EscapeArguments));
+        private static string EscapeArguments(IEnumerable<string> args)
+            => string.Join(" ", args.Select(a => EscapeArguments(a)));
```
justinvp, line 316: This is a subtle thing in C#. Specifying `args.Select(EscapeArguments)` will cause a new delegate to be allocated each time, but specifying it like `args.Select(a => EscapeArguments(a))` the compiler will cache the delegate.

provider/pkg/gen/dotnet-templates/helm/ChartBase.cs
```diff
@@ -335,7 +338,7 @@ private static string EscapeArguments(string argument)
 
                     case '\"':
                         // Escape any preceding backslashes
-                        escapedArgument.Append(new string('\\', backslashCount));
+                        escapedArgument.Append('\\', backslashCount);
```
justinvp, line 341: `StringBuilder` supports appending a character a number of times, so no need to allocate a string to do that.

provider/pkg/gen/dotnet-templates/yaml/ConfigGroup.cs
```diff
@@ -200,7 +200,7 @@ public sealed class ConfigGroup : CollectionComponentResource
         /// Create a ConfigGroup resource with the given unique name, arguments, and options.
         /// </summary>
         /// <param name="name">The unique name of the resource</param>
-        /// <param name="args">The arguments used to populate this resource's properties</param>
+        /// <param name="config">The arguments used to populate this resource's properties</param>
```
justinvp, line 203: This fixes two compiler warnings:

```
Yaml/ConfigGroup.cs(203,26): warning CS1572: XML comment has a param tag for 'args', but there is no parameter by that name [/Users/justin/go/src/github.com/pulumi/pulumi-kubernetes/sdk/dotnet/Pulumi.Kubernetes.csproj]
Yaml/ConfigGroup.cs(205,57): warning CS1573: Parameter 'config' has no matching param tag in the XML comment for 'ConfigGroup.ConfigGroup(string, ConfigGroupArgs, ComponentResourceOptions?)' (but other parameters do) [/Users/justin/go/src/github.com/pulumi/pulumi-kubernetes/sdk/dotnet/Pulumi.Kubernetes.csproj]
```

provider/pkg/gen/dotnet-templates/yaml/yaml.tmpl
```diff
@@ -161,7 +161,8 @@ namespace Pulumi.Kubernetes.Yaml
 
         }
 
-        internal static bool IsUrl(string s) => s.StartsWith("http://") || s.StartsWith("https://");
+        internal static bool IsUrl(string s)
+            => s.StartsWith("http://", StringComparison.Ordinal) || s.StartsWith("https://", StringComparison.Ordinal);
```
justinvp, line 165: `StartsWith` does a culture-sensitive string comparison by default. Which means the behavior may vary depending on the system's current culture settings. In this case, we actually want an ordinal comparison (literally comparing the bytes of the string), which will also be faster. https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

lblackstone (approved):
Thanks for helping to clean this up!

PR #1683, opened 2021-08-20
Fix misleading message for disable Helm hook warnings

vizv:
### Proposed changes

Fix the misleading message added in #1682, in schema it uses `PULUMI_K8S_SUPPRESS_HELM_HOOK_WARNINGS`, but message missed the tailing `S`.

### Related issues (optional)

Related: #1682


vizv:
cc @lblackstone

lblackstone (approved):
Ah, good catch. Thanks for the fix @vizv!

PR #1705, opened 2021-09-08
Don't replace PVCs when resources change

### LINKED ISSUE #1665: PVC size update in the spec causes the resource to be replaced instead of updating
I have the PVC provisioned using the default storage class. The storage class has the volume expansion enabled:

```
provisioner: kubernetes.io/gce-pd
parameters:
  type: pd-standard
reclaimPolicy: Delete
allowVolumeExpansion: true
volumeBindingMode: Immediate
```

According to k8s docs, I can [change the PVC spec](https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/) to increase the volume.

However, if I change the size in the PVC spec in my Pulumi stack, the PVC gets replaced, which leads to data loss on the volume as it gets deleted and a new volume is provisioned.

Here is the preview output:
```diff
++  kubernetes:core:PersistentVolumeClaim esdb-data                    create-replacement
```

and the diff:
```
esdb-data (kubernetes:core:PersistentVolumeClaim)
++ kubernetes:core/v1:PersistentVolumeClaim (create-replacement)
    [id=pita-subs/pita-subs-acl-esdb-data]
    [urn=urn:pulumi:production::pita-subs-acl::kubernetes:core/v1:PersistentVolumeClaim::esdb-data]
    __initialApiVersion: "v1"
    __inputs           : {
        apiVersion: "v1"
        kind      : "PersistentVolumeClaim"
        metadata  : {
            annotations: {}
            labels     : {
                app.kubernetes.io/managed-by: "pulumi"
            }
            name       : "pita-subs-acl-esdb-data"
            namespace  : "pita-subs"
        }
        spec      : {
            accessModes     : [
                [0]: "ReadWriteOnce"
            ]
            resources       : {
                requests: {
                    storage: "20Gi"
                }
            }
            storageClassName: "standard"
        }
    }
    metadata           : {
        annotations      : {
            kubectl.kubernetes.io/last-applied-configuration: "{"apiVersion":"v1","kind":"PersistentVolumeClaim","metadata":{"annotations":{},"labels":{"app.kubernetes.io/managed-by":"pulumi"},"name":"pita-subs-acl-esdb-data","namespace":"pita-subs"},"spec":{"accessModes":["ReadWriteOnce"],"resources":{"requests":{"storage":"20Gi"}},"storageClassName":"standard"}}
"
            pv.kubernetes.io/bind-completed                 : "yes"
            pv.kubernetes.io/bound-by-controller            : "yes"
            volume.beta.kubernetes.io/storage-provisioner   : "kubernetes.io/gce-pd"
        }
        creationTimestamp: "2021-08-03T15:28:43Z"
        finalizers       : [
            [0]: "kubernetes.io/pvc-protection"
        ]
        managedFields    : [
            [0]: {
                apiVersion: "v1"
                fieldsType: "FieldsV1"
                fieldsV1  : {
                    f:metadata: {
                        f:annotations: {
                            .                                                 : {}
                            f:kubectl.kubernetes.io/last-applied-configuration: {}
                        }
                        f:labels     : {
                            .                             : {}
                            f:app.kubernetes.io/managed-by: {}
                        }
                    }
                    f:spec    : {
                        f:accessModes     : {}
                        f:resources       : {
                            f:requests: {
                                .        : {}
                                f:storage: {}
                            }
                        }
                        f:storageClassName: {}
                        f:volumeMode      : {}
                    }
                }
                manager   : "pulumi-resource-kubernetes"
                operation : "Update"
                time      : "2021-08-03T15:28:43Z"
            }
            [1]: {
                apiVersion: "v1"
                fieldsType: "FieldsV1"
                fieldsV1  : {
                    f:metadata: {
                        f:annotations: {
                            f:pv.kubernetes.io/bind-completed              : {}
                            f:pv.kubernetes.io/bound-by-controller         : {}
                            f:volume.beta.kubernetes.io/storage-provisioner: {}
                        }
                    }
                    f:spec    : {
                        f:volumeName: {}
                    }
                    f:status  : {
                        f:accessModes: {}
                        f:capacity   : {
                            .        : {}
                            f:storage: {}
                        }
                        f:phase      : {}
                    }
                }
                manager   : "kube-controller-manager"
                operation : "Update"
                time      : "2021-08-03T15:28:45Z"
            }
        ]
        resourceVersion  : "74993236"
        uid              : "5143d38c-3010-4adb-9b67-582e2ba6d0d7"
    }
    spec               : {
        resources       : {
            requests: {
                storage: "20Gi" => "21Gi"
            }
        }
        volumeMode      : "Filesystem"
        volumeName      : "pvc-5143d38c-3010-4adb-9b67-582e2ba6d0d7"
    }
    status             : {
        accessModes: [
            [0]: "ReadWriteOnce"
        ]
        capacity   : {
            storage: "20Gi"
        }
        phase      : "Bound"
    }
```

jaxxstorm:
PVCs often contain critical data that users don't want to lose.

Having the resources field trigger a replace can mean that users can't take advantage of volume resizing.

This allows users to resize volumes and update the field without worrying about a triggering of replacements

### Related issues (optional)

fixes #1665

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## HEAD (Unreleased)
-(None)
+
+- Don't replace PVC on .spec.resources.requests change. (https://github.com/pulumi/pulumi-kubernetes/pull/1705)
 
 ## 3.7.0 (September 3, 2021)
 - Add initial support for a Helm release resource - `kubernetes:helm.sh/v3:Release. Currently available in Beta (https://github.com/pulumi/pulumi-kubernetes/pull/1677)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## HEAD (Unreleased)
 
-- Don't replace PVC on .spec.resources.requests change. (https://github.com/pulumi/pulumi-kubernetes/pull/1705)
-    - *NOTE*: User's will now need to use need to use the `replaceOnChanges` resource option for PVCs if modifying the size
+- Don't replace PVC on .spec.resources.requests or .limits change. (https://github.com/pulumi/pulumi-kubernetes/pull/1705)
+    - *NOTE*: User's will now need to use the `replaceOnChanges` resource option for PVCs if modifying requests or limits to trigger replacement
 
 ## 3.7.0 (September 3, 2021)
 - Add initial support for a Helm release resource - `kubernetes:helm.sh/v3:Release. Currently available in Beta (https://github.com/pulumi/pulumi-kubernetes/pull/1677)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## HEAD (Unreleased)
 
 - Don't replace PVC on .spec.resources.requests change. (https://github.com/pulumi/pulumi-kubernetes/pull/1705)
+    - *NOTE*: User's will now need to use need to use the `replaceOnChanges` resource option for PVCs if modifying the size
 
 ## 3.7.0 (September 3, 2021)
 - Add initial support for a Helm release resource - `kubernetes:helm.sh/v3:Release. Currently available in Beta (https://github.com/pulumi/pulumi-kubernetes/pull/1677)
--- a/provider/pkg/provider/diff.go
+++ b/provider/pkg/provider/diff.go
@@ -109,7 +109,6 @@ var core = _versions{
 				".spec.accessModes",
 				".spec.resources",
 				".spec.resources.limits",
-				".spec.resources.requests",
 				".spec.selector",
 				".spec.storageClassName",
 				".spec.volumeName",
--- a/provider/pkg/provider/diff.go
+++ b/provider/pkg/provider/diff.go
@@ -105,10 +105,7 @@ var core = _versions{
 			}),
 		"PersistentVolumeClaim": append(
 			properties{
-				".spec",
 				".spec.accessModes",
-				".spec.resources",
-				".spec.resources.limits",
 				".spec.selector",
 				".spec.storageClassName",
 				".spec.volumeName",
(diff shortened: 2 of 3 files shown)


roderik:
On AWS, the default storage class does not allow volume expansion. That takes a complex installation of the CSI driver. (which i have so no issue for me personally)

Even so, this would be a breaking change for anyone depending on a replace on request change (which is needed with AWS gp2). The "best" solution would be to check the underlying storageclass to see if it has volume expansion enabled. Otherwise, i think this needs to be an extra options to enable this.

jaxxstorm:
@roderik - unfortunately there isn't a way to examine the storageclass to determine if the PVC is replaceable.

This mechanism allows users to do a replace by opting in (using the replacement annotation)

My logic here is that destroying PVCs is inherently a bad practice as it will nearly always contain user data that is required.

roderik:
> @roderik - unfortunately there isn't a way to examine the storageclass to determine if the PVC is replaceable.
> 
> This mechanism allows users to do a replace by opting in (using the replacement annotation)
> 
> My logic here is that destroying PVCs is inherently a bad practice as it will nearly always contain user data that is required.

I follow that logic and it will error out if someone depends on this dangerous behavior and will not destroy anything or any data. That is a big improvement over the current behavior.

viveklak (approved):

lblackstone (approved):
I'm ok with this change, but let's add a note to the changelog that users can retain the previous behavior by setting the `replaceOnChanges` option, possibly with an example.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## HEAD (Unreleased)
-(None)
+
+- Don't replace PVC on .spec.resources.requests change. (https://github.com/pulumi/pulumi-kubernetes/pull/1705)
+    - *NOTE*: User's will now need to use need to use the `replaceOnChanges` resource option for PVCs if modifying the size
```
viveklak, line 4:
```suggestion
- Don't replace PVC on .spec.resources.requests or .limits change. (https://github.com/pulumi/pulumi-kubernetes/pull/1705)
    - *NOTE*: User's will now need to use the `replaceOnChanges` resource option for PVCs if modifying requests or limits to trigger replacement
```

PR #1747, opened 2021-10-01
Always override namespace for helm release operations

### LINKED ISSUE #1744: Helm release uninstall should remove things in the release namespace
## Hello!

- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details
This is a follow up to #1710 which was fixed in #1733. While install now uses the Helm Release's namespace for resources, uninstall needs to also have the namespace flow through. See comment here: https://github.com/pulumi/pulumi-kubernetes/pull/1733#issuecomment-931977347
 

### Steps to reproduce

1. Use helm release to install charts with templates missing explicit namespace. (Use a namespace not created in the stack.)
2. Ensure resources created in release's namespace
3. pulumi destroy

Expected: All resources in helm release namespace deleted
Actual: Leaked resources in helm release's namespace.

viveklak:
Fixes #1744.

In hindsight, I can't think of a reason why we wouldn't want to always override with the Release's namespace. 

Deletion cleanup is not easily assertable in the tests so not adding one.
I validated this manually with the fluent-bit helm chart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## HEAD (Unreleased)
-(None)
+- Always override namespace for helm release operations (https://github.com/pulumi/pulumi-kubernetes/pull/1747)
 
 ## 3.7.3 (September 30, 2021)
 - Use helm release's namespace on templates where namespace is left unspecified (https://github.com/pulumi/pulumi-kubernetes/pull/1733)
--- a/provider/pkg/provider/helm_release.go
+++ b/provider/pkg/provider/helm_release.go
@@ -202,20 +202,18 @@ func debug(format string, a ...interface{}) {
 	logger.V(9).Infof("[DEBUG] %s", fmt.Sprintf(format, a...))
 }
 
-func (r *helmReleaseProvider) getActionConfig(namespace string, override bool) (*action.Configuration, error) {
+func (r *helmReleaseProvider) getActionConfig(namespace string) (*action.Configuration, error) {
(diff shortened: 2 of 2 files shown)


lblackstone (approved):

ruckc:
@viveklak - How does helm cli operate?  If a resource defines a specific namespace (i.e. kube-system) does helm put it in the release namespace, or the specified namespace?  While this should be an anti-pattern, following how helm cli works should be how Pulumi's helm Release should work

viveklak:
> @viveklak - How does helm cli operate? If a resource defines a specific namespace (i.e. kube-system) does helm put it in the release namespace, or the specified namespace? While this should be an anti-pattern, following how helm cli works should be how Pulumi's helm Release should work

Some background:
Both helm CLI and the current state of helm release would honor what ever the template has by default. If the chart template specifies things go in kube-system then that is where they will go. Alternatively a number of charts have gone with setting the namespace to that of the release - this is something both will honor as well. The ambiguity is around what happens when chart templates don't specify namespaces. Helm CLI's interpretation is to fall back to the default namespace configured for the Kube client. If you run without `--namespace` specified, `default` is the namespace that it falls back to. With the `--namespace` option on the CLI, you are overriding the default namespace to use in the kube client when no namespace is specified.

To answer your question, In my interpretation #1733 and this change are exactly mirroring the CLI behavior with `--namespace` being interpreted as the namespace option for the release resource.

PR #1854, opened 2022-01-04
fix [docs/golang]: Helm omit transformation for golang

matdehaast:
### Proposed changes

Fixes the syntax needed to show how to omit a resource with golang in a helm transformation. This has been changed to be the same as other languages and to the actually required syntax to make an omission work correctly.

### DIFF
--- a/sdk/go/kubernetes/helm/v3/chart.go
+++ b/sdk/go/kubernetes/helm/v3/chart.go
@@ -196,7 +196,7 @@ import (
 //                 func(state map[string]interface{}, opts ...pulumi.ResourceOption) {
 //                     name := state["metadata"].(map[string]interface{})["name"]
 //                     if state["kind"] == "Pod" && name == "test" {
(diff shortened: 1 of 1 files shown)


lblackstone (changes requested):
Thanks for the PR!

You'll also need to update https://github.com/pulumi/pulumi-kubernetes/blob/3efecec54f144132ca9493bcad7f3b3c24842aa6/provider/pkg/gen/_go-templates/helm/v3/chart.go#L199 with this change since the SDK file is auto-generated from that template.

stack72:
Hi @matdehaast 

Thanks so much for the work here - I have rolled this up (with the requested changes @lblackstone suggested) into #1907 

We really appreciate the work here!

Paul

matdehaast:
Thanks @stack72!. @lblackstone apologies I actually seemed to have missed your message in the holiday madness. 

Keep making pulumi an amazing product! 🚀

Issue #1905, opened 2022-02-11
helm/v3 NewChart calls an Invoke without passing in `parent`


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details
https://github.com/pulumi/pulumi/issues/8853#issuecomment-1036198703 contains details. We have added the ability to prevent using default providers. This means that invokes **must** pass a parent or provider explicitly, even if they do not depend on the contents of the provider. The invoke in `helmTemplate` does not do this, causing https://github.com/pulumi/pulumi/issues/8853.

https://github.com/pulumi/pulumi-kubernetes/blob/89cd0c24438806a3ff98ffb9ae3a0944e6600639/sdk/go/kubernetes/helm/v3/chart.go#L302-L314

### Steps to reproduce

1.
2.
3.
4.

Expected: 
Actual:

PR #1906, opened 2022-02-11
Pass parent to Invoke in helmTemplate func

iwahbe:
I have no familiarity with how to use overlays, so let me know if I'm missing something. 

### Proposed changes

We should pass parent when we call the invoke used in creating a new helm `Chart`. 

### Related issues (optional)
Fixes https://github.com/pulumi/pulumi-kubernetes/issues/1905


viveklak (approved):
LGTM

viveklak (approved):

iwahbe:
@viveklak Do you know what is causing https://github.com/pulumi/pulumi-kubernetes/runs/5189217414?check_suite_focus=true?

viveklak:
The use of gotestfmt seems to make it harder to get the compilation failure messages. Getting the generated go sdk to compile with a go build should detect the issue.

PR #1936, opened 2022-03-18
Pin setup-gcloud to v0 instead of master

sethvargo:
### Proposed changes

setup-gcloud will be updating the branch name from master to main in a future release. Even though GitHub will establish redirects, this will break any GitHub Actions workflows that pin to master. This PR updates your GitHub Actions workflows to pin to v0, which is the recommended best practice.

/cc @stack72 👋


sethvargo:
There is no changelog to update

stack72:
Thank you @sethvargo - I'll get this updated in all places

sethvargo:
Thanks! You can close this out if you want - I trust you 😄

stack72 (approved):

PR #1951, opened 2022-04-01
Fix DaemonSet name on diff

alexrocco:
### Proposed changes

The DaemonSet name was set with a lower case s (Daemonset), but the kind
in the state is created with upper case (DaemonSet), and this was
preventing changes on the .spec.selector to trigger a replacement.

This issue was also causing the Kubernetes API to fail as this field is
immutable, but pulumi was trying to update it.

### Related issues (optional)

Related to https://github.com/pulumi/pulumi-kubernetes/pull/1008

Ps. I'm still testing it, but this was the only place where the DaemonSet 
    was with lower case in the code. 
    
Update: Tested locally and this fixes the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Unreleased
+- Fix the DaemonSet name on diff which prevented pulumi to replace the resource (https://github.com/pulumi/pulumi-kubernetes/pull/1951)
 (None)
 
 ## 3.18.1 (April 5, 2022)
--- a/provider/pkg/provider/diff.go
+++ b/provider/pkg/provider/diff.go
@@ -41,14 +41,14 @@ type properties []string
 
 var forceNew = _groups{
 	"apps": _versions{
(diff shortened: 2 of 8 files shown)


alexrocco:
Also, this `diff.go` could be using the types from https://github.com/pulumi/pulumi-kubernetes/blob/master/provider/pkg/kinds/kinds.go.

If the maintainers are fine with that, I could do this change as well.

viveklak:
> 

Thanks for the PR! Do you also have a sample program you can use to demonstrate the fix? I am happy to help you convert that to an integration test.

alexrocco:
> > 
> 
> Thanks for the PR! Do you also have a sample program you can use to demonstrate the fix? I am happy to help you convert that to an integration test.

Just added the integration tests, but don't know if it was the best way of verifying if the DaemonSet was replaced when changing the `.spec.selector`.

viveklak (approved):

viveklak:
Apologies @alexrocco - would you be able to rebase this on top of master? We had some breaks in the linting tool we use which has just been fixed on master. Thanks!

Issue #1851, opened 2021-12-28
NewConfigFile with remote Url


dirien:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details
When I try to use ConfigFile with a remote file (here tekton) nothing happens.

```
tekton, err := yaml.NewConfigFile(ctx, "tekton",
		&yaml.ConfigFileArgs{
			File: "https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml",
			
		}, pulumi.Provider(args.Provider),
	) 
```

Very strange, no error message to work with. 

I checked to URL, and there is no redirect going on. At least what `curl -v -L` tells me

matdehaast:
I'm having the exact same issue, it creates the resource. But nothing happens on the cluster

dirien:
Works!

PR #1962, opened 2022-04-07
Fix fetching remote yaml files

afreakk:
### Proposed changes

Currently it only treats the filepath  as a url if the path fails to parse as url.

### Related issues (optional)

#1851

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## Unreleased
-(None)
+- Fix fetching remote yaml files (https://github.com/pulumi/pulumi-kubernetes/pull/1962)
 
 ## 3.18.2 (April 6, 2022)
 - Only add keyring default value when verification is turned on (https://github.com/pulumi/pulumi-kubernetes/pull/1961)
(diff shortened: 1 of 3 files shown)


viveklak:
Thanks for the PR @afreakk! This file is actually auto generated. Could you make the change to https://github.com/pulumi/pulumi-kubernetes/blob/master/provider/pkg/gen/_go-templates/yaml/yaml.tmpl#L48 instead and then run a `make build`?

afreakk:
@viveklak ok, force pushed updated commit with change only to the tpl file

viveklak:
> @viveklak ok, force pushed updated commit with change only to the tpl file

I expect the CI to fail run to fail since we also expect the generated code checked in. Could you retain the change in the "generated" file and submit that along side the the template file? Alternatively my suggestion to run a make build would perform the generation for you and you could check the results in.

afreakk:
I ran `make build` now, it failed at some python step.
But it still changed the `.go` file.  
Force pushed it now @viveklak

viveklak (approved):

viveklak:
Thanks @afreakk for the fix!

PR #2021, opened 2022-06-16
Use go embed rather than generating a .go file

squaremo:
The newer go versions (>=1.16) include the ability to directly embed files as variables, so it's no longer necessary to generate a .go file with the contents of another file. It _is_ still necessary to have a generation step to rewrite the schema itself. This is how the boilerplate provider build is structured now:

    https://github.com/pulumi/pulumi-provider-boilerplate/blob/main/provider/cmd/pulumi-resource-xyz/main.go#L26

Since I'm porting that change, I've included a little failsafe in the Makefile for when people start using the changed build. The build would fail anyway, but it's less mysterious this way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 Pulumi.*.yaml
 /provider/pkg/gen/openapi-specs
 /provider/cmd/pulumi-resource-kubernetes/schema.go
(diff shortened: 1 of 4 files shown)


stack72 (approved):

stack72:
Thanks so much for this @stack72 - this keeps things in line with what we are doing for the other providers

Issue #2020, opened 2022-06-09
Could not find property 'stringData' listed in additional secret outputs.


yellowhat:
### What happened?

Hi,
with the latest release (`3.34.0` and `3.19.3`), the following code:

```python
casc_secrets = kubernetes.core.v1.Secret(
    "casc-secrets",
    metadata={
        "name": "casc-secrets",
        "namespace": core_ns.metadata.name,
    },
    string_data={
        "developer": "developer",
        "browser": "browser",
    },
    opts=ResourceOptions(provider=cluster),
)
```

Returns:

```bash
$ pulumi up
...
     └─ kubernetes:core/v1:Secret  casc-secrets                      1 warning
...
    warning: Could not find property 'stringData' listed in additional secret outputs.
```

The code works as expected, I am just worried about the warning

### Steps to reproduce

```bash
$ pulumi up
```

### Expected Behavior

No warning

### Actual Behavior

```
    warning: Could not find property 'stringData' listed in additional secret outputs.
```

### Versions used

```
$ pulumi about
CLI          
Version      3.34.0
Go Version   go1.17.11
Go Compiler  gc

Plugins
NAME        VERSION
gcp         6.26.0
kubernetes  3.19.3
python      unknown
random      4.7.0

Host     
OS       fedora
Version  36
Arch     x86_64

This project is written in python: executable='/usr/bin/python3' version='3.10.4'

Current Stack: dev-apps-core

TYPE                                                 URN
pulumi:pulumi:Stack                                  urn:pulumi:dev-apps-core::apps-core::pulumi:pulumi:Stack::apps-core-dev-apps-core
pulumi:providers:random                              urn:pulumi:dev-apps-core::apps-core::pulumi:providers:random::default_4_7_0
pulumi:providers:pulumi                              urn:pulumi:dev-apps-core::apps-core::pulumi:providers:pulumi::default
pulumi:pulumi:StackReference                         urn:pulumi:dev-apps-core::apps-core::pulumi:pulumi:StackReference::dev-infra
pulumi:providers:kubernetes                          urn:pulumi:dev-apps-core::apps-core::pulumi:providers:kubernetes::cluster
random:index/randomPassword:RandomPassword           urn:pulumi:dev-apps-core::apps-core::random:index/randomPassword:RandomPassword::core-passwd
kubernetes:core/v1:Namespace                         urn:pulumi:dev-apps-core::apps-core::kubernetes:core/v1:Namespace::core-ns
[...]

PR #2030, opened 2022-06-17
Remove stringData output

Frassle:
Fixes https://github.com/pulumi/pulumi-kubernetes/issues/2020

### DIFF
--- a/provider/cmd/pulumi-resource-kubernetes/schema.json
+++ b/provider/cmd/pulumi-resource-kubernetes/schema.json
@@ -18496,8 +18496,7 @@
                     "additionalProperties": {
                         "type": "string"
                     },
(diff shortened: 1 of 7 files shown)


viveklak:
Note with the change in https://github.com/pulumi/pulumi-kubernetes/pull/2030/commits/720242c6ffb01c11ad333436d8b5d3ead7efbf8d we still see the following warning on `preview` when *dry-run is not* enabled:

```
Diagnostics:
  kubernetes:core/v1:Secret (casc-secrets):
    warning: Could not find property 'data' listed in additional secret outputs.
```

Frassle:
> warning: Could not find property 'data' listed in additional secret outputs.

That _might_ be an engine bug. Although I'm not sure how we'd work around it short of not warning on preview.

Frassle:
We've reverted this warning engine side.

PR #2073, opened 2022-07-13
Update provider and tests to go 1.18

jkodroff:
This change fixes an issue building the provider due to a missing
reference to an upstream dependency when go 1.16 is used with go mod
tidy. We also change appropriate commands in the Makefile to use go
1.17 compatibility mode.


viveklak (approved):
LGTM. Hopefully CI is happy.

PR #2085, opened 2022-07-15
[internal]Update weekly-pulumi-update.yml

jkisk:
Don't update the sdk/go.mod as part of weekly update until we can do a major release and move off of go 1.16 there.

### DIFF
--- a/.github/workflows/weekly-pulumi-update.yml
+++ b/.github/workflows/weekly-pulumi-update.yml
@@ -82,14 +82,6 @@ jobs:
 
         go mod tidy
 
-        cd ../sdk
-
-        go get github.com/pulumi/pulumi/sdk/v3
-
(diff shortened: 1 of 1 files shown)


.github/workflows/weekly-pulumi-update.yml
```diff
@@ -82,14 +82,6 @@ jobs:
 
         go mod tidy
 
-        cd ../sdk
-
-        go get github.com/pulumi/pulumi/sdk/v3
-
-        go mod download
-
-        go mod tidy
```
viveklak, line 91: I think you would have needed a -compat flag here since the CI job's go version is 1.18. We can revisit this when we re-add support for this separately though.

viveklak (approved):

PR #2234, opened 2022-11-16
Update to k8s v1.26.0(-rc.0)

squaremo:
(This _presently_ does the upgrade to v1.26.0-rc.0, but can be repurposed to v1.26.0 if we wait for that.)

This PR bumps the version of K8s used for codegen, and adjusts some
things so codegen will work:

 - adds a pascalCase spelling of v1beta3 and the new API `resource`
 - qualifies the Resource type with a namespace in dotnet templates,
   since the namespace for the new API will otherwise clash

.. then does the codegen.


squaremo:
Closing in favour of making room first (#2237) then rebasing the original update PR (#2230) on that.

PR #2255, opened 2022-12-06
Make allowNullValues an InputProperty

cdibble:
### Proposed changes

- adds `allowNullValues` to the resource schema `InputProperties`
- rebuilds SDKs to expose the argument so that it can be set with the SDKs

### Related issues (optional)

Completes missing piece from #2220 which added this in the Properties but did not allow one to set the value in the SDKs. The objective is to be able to pass empty arrays through to helm deployments when desired.


cdibble:
@lblackstone LMK if this does the trick :)

lblackstone (approved):

PR #2273, opened 2022-12-17
Docs: dotnet 3.1 and pulumictl is required

jdmaguire:
### Proposed changes

While I was following the README and contributing guide, particularly around the make commands, I got two errors popping out:

- I get told that .NET 3.1, not 3.0, is needed
- The make blurts out that I need pulumictl

I've updated the README and contributing docs with these two bits.


lblackstone (approved):
Thanks for the PR!

PR #2305, opened 2023-02-08
Rework CI workflows and Makefile

squaremo:
Broadly, this is a port of the applicable changes from https://github.com/pulumi/pulumi-azure-native/pull/2180, with the aim of making maintenance easier (by removing duplication and otherwise simplifying), reducing the running time of pipelines, and making local development (closer to) possible.

 - [ ] Rewrite Makefile to use (mostly) file targets, properly
 - [ ] Factor build and test CI workflows out
 - [ ] Factor publish workflow out
 - [ ] Factor common actions out
 - [ ] Remove unnecessary env entries from workflows
 - [ ] Calculate version in workflow, and supply
 - [ ] Make sure artifacts of provider and SDKs are reused between workflows
 - ... other tasks go here as I identify them

Not an aim here:
 - ~rewrite test workflow so that each test job can be rerun individually~ (this is attempted in #2244)


Makefile
```diff
@@ -17,73 +20,108 @@ SWAGGER_URL     ?= https://github.com/kubernetes/kubernetes/raw/${KUBE_VERSION}/
...
 
-openapi_file::
-	@mkdir -p $(OPENAPI_DIR)
-	test -f $(OPENAPI_FILE) || curl -s -L $(SWAGGER_URL) > $(OPENAPI_FILE)
+# The general form for this Makefile is
+#  - faithfully represent dependencies between build targets as files, where possible (i.e., like a normal Makefile)
+#  - use phony targets to give the CI system a kind of API for building in stages
+
+.PHONY: default
+default: build
```
danielrbradley, line 39: looks like build only includes the sdks - so might want to add the provider for completeness

Issue #2331, opened 2023-03-06
External changes to labels and annotations aren't reverted with server side apply.


bernhardloos:
### What happened?

I manually removed a label or annotations from a kubernetes resource managed pulumi. Pulumi notices it in refresh, but then it doesn't repair the missing annotation or label on up and doesn't indicate any differences on subsequent runs.
A changed value gets repaired as expected.
Also, I didn't try this without server side apply.

### Expected Behavior

Pulumi should revert all external changes and restore resources to the configured state.

### Steps to reproduce

1. set enableServerSideApply to true for the provider
2. create a resource with some labels and/or annotations (I used a Secret for my tests)
3. manually delete a label or annotation from the resource with kubectl or some other tool
4. run pulumi refresh, notice that the resource is marked as changed, the removed annotation is in the details
5. run pulumi up
6. the removed annotation/label is not restored

### Output of `pulumi about`

CLI          
Version      3.56.0
Go Version   go1.20.1
Go Compiler  gc

Plugins
NAME             VERSION
kubernetes       3.24.1
kubernetes_crds  0.0.0
python           unknown
tls              4.10.0

Host     
OS       fedora
Version  37
Arch     x86_64

This project is written in python: executable='/usr/bin/python3' version='3.11.1'

Dependencies:
NAME                    VERSION
pip                     22.2.2
pulumi-kubernetes       3.24.1
pulumi-kubernetes-crds  0.0.0
pulumi-tls              4.10.0
setuptools              62.6.0

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Hi @bernhardloos and thanks for reporting this.

At the moment, this is expected behaviour since with SSA enabled, we do not force an override for any conflicts with an existing version of the resource. To enable this, you'll need to append the `"pulumi.com/patchForce": "true"` annotation to the k8s resource you are managing with Pulumi.

### Additional materials:

Please refer to our documentation about upserting a resource for more information:
https://www.pulumi.com/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply/#upsert-a-resource

#2011 also documents the design decisions behind SSA implementation within our provider.

#2280 is an enhancement request to enable patchForce on a stack basis, rather than per-resource basis.

bernhardloos:
[...]

PR #2332, opened 2023-03-08
fix: Remove empty keys when merging unstructured resources

rquitales:
When merging client-side and server-side patches for diffing, we need to remove any empty/removed keys from the server-side patch in the client-side patch as well.

Without this fix, any external changes (eg. kubectl or controllers) to resources that removes fields would not be re-set on the next `pulumi up`.

Partially fixes: #2331 as we still need an upstream fix (https://github.com/imdario/mergo/pull/231) to be merged.


lblackstone (changes requested):
This looks right to me, but will need the mergo dependency to be updated before merging the PR.

Did you confirm that this change fixes the bug in #2331? It would be great if we could cover this case in a test. Perhaps https://github.com/pulumi/pulumi-kubernetes/tree/master/tests/sdk/nodejs/server-side-apply could be extended to do that?

rquitales:
> This looks right to me, but will need the mergo dependency to be updated before merging the PR.
> 
> Did you confirm that this change fixes the bug in #2331? It would be great if we could cover this case in a test. Perhaps https://github.com/pulumi/pulumi-kubernetes/tree/master/tests/sdk/nodejs/server-side-apply could be extended to do that?

Yes, I have confirmed that this change fixes the underlying issue. Have updated mergo now that the upstream PR is merged.

Also added a new regression test for this instead of expanding on the existing `server-side-apply` test case since we need to run external commands in-between pulumi program runs.

Makefile
```diff
@@ -20,7 +20,7 @@ SCHEMA_FILE     := provider/cmd/pulumi-resource-kubernetes/schema.json
 GOPATH			:= $(shell go env GOPATH)
 
 JAVA_GEN 		 := pulumi-java-gen
-JAVA_GEN_VERSION := v0.5.2
+JAVA_GEN_VERSION := v0.8.0
```
rquitales, line 23: Note: PR is currently based on #2337. This change will not be in this PR once #2337 is merged to master and this PR is rebased.

thomas11 (approved):

lblackstone (approved):
Looks great! 🎉

kpitzen (approved):
I assume the Java worktree issue is dependent upon #2337 being merged first, and then this rebased on that.  Pending that being done, LGTM!

Issue #2409, opened 2023-05-16
Creating a Helm release from a remote chart fails with `error: Chart.yaml file is missing`


L3n41c:
### What happened?

We try to create a new Helm release from a remote chart with:

```go
	return helm.NewRelease(e.Ctx, args.InstallName, &helm.ReleaseArgs{
		Namespace: pulumi.StringPtr("datadog"),
		Name:      pulumi.StringPtr("dda"),
		RepositoryOpts: helm.RepositoryOptsArgs{
			Repo: pulumi.StringPtr("https://helm.datadoghq.com"),
		},
		Chart:            pulumi.String("datadog"),
		CreateNamespace:  pulumi.BoolPtr(true),
		DependencyUpdate: pulumi.BoolPtr(true),
		ValueYamlFiles:   valueAssets,
		Values:           args.Values,
	}, opts...)
```

We use `RepositoryOpts.Repo` to indicate that the chart is to be fetched from a remote registry.

We get the following error:
```
Updating (lenaic-eks):
     Type                               Name                                    Status                  Info
[…]
 +   └─ kubernetes:helm.sh/v3:Release   dda                                     **creating failed**     1 error


Diagnostics:
  kubernetes:helm.sh/v3:Release (dda):
    error: Chart.yaml file is missing
```

If we increase the log level, we can see that the `datadog` chart is looked locally instead of on the remote registry:
```
    I0515 17:57:19.454989   36712 log.go:71] Setting registry client with config file: "/Users/vincent.boulineau/Library/Preferences/helm/registry.json" and debug: true
    I0515 17:57:19.455431   36712 log.go:71] Looking up chart path options for release: "dda"
    I0515 17:57:19.455443   36712 log.go:71] Chart name: "datadog"
    I0515 17:57:19.455497   36712 log.go:71] Trying to load chart from path: "/Users/vincent.boulineau/Documents/Dev/test-infra-definitions/datadog"
    I0515 17:57:19.456995   36712 log.go:71] getChart failed: Chart.yaml file is missing
```

This is caused by the fact that the current working directory contains a subdirectory which has exactly the same name as the chart we want to create.
In this case, Helm tries to leverage the local directory instead of honoring the `RepositoryOpts.Repo` parameter.

### Expected Behavior

Have helm fetch the remote chart from the registry.

### Steps to reproduce

[...]

PR #2410, opened 2023-05-16
Improve local helm chart validation

L3n41c:
### Proposed changes

The `helm.NewRelease(…)` function has some logic to look for a local chart before trying to contact a remote registry, even if a `RepositoryOpts.Repo` parameter has been explicitly provided.

This is an issue when the chart we want to create a release from has a name that conflicts with a local folder name.

This change improves the logic and considers a local directory to be a valid chart only if it contains a `Chart.yaml` file.

### Related issues (optional)

* #2409

PR #2428, opened 2023-05-24
Fix chart location logic to work with compressed chart archives

kpitzen:
### Proposed changes

We added more-strict validation logic around chart location in https://github.com/pulumi/pulumi-kubernetes/pull/2410 - this caused a regression for users who use compressed charts via `.tar.gz` etc.

This PR refactors the file checking logic and ensures that compressed chart archives are correctly handled. 2 additional test cases are also added so the following flows are tested:

- Repository URL is specified and a local folder also has the same name as the chart [added in this PR]
- Compressed chart archive is specified [added in this PR]
- Repository URL is specified and no local file/folder is present
- Local folder is specified to be used as chart

### Related issues (optional)

Closes: https://github.com/pulumi/pulumi-kubernetes/issues/2426
Closes: https://github.com/pulumi/pulumi-kubernetes/issues/2409


kpitzen:
@rquitales - I can't approve this PR since I opened the initial draft, but it looks like the changes are sound, and I appreciate the extra tests added! Consider this my ✅

lblackstone (approved):

PR #2432, opened 2023-05-26
Remove server generated fields from state

rquitales:
### Proposed changes

This PR stops the storage of server generated fields: `.status` and certain nested `.metadata` fields, in our state/checkpoints. This will reduce spurious diffing (eg. when running `pulumi refresh`) as our users do not manage these fields.


lblackstone (commented):
I think the direction looks good, but it appears that this change broke some of the .NET tests. https://github.com/pulumi/pulumi-kubernetes/blob/v3.28.1/tests/sdk/dotnet/basic/Program.cs#L138-L144 is failing with an error now, and I think there were a few other similar failures.

PR #2478, opened 2023-06-28
test using gofmttest

rquitales:


rquitales:
There are issues with ProgramTest verbosity printing to stdout and causing jumbled test output. This is likely due to the spinner printed by pulumi CLI and return carriages or something to that effect. Unfortunately there isn't a clean way to inject a custom stdout, so a fix will likely have to be within the upstream ProgramTest pkg.

PR #2541, opened 2023-08-23
Revert "[sdk/python] Generate pyproject.toml (#2531)"

t0yv0:
This reverts commit 0dadcf410b6f7fe06861ced969007a577fcf1df5 due to https://github.com/pulumi/pulumi-kubernetes/issues/2540

Issue #2540, opened 2023-08-23
Recent change disabled python source distribution uploads, is this intended?


edude03:
### What happened?

Not sure if this is a bug or intended behaviour, but this used to work and doesn't anymore and I can't find any documentation highlighting it, so I'm assuming it's a bug. 

Installing `pulumi-kubernetes==4.1.0` (or any version greater than `4.1.0a1691604995`) with `--no-binary` with pip causes `ERROR: No matching distribution found for pulumi-kubernetes==4.1.0` 

### Expected Behavior

I would expect it to install. 

### Steps to reproduce

`pip install --no-binary :all: pulumi-kubernetes==4.1.0`

### Output of `pulumi about`

N/A

### Additional context

As far as I can tell it was broken in [This PR](https://github.com/pulumi/pulumi-kubernetes/pull/2531/files)

Using the git source as a work around by putting this in the requirements.txt:

```
pulumi-kubernetes @ git+https://github.com/pulumi/pulumi-kubernetes@v4.1.0#subdirectory=sdk/python
```

Doesn't work because the [version isn't set in the pyproject.toml](https://github.com/pulumi/pulumi-kubernetes/blob/master/sdk/python/pyproject.toml#L8)

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
Hi @edude03 our intention was to transition to the wheel-based distribution. We'll look into this shortly, we were not aware of the `--no-binary` option or that this would be a breaking change.

CC @rquitales

necaris:
@t0yv0 I'd be curious to understand how you intend to distribute the plugin using a wheel -- from what I can see the wheel doesn't actually include the resource plugin?

t0yv0:
That is correct. The source distribution does not include the plugin either, but auto-installs them upon Python project installation. This redundant (optional) with Pulumi CLI auto-installing the plugins upon pulumi preview/up. The best place to discuss this is in https://github.com/pulumi/pulumi/issues/13709 I think.

rquitales:
Note that v4.1.1 of the Kubernetes provider was released earlier that reverted this change. Closing this issue now but feel free to reopen should more discussion need to occur.

PR #2543, opened 2023-08-28
Reintroduce pyproject.toml and wheel publishing 

t0yv0:
### Proposed changes

This PR introduces the following changes:

- pyproject.toml is generated for the Python SDK, replacing setup.py
- during build, `make python_sdk` now builds a source distribution *and* a wheel 
- when publishing to PyPI, both the wheel and the source distribution are expected to be published
- packaging is wrapped in a dedicated venv to avoid modifying the outer Python environment

### Related issues (optional)

An earlier version of this PR as reverted:

- https://github.com/pulumi/pulumi-kubernetes/pull/2541 

Because it removed source distribution builds and publishing, and caused failures of end users:

https://github.com/pulumi/pulumi-kubernetes/issues/2540


t0yv0:
Testing: since this is a feature-* branch, once CI finishes the expectation is that we can find a dev build on PyPI; this dev build should indicate that both a wheel and a source distribution were pushed to PyPI.

t0yv0:
Closing in favor of #2544

PR #2544, opened 2023-08-28
Reintroduce pyproject.toml and wheel publishing

t0yv0:
### Proposed changes

This PR introduces the following changes:

- pyproject.toml is generated for the Python SDK, replacing setup.py
- during build, `make python_sdk` now builds a source distribution *and* a wheel 
- when publishing to PyPI, both the wheel and the source distribution are expected to be published
- packaging is wrapped in a dedicated venv to avoid modifying the outer Python environment

### Verification

Testing: since this is a feature-* branch, once CI finishes the expectation is that we can find a dev build on PyPI; this dev build should indicate that both a wheel and a source distribution were pushed to PyPI.

### Related issues (optional)

An earlier version of this PR as reverted:

- https://github.com/pulumi/pulumi-kubernetes/pull/2541 

Because it removed source distribution builds and publishing, and caused failures of end users:

https://github.com/pulumi/pulumi-kubernetes/issues/2540


t0yv0:
Good news: this PR produced https://pypi.org/project/pulumi-kubernetes/4.2.0a1693262085/#files which has both a source distribution and a built (.whl) distribution.

t0yv0:
Verified that --no-binary installs now work:

```
[nix-shell:~/tmp/2023/08/check-pywheel]$ python -m pip  install --no-binary :all: --no-cache-dir pulumi_kubernetes==4.2.0a1693262085 
DEPRECATION: --no-binary currently disables reading from the cache of locally built wheels. In the future --no-binary will not influence the wheel cache. pip 23.1 will enforce this behaviour change. A possible replacement is to use the --no-cache-dir option. You can use the flag --use-feature=no-binary-enable-wheel-cache to test the upcoming behaviour. Discussion can be found at https://github.com/pypa/pip/issues/11453
Collecting pulumi_kubernetes==4.2.0a1693262085
  Downloading pulumi_kubernetes-4.2.0a1693262085.tar.gz (1.3 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.3/1.3 MB 5.3 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... done

```

.github/workflows/build.yml
```diff
@@ -60,6 +60,7 @@ jobs:
       with:
         path: ci-scripts
         repository: pulumi/scripts
+        ref: t0yv0/more-pypi-publish-options
```
danielrbradley, line 63: Assume we'll kill off these refs before merging?

.github/workflows/build.yml
```diff
@@ -490,6 +494,9 @@ jobs:
       run: ./ci-scripts/ci/publish-tfgen-package ${{ github.workspace }}
       env:
         NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        # See https://github.com/pulumi/scripts/pull/138/files
+        # Possible values: "all", "wheel".
+        PYPI_PUBLISH_ARTIFACTS: all
```
danielrbradley, line 499: Why isn't this set to "wheel" now we're publishing the wheels?

scripts/build_python_sdk.sh
```diff
@@ -0,0 +1,52 @@
...
+python -m pip install pip --upgrade
+python -m pip install wheel build
+
+cd ./bin
+
+# Build the source distribution *.tar.gz.
+python -m build . --sdist
+
+# Build the wheel distribution *.whl.
+python -m pip wheel --use-pep517 --no-deps -w dist .
```
danielrbradley, line 52: 1. Is this hand written for this provider? 
2. It looks pretty generic and non-specific to this provider?
3. Could this just be generated by the python codegen instead into `sdk/python/build.sh`?

scripts/build_python_sdk.sh
```diff
@@ -0,0 +1,52 @@
...
+module fake_python_module // Exclude this directory from Go tools
+
+go 1.17
+EOF
+
+rm -rf ./bin/ ../python.bin/
+cp -R . ../python.bin
+mv ../python.bin ./bin
+
+sed -i.bak -e "s/^  version = .*/  version = \"$PYPI_VERSION\"/g" ./bin/pyproject.toml
```
danielrbradley, line 26: Could we make `PYPI_VERSION` an explicit argument to the script and fail if it's not set?

scripts/build_python_sdk.sh
```diff
@@ -0,0 +1,52 @@
...
+cd ./sdk/python
+
+# Removing auto-generated setup.py; this is needed until there is an option on the SDK generator to
+# opt out of generating this file.
+rm setup.py
+
+cat << EOF > go.mod
+module fake_python_module // Exclude this directory from Go tools
+
+go 1.17
```
danielrbradley, line 19: Should this be a newer version now?

danielrbradley (commented):
Direction looks good - would just consider making the build script generated. 

What options have we got for removing the dependency on the `scripts` repo? The main pain point is local testing .. 
- Could we swap to using a submodule?
- Could we just generate the PyPi script as `sdk/python/publish.sh` from the Python codegen?

t0yv0:
> Direction looks good - would just consider making the build script generated.

So kubernetes repo is a bit of a pilot test project to verify this method of building end-to-end. With the epic linked here we'll be trying to push the changes to pyproject.toml into something that Python SDK generator understands, then simplify the scripts here.

.github/workflows/build.yml
t0yv0: So it turns out we have paying customers that rely on source-only installations, as hinted in https://github.com/pulumi/pulumi-kubernetes/issues/2540 we cannot simply remove source distributions and need to publish all for the moment to be backwards-compatible.

scripts/build_python_sdk.sh
t0yv0: Done.

mjeffryes:
> AN option to consider here is to bring this project under ci-mgmt to get a standard Makefile.

Unfortunately, native providers don't have a standard Makefile in ci-mgmt; just workflow files. I don't think the kubernetes workflows or Makefiles are too bespoke right now though, so it's probably possible

mjeffryes (approved):
Looks like a good proof of concept; I agree it would be nice to get the  pyproject.toml into the sdk generation, but that wouldn't get rid of the need for a script since we'd still have to `python -m build .` This is a pretty good example of where it would be nice to have sdk gen+bundle+publish handled in a service.

t0yv0:
> ; I agree it would be nice to get the pyproject.toml into the sdk generation

That's already present in SDK generation and is extended in https://github.com/pulumi/pulumi/pull/13812 and hopefully will be available once that's reviewed.

> This is a pretty good example of where it would be nice to have sdk gen+bundle+publish handled in a service.

Well whether it's deployed in a service or not there ideally should be one system that owns how to do this (how to build the projects). It could be the language plugin even, pulumi-language-python, from some precedent in teaching pulumi-language-java to work with multiple kinds of projects. I think @Frassle had some thoughts along those lines too, for teaching pulumi itself to build these projects and not just generate them.

I'm on a schedule here so trying not to worry about it so we deploy the wheels first and foremost, and we can simplify exactly how it's done ex-post-facto.

Frassle:
> I think @Frassle had some thoughts along those lines too, for teaching pulumi itself to build these projects and not just generate them.

Yeh that's gonna happen.

t0yv0:
Can you point to the epic/design doc?

Frassle:
> Can you point to the epic/design doc?

When it's written. It's currently just that a lot of the building blocks are necessary for matrix testing, and so it's pretty obvious that we can fill in the last few bits on top that later.

t0yv0:
Stubs are good so I can bookmark. It really helps coordinate - appreciate if you could find a moment to make a stub issue. Thank you.

Frassle:
https://github.com/pulumi/pulumi/issues/13865

PR #2566, opened 2023-09-13
Gracefully handle ignoreChanges without user needing to refresh a stack

rquitales:
Note: this is stacked on top of https://github.com/pulumi/pulumi-kubernetes/pull/2570

This PR updates the logic for handling fields added to ignoreChanges. Previously, to accurately ignore a field without hitting an SSA conflict error, the Pulumi user would need to run a refresh BEFORE running `pulumi up` or `pulumi preview`. This is due to the provider using the last known value of the ignored field stored in state when computing what should be set in the ignored field when the resource is sent to the k8s API. If drift has occurred on that field and is not reflected in the Pulumi state as it is out of date, then the provider sends old information to the API server resulting in a field conflict error.

As we hit the live server to obtain the resources' up to date `resourceVersion`, we can also use this API result to gracefully handle fields marked within ignoreChanges. This PR uses the fieldpath package from upstream k8s libraries to parse the serialized json managed fields entries and compares it to a normalized version of the Pulumi program's ignoreFields value.

Fixes: #2542


provider/pkg/await/await.go
```diff
@@ -491,9 +499,69 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
+		}
+		s := fieldpath.NewSet()
+		if err := s.FromJSON(bytes.NewReader(f.FieldsV1.Raw)); err != nil {
+			return fmt.Errorf("unable to parse managed fields from resource %q: %w", c.Inputs.GetName(), err)
+		}
+		managedFieldsSet = managedFieldsSet.Union(s)
+	}
+
+	for _, ignorePath := range c.IgnoreChanges {
+		pathComponents := strings.Split(ignorePath, ".")
```
iwahbe, line 521: It looks like this is the ignore changes Pulumi Resource Option.

If so, valid paths are more complicated than this, for example `foo.bar[3].fizz[*]`. We should use the [library provided in `pulumi/sdk`](https://github.com/pulumi/pulumi/blob/4ebf61f896fe226bf455da3fc87c91abdc33dc28/sdk/go/common/resource/properties_path.go#L49) for parsing.

provider/pkg/await/await.go
```diff
@@ -491,9 +499,69 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
+			if err != nil {
+				return fmt.Errorf("unable to set field %q with last used value %q: %w", ignorePath, lastVal, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func makeInterfaceSlice(inputs []string) []interface{} {
```
iwahbe, line 550: Nit: This may as well be generic.

```suggestion
func makeInterfaceSlice[T any](inputs []T) []interface{} {
```

provider/pkg/await/await.go
```diff
@@ -491,9 +499,69 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
+		if managedFieldsSet.Has(pe) {
+			unstructured.RemoveNestedField(c.Inputs.Object, pathComponents...)
+			continue
+		}
+
+		// We didn't find another field manager that is managing this field, so we need to use the last known value applied to
+		// the cluster so we don't unset it.
+		// NOTE: If the field has been reverted to its default value, ignoreChanges will still not update this field to what is supplied
+		// by the user in their Pulumi program.
+		lastVal, found, err := unstructured.NestedFieldCopy(liveOldObj.Object, pathComponents...)
```
iwahbe, line 538: It looks like we are just ignoring `err` here if it is not nil. If so, can we explain why it is OK to ignore this error?

provider/pkg/await/await.go
```diff
@@ -531,18 +599,18 @@ func fixCSAFieldManagers(c *UpdateConfig, live *unstructured.Unstructured, clien
...
 		if err != nil {
-			return err
+			return nil, err
 		}
 		patch, err := json.Marshal(patches)
 		if err != nil {
-			return err
+			return nil, err
 		}
 		_, err = client.Patch(c.Context, live.GetName(), types.JSONPatchType, patch, metav1.PatchOptions{})
```
EronWright, line 608: Shouldn't we update the `live` variable here rather than discarding the new live object from `Patch`? Imagine that the managers were indeed migrated; you'd want to know that in the subsequent `handleSSAIgnoreFields` step.

If not, then I don't understand why the signature of `fixCSAFieldManagers` was changed to suggest that a new live object is returned.

provider/pkg/await/await.go
```diff
@@ -491,9 +499,69 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
+
+	for _, f := range managedFields {
+		if f.Manager == c.FieldManager {
+			// We don't need to keep track of fields that we are currently managing.
+			continue
+		}
+		s := fieldpath.NewSet()
+		if err := s.FromJSON(bytes.NewReader(f.FieldsV1.Raw)); err != nil {
+			return fmt.Errorf("unable to parse managed fields from resource %q: %w", c.Inputs.GetName(), err)
+		}
```
EronWright, line 516: Nit: consider using `fluxssa.FieldsToSet(f.FieldsV1)`.

provider/pkg/await/await.go
EronWright: To @iwahbe's point, I observe that `Diff` uses `resource.ParsePropertyPath(ignore)` to parse the `IgnoreChanges`.

provider/pkg/await/await.go
```diff
@@ -491,9 +499,69 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
+
+	for _, ignorePath := range c.IgnoreChanges {
+		pathComponents := strings.Split(ignorePath, ".")
+		pe, err := fieldpath.MakePath(makeInterfaceSlice(pathComponents)...)
+		if err != nil {
+			return fmt.Errorf("unable to normalize ignoreField path %q: %w", ignorePath, err)
+		}
+
+		// Drop the field from the inputs if it is present on the cluster and not managed by the current field manager. This ensures
+		// that we don't get any conflict errors, or mistakenly setting the current field manager as a shared manager of that field.
```
EronWright, line 528: This comment makes me wonder about cases where pulumi is (or should be) a shared manager.  Would this code cause Pulumi to (inadvertently) relinquish control over a shared field?

Imagine that replicas was `2` and was shared between two managers. The other manager should have to 'force' the conflict to change replicas to `3`. Pulumi shouldn't prematurely relinquish control.

provider/pkg/await/await.go
```diff
@@ -491,9 +499,69 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
+		if managedFieldsSet.Has(pe) {
```
EronWright, line 529: Is this an exact path match, or a 'contains' match? Am wondering about the case where a manager owns the parent of an ignored field.

provider/pkg/await/await.go
```diff
@@ -491,9 +499,69 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
+			unstructured.RemoveNestedField(c.Inputs.Object, pathComponents...)
+			continue
+		}
+
+		// We didn't find another field manager that is managing this field, so we need to use the last known value applied to
+		// the cluster so we don't unset it.
```
EronWright, line 535: In SSA mode, we should avoid expressing an interest in fields that aren't in our inputs.  Otherwise it would potentially create an undue conflict for other parties.

I suppose I don't understand why this is necessary.

tests/sdk/nodejs/nodejs_test.go
```diff
@@ -1802,3 +1802,104 @@ func simluateClusterKubeconfig(t *testing.T, numOfClusters int) ([]string, error
...
+					depImage, err := tests.Kubectl("get deployment -o=jsonpath='{.spec.template.spec.containers[0].image}' -n", depNS, depName)
+					assert.NoError(t, err)
+					assert.Equal(t, "'nginx:1.25.1'", string(depImage))
+
+					depReplicas, err := tests.Kubectl("get deployment -o=jsonpath='{.spec.replicas}' -n", depNS, depName)
+					assert.NoError(t, err)
+					assert.Equal(t, "'3'", string(depReplicas))
+
+					// Now use kubectl patch to update spec.replicas to 5 and see if we can correctly ignore changes to spec.replicas again when the field manager is
+					// "kubectl-patch" since we have logic to override certain field managers with manager name prefixes.
```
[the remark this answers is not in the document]
rquitales: Yes. `fluxssa.PatchReplaceFieldManagers` in https://github.com/pulumi/pulumi-kubernetes/blob/d2a3b8c217e73c874e294f9f7cbdbef9dfe5e204/provider/pkg/await/await.go#L498C39-L532 matches field manager names with a name prefix rather than an exact match. So Pulumi will take ownership of `kubectl-patch` managed fields, not just fields managed by `kubectl`.
I've expanded the comment to make this clearer.

provider/pkg/await/await.go
rquitales: Good catch! This happened after I rebased on top of merged changes to mainline branch. I've refactored the code slightly in this function to make returning a new unstructured.Unstructured instead of re-using the `live` var name to prevent an error like this from happening again.

provider/pkg/await/await.go
rquitales: Normally I prefer to reimplement upstream libraries to do simple tasks like this, but since we already import and use `fluxssa` I've updated the code to use this function instead.

provider/pkg/await/await.go
rquitales: Updated the logic to keep track if we're a shared manager or not to prevent this from happening.

provider/pkg/await/await.go
```diff
@@ -491,60 +498,147 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
-				OperationType: metav1.ManagedFieldsOperationUpdate,
-			},
-		}, c.FieldManager)
+// handleSSAIgnoreFields handles updating the inputs to either drop fields that are present on the cluster and not managed
+// by the current field manager, or to set the value of the field to the last known value applied to the cluster.
+func handleSSAIgnoreFields(c *UpdateConfig, liveOldObj *unstructured.Unstructured) error {
+	managedFields := liveOldObj.GetManagedFields()
+	// Keep track of fields that are managed by the current field manager, and fields that are managed by other field managers.
+	managedFieldsSet := fieldpath.NewSet()
+	currManagerFieldsSet := fieldpath.NewSet()
```
EronWright, line 507: Consider naming these `theirFields` and `ourFields` for clarity.

provider/pkg/await/await.go
```diff
@@ -491,60 +498,147 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
+		// 2. The field is set/exists on the cluster, but for some reason is not listed in the managed fields, in this case we need to set the field to the last
+		// known value applied to the cluster, otherwise it will be unset. This would cause the current field manager to take ownership of the field, but this edge
+		// case probably shouldn't be hit in practice.
+		//
+		// NOTE: If the field has been reverted to its default value, ignoreChanges will still not update this field to what is supplied
+		// by the user in their Pulumi program.
+		lastVal, found, err := unstructured.NestedFieldCopy(liveOldObj.Object, pathComponents...)
+		if found && err == nil {
+			// We only care if the field is found, as not found indicates that the field does not exist in the live state so we don't have to worry about changing the inputs to match
+			// the live state. If this occurs, then Pulumi will set the field back to the declared value. Or should we also ensure that the field is never touch again by Pulumi?
```
EronWright, line 557: There's a subtle aspect to ignorechanges that I did not realize until I re-read the docs, which say:
> One reason you would use the ignoreChanges option is to ignore changes in properties that lead to diffs. Another reason is to **change the defaults for a property** without forcing all existing deployed stacks to update or replace the affected resource.

I suppose this means that the inputs should be ignored even if we are the field manager. Does that help answer this question, I dunno.

rquitales: Yes, that's the logic I have defined as well. ignoreChanges always ignores what's declared in the Pulumi program unless that field is not present in the actual live resource/state, then we use the declared value.

EronWright (approved):
Good job adding the CSA support!

provider/pkg/await/await.go
```diff
@@ -498,13 +502,43 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
 	return currentOutputs, nil
 }
 
+// handleCSAIgnoreFields handles updating the inputs to use the last known value applied to the cluster. If the value is not present,
+// then we use what is declared in state as per the specs of Pulumi's ignoreChanges.
+func handleCSAIgnoreFields(c *UpdateConfig, liveOldObj *unstructured.Unstructured) error {
```
t0yv0, line 507: I've not a good intuition around  liveOldObj *unstructured.Unstructur and *UpdateConfig but this looks kind of similar to what I've seen bridged providers have to do for ignoreChanges.. It makes sense with he comment 👍

provider/pkg/await/await.go
```diff
@@ -498,13 +502,43 @@ func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
+	for _, ignorePath := range c.IgnoreChanges {
+		ipParsed, err := resource.ParsePropertyPath(ignorePath)
+		if err != nil {
+			// NB: This shouldn't really happen since we already validated the ignoreChanges paths in the parent Diff function.
+			return fmt.Errorf("unable to parse ignoreField path %q: %w", ignorePath, err)
+		}
+
+		pathComponents := strings.Split(ipParsed.String(), ".")
```
t0yv0, line 515: Are we not worried here, this conversion would just work? `ipParsed` will have strings or integers (to drill down arrays), for example:

```
"foo", 0, "barProperty"
```

Does  strings.Split(ipParsed.String(), ".") give you what you expect for the numbers?

t0yv0: Will this ever receive '*' as a path fragment? Pulumi docs permit patterns like this one:

```
foo.*
```

But we're lacking great SDK API to make it easy to implement. I wonder if the user specifies that will it hit this code or be resolved higher up the stack.

provider/pkg/await/await.go
```diff
@@ -527,6 +561,7 @@ func handleSSAIgnoreFields(c *UpdateConfig, liveOldObj *unstructured.Unstructure
 			return fmt.Errorf("unable to parse ignoreField path %q: %w", ignorePath, err)
 		}
 
+		// TODO: Enhance support for ignoreField path to support nested arrays.
```
t0yv0, line 564: Ah here we go again. Might be worth lifting this comment and the split logic into one helper function so it's in one place.

t0yv0 (approved):
This LGTM, I left some comments per Pulumi side but I lack a good intuition yet around this provider.

PR #2568, opened 2023-09-14
Helm Release detect changes to local chart

### LINKED ISSUE #2118: Helm Release does not pickup changes in locally stored templates
### What happened?

When referencing a Chart with a local path, if I apply changes to the templates and run `pulumi up`,  the release is not updated. 

As a workaround, I've set `forceUpdate: true`.

### Steps to reproduce

Using `kind`: 

`kind create cluster --name pulumi`

```typescript
const kind = new k8s.Provider("kind", {context: "kind-pulumi"});
const release = new Release("pulumi-release",{
    name: "pulumi-test",
    createNamespace: true,
    namespace: "pulumi-test",
    chart: "./pulumi-chart-test/",
    forceUpdate: true,
    values: {
        pulumi: "world",
    }
    },{
        provider: kind
    }
);
```

The pulumi-chart-test/templates folder contains a simple manifest and if I add the below:

```patch
apiVersion: v1
kind: ConfigMap
metadata:
  name: pulumi
  annotations:
+    test: "update"
data: 
    hello: {{ .Values.pulumi }}
```

And run `pulumi up` nothing happens.

### Expected Behavior

The Helm release should be updated with any addition to its manifests.

### Actual Behavior

Nothing happened, the release is not updated according to the changes in the underlying manifest.

### Versions used

```
CLI          
Version      3.37.2
Go Version   go1.18.4
Go Compiler  gc

Plugins
NAME        VERSION
kubernetes  3.20.2
nodejs      unknown

Host     
OS       darwin
Version  11.6
Arch     x86_64

This project is written in nodejs: executable='/Users/benjamin/.nvm/versions/node/v16.15.1/bin/node' version='v16.15.1'

Current Stack: helmRelease

pulumi:pulumi:Stack            urn:pulumi:helmRelease::helm::pulumi:pulumi:Stack::helm-helmRelease
pulumi:providers:kubernetes    urn:pulumi:helmRelease::helm::pulumi:providers:kubernetes::kind
kubernetes:helm.sh/v3:Release  urn:pulumi:helmRelease::helm::kubernetes:helm.sh/v3:Release::pulumi-release


Found no pending operations associated with helmRelease

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/btuffreau
User           btuffreau
Organizations  btuffreau, starburstdata

Dependencies:
NAME                VERSION
@pulumi/kubernetes  3.20.2
@pulumi/pulumi      3.37.2
@types/node         14.18.23
```

### Additional context

_No response_

### Contributing

EronWright:
### Proposed changes

This PR fixes a usability problem with the `kubernetes:helm.sh/v3:Release` resource when using a local chart.  With this fix, Pulumi automatically detects changes to the rendered output of the chart and performs a Helm upgrade accordingly. It does this by persisting a SHA256 checksum of the chart output into the Pulumi stack state.

Users may suppress change detection with the `ignoreChanges` option, i.e. `ignoreChanges: ["checksum"]`.

**Upgrade considerations**: after upgrading the provider, users should expect to see a diff (`checksum` property) and a corresponding Helm upgrade. Helm upgrades are generally idempotent.

### Related issues (optional)
Closes #2118

### DIFF
--- a/provider/pkg/provider/helm_release.go
+++ b/provider/pkg/provider/helm_release.go
@@ -682,6 +682,15 @@ func (r *helmReleaseProvider) Diff(ctx context.Context, req *pulumirpc.DiffReque
 
 	// Extract old inputs from the `__inputs` field of the old state.
 	oldInputs, _ := parseCheckpointRelease(olds)
+
+	// if the checkpointed inputs have no checksum, then assume no change since checksums were introduced,
+	// to avoid making a spurious release.
+	checksum, ok := oldInputs["checksum"]
+	if !ok || (checksum.IsString() && checksum.StringValue() == "") {
+		oldInputs["checksum"] = news["checksum"]
+	}
+
+	// Calculate the diff between the old and new inputs
 	diff := oldInputs.Diff(news)
 	if diff == nil {
 		logger.V(9).Infof("No diff found for %q", req.GetUrn())
@@ -1122,7 +1131,10 @@ func (r *helmReleaseProvider) computeResourceInfo(rel *Release, clientSet *clien
 
 	// compute the checksum of the rendered output, to be able to detect changes
 	h := sha256.New()
-	h.Write([]byte(templ))
+	_, err = h.Write([]byte(templ))
+	if err != nil {
+		return nil, err
+	}
 	result.checksum = hex.EncodeToString(h.Sum(nil))
 
 	return result, nil
--- a/provider/pkg/provider/helm_release.go
+++ b/provider/pkg/provider/helm_release.go
@@ -16,6 +16,8 @@ package provider
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -132,6 +134,8 @@ type Release struct {
 	// Manifest map[string]interface{} `json:"manifest,omitempty"`
 	// Names of resources created by the release grouped by "kind/version".
 	ResourceNames map[string][]string `json:"resourceNames,omitempty"`
+	// The checksum of the rendered template (to detect changes)
+	Checksum string `json:"checksum,omitempty"`
 	// Status of the deployed release.
 	Status *ReleaseStatus `json:"status,omitempty"`
 }
@@ -353,19 +357,23 @@ func (r *helmReleaseProvider) Check(ctx context.Context, req *pulumirpc.CheckReq
 			return nil, err
 		}
 
-		resourceNames, err := r.computeResourceNames(new, r.clientSet)
+		resourceInfo, err := r.computeResourceInfo(new, r.clientSet)
 		if err != nil && errors.Is(err, fs.ErrNotExist) {
 			// Likely because the chart is not readily available (e.g. import of chart where no repo info is stored).
 			// Declare bankruptcy in being able to determine the underlying resources and hope for the best
 			// further down the operations.
-			resourceNames = nil
+			resourceInfo = nil
 		} else if err != nil {
 			return nil, err
 		}
 
-		if len(new.ResourceNames) == 0 {
-			new.ResourceNames = resourceNames
+		if resourceInfo != nil {
+			new.Checksum = resourceInfo.checksum
+			if len(new.ResourceNames) == 0 {
+				new.ResourceNames = resourceInfo.resourceNames
+			}
 		}
+
 		logger.V(9).Infof("New: %+v", new)
 		news = resource.NewPropertyMap(new)
 	}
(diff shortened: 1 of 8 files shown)


provider/pkg/provider/helm_release.go
```diff
@@ -674,6 +682,15 @@ func (r *helmReleaseProvider) Diff(ctx context.Context, req *pulumirpc.DiffReque
 
 	// Extract old inputs from the `__inputs` field of the old state.
 	oldInputs, _ := parseCheckpointRelease(olds)
+
+	// if the checkpointed inputs have no checksum, then assume no change since checksums were introduced,
+	// to avoid making a spurious release.
+	checksum, ok := oldInputs["checksum"]
+	if !ok || (checksum.IsString() && checksum.StringValue() == "") {
+		oldInputs["checksum"] = news["checksum"]
+	}
```
EronWright, line 691: This block of code is intended to avoid causing a spurious upgrade of the Helm release, as a side-effect of updating the pulumi-kubernetes dependency in an existing stack. Without this, one sees a change to the `checksum` property (was previously unset). I haven't found a good way to write a test for this, and I tested it manually.

provider/pkg/provider/helm_release.go
```diff
@@ -353,19 +357,23 @@ func (r *helmReleaseProvider) Check(ctx context.Context, req *pulumirpc.CheckReq
...
 			// further down the operations.
-			resourceNames = nil
+			resourceInfo = nil
 		} else if err != nil {
 			return nil, err
 		}
 
-		if len(new.ResourceNames) == 0 {
-			new.ResourceNames = resourceNames
+		if resourceInfo != nil {
```
iwahbe, line 370: Will `r.computeResourceInfo` ever return `resourceInfo = nil` and `err = nil`? If not, then I think this `if` check would be clearer as the `else` block on the previous chain.

provider/pkg/provider/helm_release.go
```diff
@@ -1099,12 +1122,22 @@ func (r *helmReleaseProvider) computeResourceNames(rel *Release, clientSet *clie
...
 
+	// compute the resource names
 	_, resourceNames, err := convertYAMLManifestToJSON(templ)
 	if err != nil {
 		return nil, err
 	}
+	result.resourceNames = resourceNames
+
+	// compute the checksum of the rendered output, to be able to detect changes
+	h := sha256.New()
```
iwahbe, line 1133: Encoding the SHA into state is an efficient way to check for diffs here. I think this will work well.

---

I'm not sure how big the manifests we are talking about are. If they are small, could we just stick the whole JSON blob in state? That way, the engine can provide high quality diffs to our users, instead of just showing that two random strings differ. 

Not a blocker if not viable.

iwahbe (approved):
I have 1 nit and 1 suggestion. I don't have a ton of context on the k8s part, but this seems to do what it claims to.

provider/pkg/provider/helm_release.go
```diff
@@ -132,6 +134,8 @@ type Release struct {
 	// Manifest map[string]interface{} `json:"manifest,omitempty"`
 	// Names of resources created by the release grouped by "kind/version".
 	ResourceNames map[string][]string `json:"resourceNames,omitempty"`
+	// The checksum of the rendered template (to detect changes)
+	Checksum string `json:"checksum,omitempty"`
```
t0yv0, line 138: I've not worked on this provider before so I'm curious if this struct can simply be extended? There's no need to mark it as computed or some such? I see no changes to schema in this PR so this is not visible in the schema? 

Does it become visible in diffs? 

I recently had a quick discussion with Fraser about the use of private state. This is something that we use in bridged providers where the resource is allowed to write opaque blobs in to Pulumi state under __meta flag that are generally excluded from treatment by the engine or any of the plumbing, this is purely for the provider to manage. And I was wondering if any native providers do that and how do they do that.

t0yv0:
LGTM but I've never worked with this repo before :) Deferring to @rquitales

provider/pkg/provider/helm_release.go
EronWright: It is essentially a computed input (i.e. computed in Check). It does become visible in diffs. I suppose it could be useful as a computed output property. Sorry I don't know what the next step here would be.
```
View in Browser (Ctrl+O): https://app.pulumi.com/eron-pulumi-corp/issue-2118/dev/updates/64

     Type                              Name            Status              Info
     pulumi:pulumi:Stack               issue-2118-dev                      2 warnings
 ~   └─ kubernetes:helm.sh/v3:Release  release         updated (0.10s)     [diff: +checksum]
```

t0yv0: That's probably fine then! If we had private state facility you could hide this from users so it's not visible in the diff. But sounds like exposing it is intentional.

provider/pkg/provider/helm_release.go
EronWright: The size is arbitrarily large but I agree that the diff would be more useful.  That said, Helm keeps historical information and even has a `diff` plugin to show what's changed after it is applied. I'm not sure whether pulumi diff can show arbitrary information that isn't directly related to the properties.  I think we should consider a future enhancement with a larger time budget.

provider/pkg/provider/helm_release.go
rquitales: Agree. I think exposing it would provide more value for our users to understand the diff that is occurring.

rquitales (approved):
LGTM!

provider/pkg/provider/helm_release.go
rquitales: There is an upper bound of 1.5 mb I think. Essentially the max size allowed for a k8s object which is ultimately defined by underlying etcd.

provider/pkg/provider/helm_release.go
```diff
@@ -683,6 +683,13 @@ func (r *helmReleaseProvider) Diff(ctx context.Context, req *pulumirpc.DiffReque
 	// Extract old inputs from the `__inputs` field of the old state.
 	oldInputs, _ := parseCheckpointRelease(olds)
 
+	// apply ignoreChanges
+	for _, ignore := range req.GetIgnoreChanges() {
+		if ignore == "checksum" {
+			news["checksum"] = oldInputs["checksum"]
```
rquitales, line 689: Can we add a testcase here for this functionality as well?

EronWright: Done

EronWright:
Note that this fix applies to the `Release` resource specifically.

EronWright:
Note that this change was rolled back in https://github.com/pulumi/pulumi-kubernetes/pull/2672

PR #2591, opened 2023-10-02
Rename Utilities Module

### LINKED ISSUE #2580: Missing PkgVersion function
### What happened?

Since version 4.1.0 the `PkgVersion()` function is missing from the sdk. It seems that https://github.com/pulumi/pulumi-kubernetes/pull/2519 moved the `pulumiUtilites.go` file to an internal pkg but that isn't importable. 

The problem is the tools such as `crd2pulumi` relay on that function when generating the code and the code is broken as of 4.1.0. 

A possible solution (see https://github.com/pulumi/pulumi/pull/13749) would be to rename the internal pkg to something else (e.g. utilities) so it can be importable

### Example

Run `crd2pulumi@1.2.5` on any crd yaml file and observe that the generated code is broken

### Output of `pulumi about`

CLI          
Version      3.85.0
Go Version   go1.21.1
Go Compiler  gc

Plugins
NAME        VERSION
go          unknown
kubernetes  4.0.3

Host     
OS       ubuntu
Version  23.04
Arch     x86_64

This project is written in go: executable='/usr/local/go/bin/go' version='go version go1.21.1 linux/amd64'

*backend and resource list redacted for brevity*

Dependencies:
github.com/bringg/pulib                     1.0.5
github.com/creasty/defaults                 1.7.0
github.com/onsi/ginkgo/v2                   2.12.1
github.com/onsi/gomega                      1.27.10
github.com/pulumi/pulumi-kubernetes/sdk/v4  4.0.3
github.com/pulumi/pulumi/sdk/v3             3.85.0
golang.org/x/exp                            0.0.0-20230811145659-89c5cff77bcb

### Additional context

_No response_

### Contributing

guineveresaenger:
- Set Go's internalModuleName to utilities instead of internal
- Regenerate Go SDK

### Proposed changes

Rename `internal` to `utilities` to make package accessible for dependent libraries

### Related issues (optional)

Fixes #2580


t0yv0:
Does this fix `crd2pulumi@1.2.5` as specified? I don't see a test or any evidence it's now working.

thomas11 (approved):

guineveresaenger:
[crd2pulumi](https://github.com/pulumi/crd2pulumi) is specific to Kubernetes Custom Resource Definitions, so this would be unlikely to apply to other providers.

We could file a ticket against crd2pulumi to no longer rely on `pkgVersion` in quite the same way.

t0yv0:
The issue here is this is an incomplete fix since nothing prevents us from silently regressing on this again.. Please file a ticket in crd2pulumi to add a test that it continues to work, there's obviously a gap here.

PR #2599, opened 2023-10-07
Strict mode for provider kubeconfig handling

EronWright:
### Proposed changes

This PR tightens the kubeconfig loading logic of the provider.
1. ~Errors are not suppressed for the non-default provider (see https://github.com/pulumi/pulumi-kubernetes/pull/960). The rationale is to stabilize the diff when the inputs are unprocessable, without breaking the invoke use-case.~
2. Use a non-interactive loader for the kubeconfig, since the provider is a non-interactive server (see [original PR](https://github.com/pulumi/pulumi-kubernetes/commit/e85e03360ab8b1657d47016b4d38365ebb30e077#diff-5e83e1a18ae8dd9641b49e491350ae5b8afaf7fdb60dfcf4d4447a17e334e473R26)). The rationale is to be consistent with the (newer) Helm code.
3. When the `kubeconfig` property is path-like, don't try to parse it as content. This improves the error message.

TODO: fix test: `TestSkipUpdateUnreachableFlag`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,7 +71,7 @@ Other changes:
 - Upgrade to latest helm dependency (https://github.com/pulumi/pulumi-kubernetes/pull/2474)
 - Improve error handling for List resources (https://github.com/pulumi/pulumi-kubernetes/pull/2493)
 - Normalize provider inputs and make available as outputs (https://github.com/pulumi/pulumi-kubernetes/pull/2598)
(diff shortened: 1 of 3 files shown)


EronWright:
Sorry for any confusion about this PR, it was merged into a parent feature branch, see this PR: https://github.com/pulumi/pulumi-kubernetes/pull/2598

Issue #2600, opened 2023-10-09
Doc about specify helm chart with values file not clear on golang


iluwaa:
### What happened?

Usage of code example from official docs appears to error "cannot use pulumi.NewFileAsset("test") (value of type pulumi.Asset) as pulumi.AssetOrArchiveArrayInput value in struct literal: pulumi.Asset does not implement pulumi.AssetOrArchiveArrayInput (missing method ToAssetOrArchiveArrayOutput)"

### Example

https://www.pulumi.com/registry/packages/kubernetes/api-docs/helm/v3/release/#specify-helm-chart-values-in-file-and-code

copy code from example to IDE 

### Output of `pulumi about`

CLI          
Version      3.87.0
Go Version   go1.21.2
Go Compiler  gc

Plugins
NAME        VERSION
aws         6.2.1
go          unknown
kubernetes  4.3.0

Host     
OS       darwin
Version  13.6
Arch     arm64

This project is written in go: executable='/opt/homebrew/bin/go' version='go version go1.21.1 darwin/arm64'

Current Stack: iluwa/vault_test/sandbox

Found no resources associated with sandbox

Found no pending operations associated with sandbox

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/iluwa
User           iluwa
Organizations  iluwa
Token type     personal

Dependencies:
NAME                                 VERSION
github.com/pulumi/pulumi-aws/sdk/v6  6.2.1
github.com/pulumi/pulumi/sdk/v3      3.86.0

Pulumi locates its logs in /var/folders/gy/_r0bdh9s0j38bt679tc0m8wc0000gq/T/ by default

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mikhailshilkov:
Thank you for reporting this @iluwaa. The example should be fixed with the next provider release.

PR #2601, opened 2023-10-09
Fix examples for helm.Release in Go

mikhailshilkov:
### Proposed changes

I fixed all the examples for helm.Release in Go. They are all manually authored and it looks like they aren't compiling anymore (all but one). I verified by copying snippets to a brand-new Pulumi k8s project.

### Related issues (optional)

Fix https://github.com/pulumi/pulumi-kubernetes/issues/2600


rquitales (approved):
LGTM. It'd be nice if we could also have automated testing of our examples to prevent drift like this.

Issue #2596, opened 2023-10-05
Pulumi import by passing json file is failing


yogeshghotekar:
### What happened?

I am trying to import resource define in json file to pulumi but is failing to import the helm charts
This started failing on pulumi-kubernetes provider v4.3.0 It works fine with v4.2.0

### Example

To reproduce this issue. You can create resources in json file. In my case I am creating this resource json file from my terraform deployed resources and then trying to import it in pulumi.

json file

```
{
    "nameTable": {
        "edge": "urn:pulumi:production::deployer::myinfra:deployer:Cluster::edge",
        "edge-provider": "urn:pulumi:production::deployer::pulumi:providers:kubernetes::edge"
    },
    "resources": [
       {
            "name": "edge/metallb-system",
            "type": "kubernetes:core/v1:Namespace",
            "id": "metallb-system",
            "parent": "edge",
            "provider": "edge-provider"
        },
		{
            "name": "edge/metallb-system/my-secret",
            "type": "kubernetes:core/v1:Secret",
            "id": "metallb-system/my-secret",
            "parent": "edge",
            "provider": "edge-provider"
        },
        {
            "name": "edge/metallb-system/metallb",
            "type": "kubernetes:helm.sh/v3:Release",
            "id": "metallb-system/metallb",
            "parent": "edge",
            "provider": "edge-provider"
        }
    ]
}
```

my resource creation code
```
release, err := helmv3.NewRelease(ctx, fmt.Sprintf("%s/%s/%s", clusterName, helmRelease.Namespace, helmRelease.Name),
                &helmv3.ReleaseArgs{
                    Name:            pulumi.String("metallb"),
                    Chart:           pulumi.String("metallb-system/metallb"),
                    Namespace:       pulumi.StringPtr("metallb-system"),
                    CreateNamespace: pulumi.BoolPtr(true),
                    WaitForJobs:     pulumi.BoolPtr(true),
                    SkipAwait:       pulumi.BoolPtr(false),
                    CleanupOnFail:   pulumi.BoolPtr(true),
                    Atomic:          pulumi.BoolPtr(false),
                    Version:         pulumi.StringPtr("1.18.0"),
                    Timeout:         pulumi.IntPtr(900),
                    Values:          pulumi.ToMap(helmRelease.Values),
                    Lint:            pulumi.BoolPtr(false),
                    Verify:          pulumi.BoolPtr(false),
                    ResetValues:     pulumi.BoolPtr(true),
[...]

PR #2605, opened 2023-10-12
Fix for Helm Import Regression

EronWright:
### Proposed changes

This is a series of commits to fix a p1 and to improve the import of Helm Releases. The regression occurred because `Read` attempted to use the provider-generated inputs to compute a checksum, and returned an error if it couldn't, and this caused the `pulumi import` tool to fail in cases where the inputs aren't guessable.  The fix is to eat the problem during `Read`.

While implementing the fix, it became apparent that the templating code for the `Release` resource should be decoupled from the `Chart` resource, mainly to improve the consistency of how chart names are handled.

Specific changes:
- suppress error in computing inputs during `Read`.
- decouple the templating code.
- simplify the chart linting code.
- use a timeout that is consistent with the Delete op.
- new tests for the import variants.

Closes https://github.com/pulumi/pulumi-kubernetes/issues/2596


provider/pkg/provider/helm_release.go
```diff
@@ -324,6 +322,7 @@ func (r *helmReleaseProvider) Check(ctx context.Context, req *pulumirpc.CheckReq
 	urn := resource.URN(req.GetUrn())
 	label := fmt.Sprintf("Provider[%s].Check(%s)", r.name, urn)
 
+	failures := []*pulumirpc.CheckFailure{}
```
rquitales, line 325: nit: prefer simple declaration (`var failures []*pulumirpc.CheckFailure`) if we don't know the size beforehand.

provider/pkg/provider/helm_release.go
```diff
@@ -432,6 +422,113 @@ func (r *helmReleaseProvider) setDefaults(target resource.PropertyMap) {
...
+
+func (r *helmReleaseProvider) helmTemplate(ctx context.Context, urn resource.URN, newRelease *Release) (string, error) {
+	conf, err := r.getActionConfig(newRelease.Namespace)
+	if err != nil {
+		return "", err
+	}
+	client := action.NewInstall(conf)
+	c, path, err := getChart(&client.ChartPathOptions, conf.RegistryClient, r.settings, newRelease)
+	if err != nil {
+		logger.V(9).Infof("getChart failed: %+v", err)
```
rquitales, line 460: nit: prefer `%v` here to be consistent with other error formatting elsewhere, though functionally the same logged outcome.

provider/pkg/provider/helm_release.go
```diff
@@ -432,6 +422,113 @@ func (r *helmReleaseProvider) setDefaults(target resource.PropertyMap) {
...
+	}
+
+	logger.V(9).Infof("template helm chart")
+	rel, err := client.RunWithContext(r.canceler.context, c, values)
+	if err != nil {
+		return "", err
+	}
+
+	manifest := getManifest(rel, !newRelease.DisableWebhooks, true)
+	return manifest, err
```
rquitales, line 529: nit: err is always nil here, can we return nil instead of err just to make that clearer?

provider/pkg/provider/helm_release.go
```diff
@@ -483,7 +580,11 @@ func (r *helmReleaseProvider) helmCreate(ctx context.Context, urn resource.URN,
 	client.WaitForJobs = !newRelease.SkipAwait && newRelease.WaitForJobs
 	client.Devel = newRelease.Devel
 	client.DependencyUpdate = newRelease.DependencyUpdate
-	client.Timeout = time.Duration(newRelease.Timeout) * time.Second
+	timeout := newRelease.Timeout
```
rquitales, line 583: It looks like this logic occurs three times in this file. Can we refactor it out into a function, eg. `timeoutOrDefault` or something like that?

rquitales (commented):
First pass looks pretty good to me with some minor comments.

provider/pkg/provider/helm_release.go
iwahbe: ➕

provider/pkg/provider/helm_release.go
```diff
@@ -483,7 +580,11 @@ func (r *helmReleaseProvider) helmCreate(ctx context.Context, urn resource.URN,
...
+	if timeout == 0 {
+		timeout = defaultTimeoutSeconds
+	}
+	client.Timeout = time.Duration(timeout) * time.Second
```
iwahbe:
```suggestion
	client.Timeout = timeoutOrDefault(newRelease.Timeout)
```

With `timeoutOrDefault` as:
```go
func timeoutOrDefault(timeout int) time.Duration {
	if timeout == 0 {
		timeout = defaultTimeoutSeconds
	}
	return time.Duration(timeout) * time.Second
}

File: provider/pkg/provider/helm_release.go [hunk 6]
```diff
@@ -360,21 +359,12 @@ func (r *helmReleaseProvider) Check(ctx context.Context, req *pulumirpc.CheckReq
...
-		} else if err != nil {
-			return nil, err
-		}
-
-		if resourceInfo != nil {
-			new.Checksum = resourceInfo.checksum
-			if len(new.ResourceNames) == 0 {
-				new.ResourceNames = resourceInfo.resourceNames
-			}
+		err = r.setComputedInputs(ctx, urn, new)
```
iwahbe, line 362: It's not obvious to me why `helmReleaseProvider.setComputedInputs` should return the failure message for `"chart"`. Can we change the naming so this makes sense?

It sounds like `setComputedInputs` should be called `setChart` from this error message. Reading the body of `setComputedInputs` makes it seem like this should be called `setTemplate`, but maybe helm thinks a chart and a template are the same. I don't know.

The function definitely should not be called `setComputedInputs`, since Pulumi doesn't have computed inputs.

EronWright: The function's purpose is to compute the `checksum` and `resourceNames` properties by rendering the chart. When it fails, the `chart` parameter is the likely culprit, though technically there could be problems with the `repositoryOpts`. I agree the name isn't perfect but your suggestions aren't better.

iwahbe: Ok. I see that `checksum` doesn't appear anywhere in the schema. When a user changes the `chart` property, does `pulumi` show them a diff on the `checksum` property? I'm assuming that the answer is yes and that is what we want. Is that right?

EronWright: It is true that the `checksum` isn't in the schema, because it was not intended for the user to set directly. It does show up in diff. Honestly I don't have a perfect grasp of the nuances, and am seeking opportunities to learn more.

iwahbe: I think there is some room for improvement on the UX,[^1] but nothing I can think of is breaking and this fixes a P1. I'm happy here. I'll leave it to @rquitales for the final approval.

[^1]: In the long term, we probably don't want to expose this to the user at all. It sounds like we should change `checksum` to `__checksum` to tell the engine that it isn't user visible, and override `Diff` so that it shows a diff on `chart` (or the appropriate input property) if `new.__checksum != old.__checksum`. To maintain backwards compatibility we will need to listen for `ingnoreChanges=["checksum"]`, but we can do that easily.

rquitales (approved):

PR #2610, opened 2023-10-16
Improved search functionality for Helm Import

EronWright:
### Proposed changes
This PR enhances the `Release` resource to search harder for a matching chart when importing a release. When Helm installs a chart, it doesn't record any information about the chart location, simply its name. Pulumi attempts to locate a chart:
1. (existing) by looking for an expanded chart directory (e.g. `nginx/`) in the program directory.
2. (new) by looking for a chart archive in the program directory, with or without the version info (e.g. `nginx-15.3.4.tgz`).
3. (new) by looking in the local Helm repositories for a chart, matching the name and version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - helm.v3.ChartOpts: Add KubeVersion field that can be passed to avoid asking the kubernetes API server for the version (https://github.com/pulumi/pulumi-kubernetes/pull/2593)
 - Fix for Helm Import regression (https://github.com/pulumi/pulumi-kubernetes/pull/2605)
+- Improved search functionality for Helm Import (https://github.com/pulumi/pulumi-kubernetes/pull/2610)
 
 ## 4.4.0 (October 12, 2023)
 
--- a/provider/pkg/provider/helm_release.go
+++ b/provider/pkg/provider/helm_release.go
@@ -1305,7 +1305,7 @@ func (r *helmReleaseProvider) importRelease(ctx context.Context, urn resource.UR
 	// Attempt to resolve the chart's origin in either a local or remote repository.
 	// Note that the local chart is not verified.
 	if name, _, found := searchProgramDirectory(hr.Chart.Metadata.Name, hr.Chart.Metadata.Version); found {
-		release.Chart = *name
+		release.Chart = name
 	} else if repo, chart, found := searchHelmRepositories(r.settings, hr.Chart.Metadata.Name, hr.Chart.Metadata.Version); found {
 		// use a local repository reference, rather than reconstructing all the repository opts
 		release.Chart = fmt.Sprintf("%s/%s", repo.Name, chart.Name)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - helm.v3.Release: Improved cancellation support (https://github.com/pulumi/pulumi-kubernetes/pull/2579)
 - Update Kubernetes client library to v0.28.2 (https://github.com/pulumi/pulumi-kubernetes/pull/2585)
 - Fix for Helm Import regression (https://github.com/pulumi/pulumi-kubernetes/pull/2605)
+- Improved search functionality for Helm Import (https://github.com/pulumi/pulumi-kubernetes/pull/2610)
```
rquitales, line 7: Can you rebase this PR on top of the latest master? This line might end up in the wrong place if we don't rebase correctly.

provider/pkg/provider/helm_release.go
```diff
@@ -1441,6 +1455,54 @@ func mapToInterface(in any) any {
 	return in
 }
 
+// searchProgramDirectory implements a best-effort search for a chart in the program directory.
+func searchProgramDirectory(name, version string) (*string, *helmchart.Metadata, bool) {
+	file := fmt.Sprintf("%s.tgz", name)
+	if c, err := loader.LoadFile(file); err == nil {
+		return &file, c.Metadata, true
+	}
+	file = fmt.Sprintf("%s-%s.tgz", name, version)
```
rquitales, line 1464: Should we prefer returning the `name-version.tgz` chart first?

EronWright: That's reasonable.

provider/pkg/provider/helm_release.go
```diff
@@ -1441,6 +1455,54 @@ func mapToInterface(in any) any {
 	return in
 }
 
+// searchProgramDirectory implements a best-effort search for a chart in the program directory.
+func searchProgramDirectory(name, version string) (*string, *helmchart.Metadata, bool) {
```
iwahbe, line 1459: Can you explain the ordering here?

It goes (as implemented):
- `NAME.tgz`
- `NAME-VERSION.tgz`
- `NAME` (dir)
For example, why isn't there a `NAM-VERSION` (dir) option?

Unit tests for this function and `searchHelmRepositories` would be appreciated. It's not clear to me how the go test you added exercises this functionality.

EronWright: I hadn't considered a `NAME-VERSION` directory because I've not encountered it before; the other options stem from ordinary Helm usage, e.g. saving or extracting a chart archive. It is a reasonable extension, I'll try it.

The go test named `Helm Release Import (Tool)` is exercising each of these options by selecting a directory in the initial step, e.g. `step1-local-tar`. As for how it is asserted, the import step fails if the synthesized inputs don't match the program inputs.

rquitales (approved):

Issue #2315, opened 2023-02-16
k8s.core.v1.Service of kind LoadBalancer `status.loadBalancer.ingress` is undefined


KaoruDev:
### What happened?

I enabled the SSA feature on my K8S provider and started getting errors on my Typescript stat:

```typescript
service.status.loadBalancer.ingress.apply(
  (ingresses) => ingresses.map((ingress) => ingress.ip)
)
```

`ingresses` in the above is `undefined`.

### Expected Behavior

`ingresses` is set to `outputs.core.v1.LoadBalancerIngress[]` representing a list of IP addresses.

### Steps to reproduce

1. Create a service with a K8S provider `enableServerSideApply: false`.
2. Attempt to extract the IP address of the LB.
3. Switch `enableServerSideApply: true`, see references to `ingress` throw an error.

### Output of `pulumi about`

```
CLI
Version      3.46.1
Go Version   go1.19.2
Go Compiler  gc

Plugins
NAME    VERSION
nodejs  unknown

Host
OS       darwin
Version  12.6.3
Arch     x86_64
```

### Additional context

Kubernetes lib `3.22.1`
Pulumi version `3.46.1`
Node version: `16.14.2`

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

squaremo:
I tried to reproduce this with the following program (adapted from `pulumi new kubernetes-typescript`):

```
import * as k8s from "@pulumi/kubernetes";
import * as kx from "@pulumi/kubernetesx";

const provider = new k8s.Provider("local", { enableServerSideApply: false });

const appLabels = { app: "nginx" };
const deployment = new k8s.apps.v1.Deployment("nginx", {
    spec: {
        selector: { matchLabels: appLabels },
        replicas: 1,
        template: {
            metadata: { labels: appLabels },
            spec: { containers: [{ name: "nginx", image: "nginx" }] }
        }
    }
}, { provider });

const service = new k8s.core.v1.Service("nginx", {
    spec: {
        selector: appLabels,
        'type': "LoadBalancer",
        ports: [{name: 'http", port: 80}, {name: 'https', port: 443}],
    },
}, { dependsOn: [deployment], provider });

export const ips = service.status.loadBalancer.ingress.apply(
  (ingresses) => ingresses.map((ingress) => ingress.ip)
);

export const name = deployment.metadata.name;
```

[...]

PR #2615, opened 2023-10-18
Fix SSA dry-run when an Apply is called on outputs

rquitales:
### Proposed changes

This PR unsets the empty initialized status subresource from a SSA dry-run to prevent panics during `pulumi preview` when the Pulumi program contains an Apply on a status field. By removing the field completely, the Pulumi engine will be able to defer returning until after the resource is created, which is when the status subresource is populated.

### Related issues (optional)

Fixes: #2557 
Fixes: #2315


provider/pkg/provider/provider.go
```diff
@@ -1911,6 +1911,13 @@ func (k *kubeProvider) Create(
 		initialized = partialErr.Object()
 	}
 
+	// We need to delete the empty status field returned from the API server if we are in
+	// preview mode. Having the status field set will cause a panic during preview if the Pulumi
+	// program attempts to read the status field.
+	if req.GetPreview() {
+		unstructured.RemoveNestedField(initialized.Object, "status")
+	}
```
iwahbe, line 1919: What is the marshaled value of status? I ask because `plugin.MarshalProperties` is called with `SkipNulls: true`, which implies to me that this shouldn't be necessary. Since it is necessary, does `SkipNulls: true` not work?

provider/pkg/provider/provider.go
```diff
@@ -1911,6 +1911,13 @@ func (k *kubeProvider) Create(
...
+
```
EronWright, line 1920: I wonder, should the provider return _unknown_ rather than _undefined_.  When it is a preview, I do believe that the outputs may include unknowns, but I don't know the downstream effects.  WDYT?

tests/sdk/nodejs/preview_test.go
```diff
@@ -142,3 +142,47 @@ func createSAKubeconfig(t *testing.T, saName string) (string, error) {
...
+		assert.NoError(t, destroyErr)
+	})
+
+	// Run a preview and assert no error.
+	err = pt.RunPulumiCommand("preview", "--non-interactive", "--diff", "--refresh", "--show-config")
+	assert.NoError(t, err)
+
+	// Run pulumi up and assert no error creating the resources.
+	err = pt.TestPreviewUpdateAndEdits()
+	assert.NoError(t, err)
```
EronWright, line 187: Could you make some assertions about the value of the `ip` output property?  e.g. that it is unknown?

provider/pkg/provider/provider.go
EronWright: The `status` field is typically not nullable in Kubernetes objects, so I would expect that the `status` is empty, not null, here.

@rquitales could you elaborate on where the panic comes from? Is it in the application code or in the engine? 
`const ip = svc.status.apply((s) => s.loadBalancer.ingress[0].ip);`

iwahbe: I'm almost certain that the panic comes from `s` being `null` (or `undefined`) in the lambda that is run during preview, instead of `svc.status` being unknown and the apply not running.

tests/sdk/nodejs/preview-apply/index.ts
```diff
@@ -0,0 +1,84 @@
...
+            ],
+            selector: {
+                app: "nginx",
+            },
+        },
+    },
+    { provider: k8sProvider }
+);
+
+const ip = svc.status.apply((s) => s.loadBalancer.ingress[0].ip);
```
iwahbe, line 84: To make the intent clearer, can we just assert that this has not run during preview?

provider/pkg/provider/provider.go
rquitales: Yes, the status field returned from a SSA dry run is empty. The panic comes from the application code: `const ip = svc.status.apply((s) => s.loadBalancer.ingress[0].ip)` since it's attempting to grab the IP address.

Without this fix, the status field is populated with an empty, but initialized object. So attempting to access `ingress[0]` will cause a panic. If we nullify the status field, then the panic is averted.

provider/pkg/provider/provider.go
rquitales: In a create dry run, the status field should really be undefined, so my preference would be to just unset the field.

tests/sdk/nodejs/preview-apply/index.ts
rquitales: Do you have an example of how to do this?

iwahbe: I was thinking something like this:

I don't remember the syntax to raise a TS exception, so this might be wrong.
```suggestion
if (pulumi.runtime.isDryRun() && !pulumi.isUnknown(svc.status)) {
    throw new Exception("Failed");
}
```

provider/pkg/provider/provider.go
iwahbe: If `status` will be defined with some value after the create is finished, then I agree with @EronWright that it should be unknown during preview.

EronWright (approved):

EronWright (approved):

tests/sdk/nodejs/preview-apply/index.ts
rquitales: I'm more inclined to having this intent in our test code, rather than within the Pulumi program. This way, the Pulumi program is representative to what our users normally specify within their code.

PR #2617, opened 2023-10-20
Deepcopy unstructured objects for normalization

rquitales:
### Proposed changes

The logic for normalizing Unstructured resources was inadvertently modifying resources in place because the argument was a pointer type. Change the logic to make a copy of the resource and then return a normalized version of the copy. This avoids returning a partially-modified version of the resource if normalization fails partway through the process.

As part of this change, the following updates were made:

1. Make a deep copy of the Unstructured resource rather than modifying the original in-place.
3. Fix invalid unstructured test data. Unstructured objects should be slices of interfaces, not concrete types.


EronWright (approved):
Seems fine to me if you have verified that all the callers are using the return value, as opposed to assuming that `Normalize` mutates in-place.

Issue #2639, opened 2023-10-26
setting ignoreChanges on patch resource causes fields not owned by Pulumi to be wiped out


phillipedwards:
### What happened?

Using a `DaemonsetPatch` resource to update a daemonset property such as `containers[0].image` results in Pulumi patch resource unsetting all other fields present in the daemonset when including ignoreChanges directives on unrelated fields such as spec.metadata.

### Example

1. create an EKS cluster so that kube-proxy daemonset exists
2. [kube-proxy-before-patch.txt](https://github.com/pulumi/pulumi-kubernetes/files/13171680/kube-proxy-before-patch.txt) - base manifest for the daemonset
3. Use the below program
```
package main

import (
	"encoding/json"
	// "github.com/pulumi/pulumi-aws/sdk/v6/go/aws"
	"github.com/pulumi/pulumi-kubernetes/sdk/v4/go/kubernetes"
	appsv1 "github.com/pulumi/pulumi-kubernetes/sdk/v4/go/kubernetes/apps/v1"
	corev1 "github.com/pulumi/pulumi-kubernetes/sdk/v4/go/kubernetes/core/v1"
	metav1 "github.com/pulumi/pulumi-kubernetes/sdk/v4/go/kubernetes/meta/v1"
	p "github.com/pulumi/pulumi/sdk/v3/go/pulumi"
	cfg "github.com/pulumi/pulumi/sdk/v3/go/pulumi/config"
)

func main() {
	p.Run(func(ctx *p.Context) error {

		sr, err := p.NewStackReference(ctx, "org/project/team", nil)
		if err != nil {
			return err
		}

		kubeconfig, err := sr.GetOutputDetails("kubeconfig")
		if err != nil {
			return err
		}

		k, err := json.Marshal(kubeconfig.Value)
		if err != nil {
			return err
		}

		provider, err := kubernetes.NewProvider(ctx, "k8s", &kubernetes.ProviderArgs{
			Kubeconfig: p.String(string(k)),
		})

		if err != nil {
			return err
		}

		c := cfg.New(ctx, "aws")
		region := c.Get("region")

		// account, err := aws.GetCallerIdentity(ctx, nil)
		// if err != nil {
		// 	return err
		// }

		latestContainerImage := p.Sprintf("602401143452.dkr.ecr.%s.amazonaws.com/eks/kube-proxy:v1.28.1-minimal-eksbuild.1", region)

[...]

PR #2640, opened 2023-10-26
Do not patch field managers for Patch resources

rquitales:
### Proposed changes

This pull request addresses the issue where Patch resources unintentionally patch field managers during update operations (ie. when a Patch resource is updated and reused). Such modification results in the Patch resource taking control of all fields managed owned by a field manager with the prefix `kubectl`. This, in turn, leads to unintended unset fields when running `pulumi up`. Patch resources are designed to contain a specific subset of fields for updating, and they should not interfere with the field managers of normal resources. It's important to note that field managers still need to be patched for regular resources, as they facilitate the transition from Client-Side Apply (CSA) to Server-Side Apply (SSA) for existing Pulumi-managed resources.

Additionally, this PR includes a new test to validate this behavior. The test will fail if the logic to avoid field manager patching for Patch resources is not implemented.

### Related issues:
Fixes: #2639

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Fix: Do not patch field managers for Patch resources (https://github.com/pulumi/pulumi-kubernetes/pull/2640)
+
 ## 4.5.1 (October 24, 2023)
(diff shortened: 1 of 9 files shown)


lblackstone (commented):
LGTM

PR #2648, opened 2023-10-30
Helm OCI chart deployment fails in Windows

### LINKED ISSUE #2644: Helm OCI chart deployment fails in Windows
### What happened?

The deployment of a OCI Helm chart fails under Windows (tested with Windows 11) starting with version `3.29.0` of the Kubernetes provider with following message: `The filename, directory name, or volume label syntax is incorrect.`

image

I tested the code in Go and Typescript and both of the time I get the same error. Here the go version during `pulumi up`

image

With `3.28.1` it works fine:

image

OCI support was introduced with #1892 since version `3.16.0` of the provider-``

### Non-Working example 

```json
{
    "name": "pulumi-stacks",
    "main": "index.ts",
    "devDependencies": {
        "@types/node": "^16"
    },
    "dependencies": {
        "@pulumi/pulumi": "^3.0.0",
        "@pulumi/kubernetes": "3.29.0"
    }
}
```

And snippet of code:

```typescript
//...

const ingressController = new kubernetes.helm.v3.Release("kubecost", {
    chart: "oci://public.ecr.aws/kubecost/cost-analyzer",
    namespace: "kubecost",
    createNamespace: true,
    version: "1.106.3",
});
```

Output:

```shell
 pulumi up -y -f
Updating (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/dirien/pulumi-stacks/dev/updates/13

     Type                              Name               Status                  Info
     pulumi:pulumi:Stack               pulumi-stacks-dev  **failed**              1 error
 +   └─ kubernetes:helm.sh/v3:Release  kubecost           **creating failed**     1 error
Diagnostics:
  pulumi:pulumi:Stack (pulumi-stacks-dev):
    error: update failed

  kubernetes:helm.sh/v3:Release (kubecost):
    error: CreateFile oci://public.ecr.aws/kubecost/cost-analyzer: The filename, directory name, or volume label syntax is incorrect.

Outputs:
    clusterName: "myFirstCivoCluster"
  + kubeconfig : [secret]

Resources:
    4 unchanged

Duration: 23s
```

### Output of `pulumi about`

```shell
PS C:\Users\dirien\pulumi-stacks> pulumi about 
CLI
Version      3.91.1
Go Version   go1.21.3
Go Compiler  gc

Plugins
NAME        VERSION
civo        2.3.11
kubernetes  3.29.0
nodejs      unknown

Host
OS       Microsoft Windows 11 Pro
Version  10.0.22621 Build 22621
Arch     x86_64
```

### Additional context

There was a similar error reported back in 2018 with #143, maybe it is a regression?

### Contributing

EronWright:
### Proposed changes

Closes #2644 

This PR fixes support for OCI charts on Windows, by making the code be more consistent with Helm (see [code](https://github.com/helm/helm/blob/main/pkg/action/install.go#L724)). I believe the error comes from the call to `os.Stat` (see Golang implementation which is based on CreateFile, [here](https://go.dev/src/os/stat_windows.go)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Unreleased
+- Fix: Helm OCI chart deployment fails in Windows (https://github.com/pulumi/pulumi-kubernetes/pull/2648)
 
 ## 4.5.4 (November 8, 2023)
 - Fix: Helm Release: chart requires kubeVersion (https://github.com/pulumi/pulumi-kubernetes/pull/2653)
--- a/provider/pkg/provider/helm_release.go
+++ b/provider/pkg/provider/helm_release.go
@@ -1536,15 +1536,12 @@ func getChart(cpo *action.ChartPathOptions, registryClient *registry.Client, set
 func localChart(name string, verify bool, keyring string) (string, bool, error) {
 	fi, err := os.Stat(name)
 	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			return "", false, nil
-		}
-
-		return "", false, err
+		// Helm eats all errors at this point.
+		return "", false, nil
 	}
 
 	// If a folder is of the same name as a chart, use the folder if it contains a Chart.yaml.
-	if err == nil && fi.IsDir() {
+	if fi.IsDir() {
 		if _, err := os.Stat(filepath.Join(name, "Chart.yaml")); err != nil {
 			// This is not a chart directory, so do not error as Helm could still
 			// resolve this as a locally added chart repository, eg. `helm repo add`.


rquitales:
> I believe that we run tests on windows? Can we make sure that the scenario we are addressing is under test.

GHA supports Windows runners, but I believe our providers only tests on Linux environments currently. It makes sense to increase the scope of testing to all supported envs/architectures, but perhaps we need a bigger discussion about the approach for this and any optimizations we might need to undertake?

It'd be great to get Windows testing enabled to ensure we don't have any regressions here, but that is probably a larger orthogonal effort to get it landed. I'd be happy if we can manually validate that this fix works within a Windows environment to get it submitted for now.

iwahbe:
> > I believe that we run tests on windows? Can we make sure that the scenario we are addressing is under test.
> 
> GHA supports Windows runners, but I believe our providers only tests on Linux environments currently. It makes sense to increase the scope of testing to all supported envs/architectures, but perhaps we need a bigger discussion about the approach for this and any optimizations we might need to undertake?

I agree.

> It'd be great to get Windows testing enabled to ensure we don't have any regressions here, but that is probably a larger orthogonal effort to get it landed. I'd be happy if we can manually validate that this fix works within a Windows environment to get it submitted for now.

We still need to test that `localChart` swallows all errors here, not just file not found. We can do this on unix systems (permission errors come to mind). Otherwise I worry about someone accidentally reverting this change. Since we don't test on windows here, we can keep the integration test manual.

EronWright:
Thanks both for the discussion. I'll provision a Windows VM to test this PR, and will incorporate the notion of a cross-platform test pass into a future "Release Criteria" design doc.

EronWright:
I tested the fix on a Windows PC, seems to work well.

rquitales (approved):
LGTM. Thanks for validating on Windows.

Issue #2563, opened 2023-09-13
Support explicitly passing --kube-version to helm 


henry-fn:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

https://github.com/pulumi/pulumi-kubernetes/issues/753 explained this issue well, but the resolution does not cover one edge case that we cared about. We have a workflow where we would like to run previews without access to the actual K8s API server (just based on pulumi state which IIUC in the case of helm charts should only use helm template). 

https://github.com/pulumi/pulumi-kubernetes/issues/753#issuecomment-875148955 asked for a way to explicitly set `kubeVersion` in the helm args which would have solved our issue as we can explicitly pass this value when running in this semi-offline mode, however the [resolution](https://github.com/pulumi/pulumi-kubernetes/issues/196) to the previous issue was to auto-derive it from the server which doesn't work in our case. Could a `kubeVersion` argument still be added to the helm release / chart options to allow explicit override in these cases?

### Affected area/feature

helm
sdk

henry-fn:
Mistakenly put a `closes` comment in my fork's MR

hypesystem:
Did this make it into Releases to, or just Chart? I'm seeing the same issue using an imported Release, and am not familiar enough with the code here to understand what it affects :smile:

EronWright:
I believe the fix was developed for `Chart` only. @rquitales I am re-opening this to consider a fix for `Release`.

hypesystem:
If there is anything I can help with (including writing code), please let me know, and I'll be happy to contribute in any way I can to help move it forward :smile:

hypesystem:
I've been looking into the code, and adding KubeVersion seems like it would be pretty straight-forward. I've noticed that the automatic discovery of api versions and kubeversion is also only implemented for chart, and not release.

This has lead to my confusing situation of importing a release from a running cluster, that now cannot be previewed or upped because the validation assumes a wrong kubeVersion (from local helm install).

I think it would make sense to try and align the two, maybe even reuse some common code for the templating in the two cases, to minimize the risk of them diverging in the future?

omidraha:
I think I have related issue when I use `pulumi` to install `longhorn` by  `helm`  chart on 'k3s' : 
https://github.com/longhorn/longhorn/issues/4788#issuecomment-1792937587

[...]

PR #2593, opened 2023-10-03
#2563: Expose Helm --kube-version argument for contextless diffs

henry-fn:
### Proposed changes

This PR exposes the helm template `--kube-version` argument so it can be manually specified in situations when there is no access to the Kubernetes API server (e.g running a diff in CSA mode with no connection to the cluster).

Prior to this merge:
* If there were zero k8s contexts --> this field was unset meaning helm would default to some old version (1.20.0 at the time of writing) if it couldn't connect to the k8s context
* 1+ k8s contexts --> would always try to contact the API server to get the version leading to an error if the API server was unreachable.

This variable will override the behavior in either case, but is mostly useful in the first case or when the context in the second case is malformed / unreachable. 

### Related issues (optional)

Closes #2563

### Notes:

* The files under sdk/* are generated from provider/pkg/gen, please review the files under provider/pkg/gen
* Most changes under provider/pkg/gen seemed straightforward but I did have some trouble with the dotnet code since the tuple in `Unpack.cs` was already at max length. I tried to consolidate the args in a similar style to the rest of the unpack logic but it is likely that this section in particular deserves some extra time whenever someone reviews this code.

Issue #2649, opened 2023-11-02
Helm Chart Trigging a Diff on Each Run for Checksum Changes


zbuchheit:
### What happened?

When attempting to use helm.v3.release, each run is reporting a diff on the checksum property which is unexpected.

### Example

```
const alb = new k8s.helm.v3.Release("alb", {
    chart: "aws-load-balancer-controller",
    namespace: 'default',
    version: "1.6.2",
    skipAwait: false,
    repositoryOpts: {
        repo: "https://aws.github.io/eks-charts",
    },
    values: {
        clusterName: cluster.eksCluster.name
    }
}, { 
    dependsOn: [cluster],
    provider: k8sProvider
});
```
each additional pulumi up after resource creation will result in

```
     Type                              Name                  Plan       Info
     pulumi:pulumi:Stack               aws-ts-eks-zbuchheit             
 ~   └─ kubernetes:helm.sh/v3:Release  alb                   update     [diff: ~checksum]
```

### Output of `pulumi about`

```
CLI          
Version      3.90.1
Go Version   go1.21.3
Go Compiler  gc

Plugins
NAME        VERSION
aws         6.0.3
aws         5.42.0
aws         5.31.0
awsx        1.0.5
docker      3.6.1
eks         1.0.3
kubernetes  4.3.0
kubernetes  3.30.2
nodejs      unknown

Host     
OS       darwin
Version  14.0
Arch     arm64
```

### Additional context

Ignore changes allows for a workaround of this issue, but obviously a diff on the checksum on each subsequent up is not a desired result, especially when attempting to use `--expect-no-changes` in a ci/cd environment. 

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mikhailshilkov:
@EronWright Could you please take a look?

rhuanbarreto:
the latest version fixed this. Please try.

mikhailshilkov:
@zbuchheit Indeed, this should have been fixed already, can you give the latest version a try?

zbuchheit:
I am still experiencing this behavior consistently with the provided code snippet.

[...]

PR #2654, opened 2023-11-04
Helm Release: kubeVersion/apiVersions properties

EronWright:
### Proposed changes

This PR is a follow-up to https://github.com/pulumi/pulumi-kubernetes/pull/2593 to add `kubeVersion` and `apiVersions` to the Helm Release object.

Note that the properties are applicable only during templating; Helm seems to always use server detection in the actual install/upgrade operation.

~A possible enhancement to be considered is to automatically incorporate the CRD information into the `Capabiltiies.APIVersions` that are used during templating. Imagine that the chart includes some CRDs that aren't yet installed on the server, what is the templating behavior?  Helm itself doesn't seem to solve it.~

Fixes: https://github.com/pulumi/pulumi-kubernetes/issues/2563

### DIFF
--- a/provider/pkg/gen/overlays.go
+++ b/provider/pkg/gen/overlays.go
@@ -679,6 +679,21 @@ var helmV3ReleaseResource = pschema.ResourceSpec{
 				},
 				Description: "Whether to allow Null values in helm chart configs.",
 			},
+			"apiVersions": {
+				TypeSpec: pschema.TypeSpec{
+					Type: "array",
+					Items: &pschema.TypeSpec{
+						Type: "string",
+					},
+				},
+				Description: "The optional Kubernetes API versions used for Capabilities.APIVersions. By default is detected from the server.",
+			},
+			"kubeVersion": {
+				TypeSpec: pschema.TypeSpec{
+					Type: "string",
+				},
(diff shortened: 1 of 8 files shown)


EronWright:
@rquitales do you happen to know the meaning of this block, and whether the new properties should be listed there?
https://github.com/pulumi/pulumi-kubernetes/pull/2654/files#diff-81d686e8bffdb383464867f6848af2152fe29b75fd70da67eaf1e8603d4e4207L689-L704

rquitales:
> @rquitales do you happen to know the meaning of this block, and whether the new properties should be listed there? https://github.com/pulumi/pulumi-kubernetes/pull/2654/files#diff-81d686e8bffdb383464867f6848af2152fe29b75fd70da67eaf1e8603d4e4207L689-L704

This is a NodeJS specific codegen extension to mark the output as a required output. Looking at all the other types, we should indeed be adding new properties here.

rquitales (changes requested):
Would be great if we could have some tests for these changes as well. Thanks.

EronWright:
Am closing as obsoleted by #2649; the `Release` resource won't do templating, and the overrides weren't used by Helm in the actual install/upgrade procedure anyway.

Issue #2638, opened 2023-10-25
Installing Helm v3 Chart resource causes errors


scottslowe:
### What happened?

I was trying to use/update the [`kubernetes-go-helm-wordpress` example](https://github.com/pulumi/examples/tree/master/kubernetes-go-helm-wordpress) with the Kubernetes v4 provider. I kept getting this error:
```
error: an unhandled error occurred: program exited with non-zero exit code: 2

    panic: interface conversion: interface {} is nil, not *v1.Service
    goroutine 119 [running]:
    main.main.func1.1({0x0?, 0x0?})
    	/Users/slowe/Work/Code/testarea/main.go:24 +0x5c
```
Through additional testing, @rquitales verified that this is broken if the chart is installed into the "default" namespace, and is broken in preview for all namespaces. 

### Example

1. Stand up a Kubernetes cluster and obtain the Kubeconfig.
2. Clone or copy down the `kubernetes-go-helm-wordpress` example with the Kubernetes v4 provider.
3. With the Kubeconfig from step 1 active, run `pulumi up`. It should error during preview.

### Output of `pulumi about`

```
CLI
Version      3.88.0
Go Version   go1.21.2
Go Compiler  gc

Plugins
NAME    VERSION
aws     6.5.0
awsx    2.0.2
docker  4.4.3
go      unknown

Host
OS       darwin
Version  13.5.2
Arch     arm64
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
We omit the namespace if it's `default` in the GetResource function: https://github.com/pulumi/pulumi-kubernetes/blob/f3498a9f3fccb9620cd9e46b6a09fce82a51444d/provider/pkg/gen/_go-templates/helm/v3/chart.go#L351-L353

We always need to include the namespace in generating the key. We should also double check the implementation for ConfigFile resources, and ensure it works as expected too!

PR #2655, opened 2023-11-07
Fix Helm Chart resource lookup key handling for objects in default namespace

rquitales:
### Proposed changes

This pull request modifies the `GetResource` Go method for the Helm Chart resource. Previously, when a resource was deployed to the default namespace, the `default` namespace was always omitted from the resource lookup key. However, in cases where the namespace is explicitly defined in the Helm chart template, the `default` namespace was inadvertently excluded from the resource lookup key despite it being needed. This behavior in Helm is documented in the upstream issue: https://github.com/helm/helm/issues/3553.

**Changes Made:**
- Implemented a fallback mechanism to include the `default` namespace in the resource lookup key when necessary for the `GetResource` method.

**Test Added:**
- Added a test (`ChartGetResource`) to verify the `GetResource` method for both types of Helm charts—those with the explicitly defined default namespace and those without.

**Verification:**
- Verified that the modified `GetResource` method successfully handles scenarios where the namespace is explicitly defined in the Helm chart template.
- Confirmed that the added test covers both types of Helm charts and fails when the fallback logic is reverted.

### Related issues (optional)

Fixes: #2638

### DIFF
--- a/provider/pkg/gen/_go-templates/helm/v3/chart.go
+++ b/provider/pkg/gen/_go-templates/helm/v3/chart.go
@@ -351,9 +351,23 @@ func (c *Chart) GetResource(gvk, name, namespace string) pulumi.AnyOutput {
 	if len(namespace) > 0 && namespace != "default" {
 		id = fmt.Sprintf("%s/%s", namespace, name)
 	}
+
 	key := fmt.Sprintf("%s::%s", gvk, id)
+
 	return c.Resources.ApplyT(func(x interface{}) interface{} {
 		resources := x.(map[string]pulumi.Resource)
-		return resources[key]
+		if r, ok := resources[key]; ok {
+			return r
+		}
+
+		// Some resources in the default namespace do not adhere to the name only pattern, so we need to fallback
+		// to searching for the resource by namespace and name. This occurs when the Helm template manifest explicitly
+		// sets the namespace field on the resource.
+		if namespace == "default" || namespace == "" {
+			return resources[fmt.Sprintf("%s::default/%s", gvk, name)]
+		}
+
+		return nil
+
 	}).(pulumi.AnyOutput)
 }
--- a/tests/sdk/go/helm-get-default-namespace/step2/local-chart/templates/deployment.yaml
+++ b/tests/sdk/go/helm-get-default-namespace/step2/local-chart/templates/deployment.yaml
@@ -1,0 +1,37 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "local-test-chart.fullname" . }}-new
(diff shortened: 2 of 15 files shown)


provider/pkg/gen/_go-templates/helm/v3/chart.go
```diff
@@ -351,9 +351,23 @@ func (c *Chart) GetResource(gvk, name, namespace string) pulumi.AnyOutput {
...
-		return resources[key]
+		if r, ok := resources[key]; ok {
+			return r
+		}
+
+		// Some resources in the default namespace do not adhere to the name only pattern, so we need to fallback
+		// to searching for the resource by namespace and name. This occurs when the Helm template manifest explicitly
+		// sets the namespace field on the resource.
+		if namespace == "default" || namespace == "" {
+			return resources[fmt.Sprintf("%s::default/%s", gvk, name)]
```
rquitales, line 367: The fallback logic to also explicitly specify `default` namespace.

tests/sdk/go/helm-get-default-namespace/step2/local-chart/templates/deployment.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "local-test-chart.fullname" . }}-new
+  namespace: {{ .Release.Namespace }} # This is what causes the difference when needing to specify a namespace or not during templating.
```
rquitales, line 5: This is the helm template format that requires explicit specification of `default` in our resource lookup key.

provider/pkg/gen/_go-templates/helm/v3/chart.go
```diff
@@ -351,9 +351,23 @@ func (c *Chart) GetResource(gvk, name, namespace string) pulumi.AnyOutput {
 	if len(namespace) > 0 && namespace != "default" {
 		id = fmt.Sprintf("%s/%s", namespace, name)
 	}
+
 	key := fmt.Sprintf("%s::%s", gvk, id)
+
```
blampe, line 356: Maybe more of a nit. It seems like we're already trying to account for the default namespace here -- could we also set `key = fmt.Sprintf("%s::default/%s", gvk, name` here instead of below?

blampe (approved):

provider/pkg/gen/_go-templates/helm/v3/chart.go
rquitales: This is the format for explicit namespaces. It is in the form of `gvk::namespace/name`.
For the default namespace, we could have `gvk::default/name` or `gvk::name`. Notice the potential lack of `default` here. So there's 2 possible keys for an object in the default namespace, hence the fallback in the below lines.

PR #2665, opened 2023-11-10
Fix: Make the invoke calls for Helm charts and YAML config resilient to the value being `None` or an empty dict

### LINKED ISSUE #2664: Python: Error raised when using Helm Chart or YAML ConfigFile with an unconfigured provider with `pulumi` 3.92.0
More details here: https://github.com/pulumi/pulumi/issues/14508

**Summary:**

When a provider returns an empty result from an `invoke`, the Python SDK was returning `None` for the value, which generated provider SDKs can’t handle (https://github.com/pulumi/pulumi/issues/13985). So we fixed this in the core SDK with https://github.com/pulumi/pulumi/pull/14470 which was released in v3.92.0 to return an empty dict rather than `None` for empty results, which aligns the behavior with other language SDKs (e.g. Node.js and Go) and is actually the original behavior of the Python SDK (it unintentionally regressed to return `None` in the v2 timeframe).

The change in v3.92.0 breaks the hand-rolled `invoke` code for Helm charts and YAML in the Kubernetes Python SDK.

We want to keep the behavior in the core `pulumi` Python SDK as it is the right long-term behavior.

To address this, we'll do two things:

1. Fix the hand-rolled `invoke` code in the Kubernetes Python SDK to be resilient to getting either `None` or an empty dict for empty results. Basically, by applying the same change as https://github.com/pulumi/pulumi-kubernetes/commit/a907bddce8da5141992c2c9946e2f6be3fe9f79a to Helm and YAML.
2. We'll add a fix in the core `pulumi` SDK that makes it return `None` when there are empty results and the invoke is for one of the affected `kubernetes` functions.

justinvp:
Note: The changes are broken up into multiple commits for easier review.

### Proposed changes

In v3.92.0 of the core `pulumi` Python SDK, invokes were changed to return an empty dict rather than `None` for empty results, which is consistent with other language SDKs and the original behavior of the Python SDK, and makes generated provider Python SDKs work with empty results. However, the hand-rolled invoke code in the Kubernetes Python SDK for Helm and YAML are not resilient to it getting an empty dict. This change fixes Helm and YAML to be resilient, in the same way that the invoke for Kustomize was updated in a907bddce8da5141992c2c9946e2f6be3fe9f79a.

An invoke will return empty results when called on an unconfigured provider (i.e. a provider whose configuration contains an unknown value). The first three commits add regression tests that pass an unknown value as a provider's configuration. The Helm and YAML regression tests fail before the fix and succeed after the fix. The test for Kustomize already succeeds because of the change in a907bddce8da5141992c2c9946e2f6be3fe9f79a. The new tests are based on existing tests.

To ensure older versions of `pulumi-kubernetes` continue to work with newer versions of `pulumi`, we're planning to patch the `pulumi` SDK to return `None` for empty invoke returns when the function is one of the affected kubernetes function tokens. This is tracked by https://github.com/pulumi/pulumi/issues/14508 (PR: https://github.com/pulumi/pulumi/pull/14535).

### Related issues

Fixes #2664


rquitales (approved):

tests/sdk/python/yaml-test-unconfigured-provider/__main__.py
```diff
@@ -0,0 +1,53 @@
...
+# An error shouldn't be raised when called using the unconfigured provider.
+cf_local = ConfigFile(
+    "yaml-test",
+    "manifest.yaml",
+    transformations=[
+        set_namespace(ns),
+        secret_status,
+    ],
+    opts=ResourceOptions(provider=provider),
+)
```
EronWright, line 53: nit: add a `ConfigGroup` resource (which also uses `invoke_yaml_decode` and seems like it is fixed with this PR).

tests/sdk/python/helm-local-unconfigured-provider/__main__.py
```diff
@@ -0,0 +1,29 @@
...
+# This will be unknown during the initial preview.
+unknown = Provider("provider").id.apply(lambda _: True)
+
+# This provider will be unconfigured when the passed-in configuration has an unknown value.
+provider = Provider("k8s", suppress_deprecation_warnings=unknown)
+
+values = {"service": {"type": "ClusterIP"}}
+
+# An error shouldn't be raised when called using the unconfigured provider.
+chart = Chart("nginx", LocalChartOpts(path="nginx", values=values), opts=ResourceOptions(provider=provider))
```
EronWright, line 29: nit: there is a well-known nginx chart in `<repo>/tests/testdata/helm/nginx` that could be used in lieu of a local copy.

EronWright (approved):

PR #2672, opened 2023-11-17
compute version field in Check, use 3-way diff

### LINKED ISSUE #2649: Helm Chart Trigging a Diff on Each Run for Checksum Changes

EronWright:
### Proposed changes

This PR stabilizes Helm upgrades **by abandoning the use of checksums** (which was added in v4.5.0, see [PR](https://github.com/pulumi/pulumi-kubernetes/pull/2568)), while allowing for local chart change detection via the chart's version number.  That is, the user would be expected to touch the chart's version to compel Pulumi to roll out the change.

Implementation-wise, this is accomplished by resolving the `version` property during `Check`.  The checked input reflects the goal version, and if the goal has changed, then `Diff` detects this in the natural way and effects an upgrade.  `Check` no longer uses `helm template` to render the chart, it simply resolves the chart version.

The diff logic is improved here to use a three-way merge, to be able to detect drift in the actual Helm release.

The refresh logic is improved to not mutate inputs (as was done in the past to help with drift detection).

### Demonstration
This video clip demonstrates:
1. An initial deployment using the current version of the provider.
2. A repro of the issue that Pulumi sees a spurious change to `checksum`.
3. Upon building the new provider, the spurious change is gone because the `checksum` property is gone and the diff logic for `version` property compares new input to old state.
4. Upon a change to the local chart's version, Pulumi sees a change to `version`.
5. Drift detection, by using `helm upgrade` and then `pulumi refresh` such that Pulumi sees the change.

Note that the demo app doesn't set the `version` property, it relies on the new logic in `Check` to resolve the version.

https://github.com/pulumi/pulumi-kubernetes/assets/1775518/c405077e-d55b-47c9-ba34-87822eb86009

### Summary of Old/New Behavior

Old:
Action | Op | Raw Input | Checked | Old (Input) | Old (State) | New (State)
-- | -- | -- | -- | -- | -- | --
Deploy 3.47.1 | Check | ~3.47.0 | ~3.47.0 | - | - | -
  | Create | - | ~3.47.0 | - | - | 3.47.1
Upgrade to 3.47.2 | Check | ~3.47.0 | ~3.47.0 | - | - | -
  | Diff | - | ~3.47.0 | ~3.47.0 | 3.47.1 | -
  | ~Update~ | - | - | - | - | -
Drift to 3.47.3 | Read | - | __3.47.3__ | - | - | 3.47.3
  | Check | ~3.47.0 | ~3.47.0 | 3.47.3 | - | -
  | Diff | - | ~3.47.0 | 3.47.3 | 3.47.3 | -

New:
Action | Op | Raw Input | Checked | Old (Input) | Old (State) | New (State)
-- | -- | -- | -- | -- | -- | --
Deploy 3.47.1 | Check | ~3.47.0 | 3.47.1 | - | - | -
  | Create | - | 3.47.1 | - | - | 3.47.1
Upgrade to 3.47.2 | Check | ~3.47.0 | 3.47.2 | - | - | -
  | Diff | - | 3.47.2 | 3.47.1 | 3.47.1 | -
  | Update | - | 3.47.2 | 3.47.1 | 3.47.1 | 3.47.2
Drift to 3.47.3 | Read | - | 3.47.2 | - | - | 3.47.3
  | Check | ~3.47.0 | 3.47.2 | 3.47.2 | - | -
  | Diff | - | 3.47.2 | 3.47.2 | 3.47.3 | -

### Related issues (optional)

Closes #2649


provider/pkg/provider/helm_release.go
```diff
@@ -873,9 +795,11 @@ func (r *helmReleaseProvider) Diff(ctx context.Context, req *pulumirpc.DiffReque
 					"converting JSON patch describing resource changes to a diff",
 				newRelease.Namespace, newRelease.Name)
 		}
-		for _, v := range detailedDiff {
-			v.InputDiff = true
-		}
+
```
rquitales, line 798: Let's delete this dead code.

provider/pkg/provider/helm_release.go
```diff
@@ -1758,3 +1683,23 @@ func getTimeoutOrDefault(timeout int) time.Duration {
 	}
 	return time.Duration(timeout) * time.Second
 }
+
```
rquitales, line 1686: Can we add a brief godoc here to describe why no schema (ie. empty return in the implemented functions) is sufficient in our use case?

provider/pkg/provider/helm_release.go
```diff
@@ -361,19 +359,24 @@ func (r *helmReleaseProvider) Check(ctx context.Context, req *pulumirpc.CheckReq
...
-		err = r.setComputedInputs(ctx, urn, new)
+		chart, err := r.helmBuild(ctx, urn, new)
 		if err != nil {
-			// setComputedInputs fails when the chart cannot be rendered, and we report it as a problem with the `chart` input.
 			failures = append(failures, &pulumirpc.CheckFailure{
 				Property: "chart",
 				Reason:   fmt.Sprintf("%v; check the chart name and repository configuration.", err),
 			})
+		} else {
+			new.Version = chart.Metadata.Version
```
rquitales, line 369: Let's add a quick comment to document why we want the new input's version to be updated with the "computed" version.

provider/pkg/provider/helm_release.go
```diff
@@ -819,34 +741,34 @@ func (r *helmReleaseProvider) Diff(ctx context.Context, req *pulumirpc.DiffReque
...
-	logger.V(9).Infof("oldInputsJSON: %+v", string(oldInputsJSON))
 	logger.V(9).Infof("newInputsJSON: %+v", string(newInputsJSON))
-	patch, err := jsonpatch.CreateMergePatch(oldInputsJSON, newInputsJSON)
+	oldStateJSON, err := json.Marshal(olds.MapRepl(nil, mapReplStripSecrets))
 	if err != nil {
-		return nil, err
+		return nil, pkgerrors.Wrapf(err, "internal error: json.Marshal(oldStateJson)")
+	}
+	logger.V(9).Infof("oldStateJson: %+v", string(oldStateJSON))
+	strategicPatch, err := strategicpatch.CreateThreeWayMergePatch(oldInputsJSON, newInputsJSON, oldStateJSON, schema, true)
```
rquitales, line 761: nit: Could we just do `&noSchema{}` here and get rid of L745?

provider/pkg/provider/helm_release.go
```diff
@@ -819,34 +741,34 @@ func (r *helmReleaseProvider) Diff(ctx context.Context, req *pulumirpc.DiffReque
...
+	if err != nil {
+		return nil, pkgerrors.Wrapf(err, "internal error: CreateThreeWayMergePatch")
 	}
+	logger.V(1).Infof("strategicPatch: %+v", string(strategicPatch))
```
rquitales, line 765: Let's update it to `%s` since it's a string output. I believe linters would catch it if we have a type mismatch incase we accidentally drop the `string()` process in the future.

provider/pkg/provider/helm_release.go
```diff
@@ -378,7 +376,6 @@ func (r *helmReleaseProvider) Check(ctx context.Context, req *pulumirpc.CheckReq
 	}
 
 	// remove deprecated inputs
-	delete(news, "checksum")
```
rquitales: I assume the removal of this line is safe and won't cause erroneous diff'ing?

rquitales (approved):
LGTM

PR #2677, opened 2023-11-28
Fix: Helm Release fails with "the server could not find the requested resource"

### LINKED ISSUE #2481: helm Release tries to call non-existent resources
### What happened?

We have some helm charts that we use with pulumi.

When trying to install the releases to our GKE cluster, we sometimes (non-reliably) receive the `the server could not find the requested resource` error.

Looking in the messages in the Google Cloud Explorer we can see that it tried to call a method named `io.k8s.core.v1.virtualservices.create`:

```json
{
  "protoPayload": {
    "@type": "type.googleapis.com/google.cloud.audit.AuditLog",
    "authenticationInfo": {
    },
    "authorizationInfo": [
      {
        "granted": true,
        "permission": "io.k8s.core.v1.virtualservices.create",
        "resource": "core/v1/namespaces/sv-2/virtualservices"
      }
    ],
    "methodName": "io.k8s.core.v1.virtualservices.create",
    "requestMetadata": {
      "callerIp": "",
      "callerSuppliedUserAgent": "Go-http-client/2.0"
    },
    "resourceName": "core/v1/namespaces/sv-2/virtualservices",
    "response": {
      "@type": "core.k8s.io/v1.Status",
      "apiVersion": "v1",
      "code": 404,
      "details": {},
      "kind": "Status",
      "message": "the server could not find the requested resource",
      "metadata": {},
      "reason": "NotFound",
      "status": "Failure"
    },
    "serviceName": "k8s.io",
    "status": {
      "code": 5,
      "message": "the server could not find the requested resource"
    }
  },
  "insertId": "...",
  "resource": {
    "type": "k8s_cluster",
    "labels": {
      "cluster_name": "...",
      "location": "us-central1",
      "project_id": "..."
    }
  },
  "timestamp": "2023-06-27T09:26:24.949679Z",
  "labels": {
    "authorization.k8s.io/reason": "access granted by IAM permissions.",
    "authorization.k8s.io/decision": "allow"
  },
  "logName": "...",
  "operation": {
    "id": "...",
    "producer": "k8s.io",
    "first": true,
    "last": true
  },
  "receiveTimestamp": "2023-06-27T09:26:29.744157Z"
}
```

We have several such entries for that method in our logs over the past 14 days, but *all* of them fail with error code 5, "the server could not find the requested resource".

But we do have calls to the method `io.istio.networking.v1alpha3.virtualservices.create` which return successfully. This looks to me like some kind of race condition that puts together the wrong API prefix plus resource name and operation.

Calling `pulumi up` again usually gets rid of the problem (at least after some retries).

We also have more such method calls that fail in the same way, such as `io.k8s.core.v1.deployments.create` (which maybe should have been calls to `io.k8s.apps.v1.deployments.create` instead).

### Expected Behavior

There should be no errors and `pulumi up` should work reliably.

### Steps to reproduce

This is a bit hard to reproduce. I'll add more details or a minimal reproducer when possible.

### Output of `pulumi about`

CLI          
Version      3.72.1
Go Version   go1.20.5
Go Compiler  gc

Plugins
NAME    VERSION
nodejs  unknown

Host     
OS       nixos
Version  23.05 (Stoat)
Arch     x86_64

This project is written in nodejs: executable='/nix/store/p0f8i04zwf1dd66n2qkazk5x0fbsy7mp-nodejs-18.16.1/bin/node' version='v18.16.1'

Backend        
Name           x1
URL            gs://...
User           claudio
Organizations  

### Additional context

_No response_

### Contributing

EronWright:
### Proposed changes

_Kudos to @cbley-da for identifying the root cause and to @pdf for suggesting a fix, [here](https://github.com/pulumi/pulumi-kubernetes/issues/2481#issuecomment-1628879174)!_

This PR fixes an intermittent problem that resembles:
```
 kubernetes:helm.sh/v3:Release (mongodb):
    error: 1 error occurred:
    	* Helm release "mongodb/mongodb-456643ba" failed to initialize completely. Use Helm CLI to investigate.: failed to become available within allocated timeout. Error: 
          Helm Release mongodb/mongodb-456643ba: release mongodb-456643ba failed, and has been rolled back due to atomic being set: 
          failed to create resource: the server could not find the requested resource
```

The problem was that a Go struct (containing a client-go rest config) was being shared and mutated by numerous independent routines.  The solution was to copy the struct.

### Related issues (optional)

Closes #2481

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 - Fix: compute version field in Check for content detection (https://github.com/pulumi/pulumi-kubernetes/pull/2672)
 
+- Fix: Fix: Helm Release fails with "the server could not find the requested resource" (https://github.com/pulumi/pulumi-kubernetes/pull/2677)
+
 ## 4.5.4 (November 8, 2023)
 - Fix: Helm Release: chart requires kubeVersion (https://github.com/pulumi/pulumi-kubernetes/pull/2653)
 
--- a/provider/pkg/provider/kubeconfig.go
+++ b/provider/pkg/provider/kubeconfig.go
@@ -19,17 +19,19 @@ type KubeConfig struct {
 
 // ToDiscoveryClient implemented interface method
 func (k *KubeConfig) ToDiscoveryClient() (discovery.CachedDiscoveryInterface, error) {
+	c := rest.CopyConfig(k.restConfig)
+
 	// The more groups you have, the more discovery requests you need to make.
 	// given 25 groups (our groups + a few custom resources) with one-ish version each, discovery needs to make 50 requests
 	// double it just so we don't end up here again for a while.  This config is only used for discovery.
-	k.restConfig.Burst = 100
+	c.Burst = 100
 
-	return clients.NewMemCacheClient(discovery.NewDiscoveryClientForConfigOrDie(k.restConfig)), nil
+	return clients.NewMemCacheClient(discovery.NewDiscoveryClientForConfigOrDie(c)), nil
 }
 
 // ToRESTConfig implemented interface method
 func (k *KubeConfig) ToRESTConfig() (*rest.Config, error) {
-	return k.restConfig, nil
+	return rest.CopyConfig(k.restConfig), nil
 }
 
 // ToRESTMapper implemented interface method


rquitales (approved):
LGTM. Thanks!

PR #2678, opened 2023-11-28
Prepare for v4.5.5 release

rquitales:


guineveresaenger (approved):

EronWright (changes requested):
I don't think the master is ready to be released as v4.5.5, it should be v4.6.0 and undergo more preparations.

EronWright (approved):
I understand now that the changelog is being updated 'after the fact', and that v4.5.5 will not include this PR.

Issue #2683, opened 2023-12-04
Invalid `urn` property in `Chart` resource


pawelprazak:
### What happened?

The schema contains illegal parameter name:
```
        "kubernetes:helm.sh/v3:Chart": {
            "description": ...
            "properties": {
                ...
                "urn": {
                    "type": "string",
                    "description": "The stable logical URN used to distinctly address a resource, both before and after deployments."
                }
            },
            ...
```

That results in error in:
```
pulumi --logtostderr package get-schema kubernetes
error: error binding resource kubernetes:helm.sh/v3:Chart: failed to bind properties for kubernetes:helm.sh/v3:Chart: property name "urn" is reserved
```

### Example

```
pulumi --logtostderr package get-schema kubernetes
error: error binding resource kubernetes:helm.sh/v3:Chart: failed to bind properties for kubernetes:helm.sh/v3:Chart: property name "urn" is reserved
```

### Output of `pulumi about`

```
pulumi about
CLI
Version      3.95.0
Go Version   go1.21.4
Go Compiler  gc

Host
OS       darwin
Version  14.1
Arch     arm64

Backend
Name           pulumi.com
URL            https://app.pulumi.com/pprazak
User           pprazak
Organizations  pprazak, besom
Token type     personal
```

### Additional context

This is the change that introduced the strict schema validation:
https://github.com/pulumi/pulumi/pull/14637

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

pawelprazak:
As a workaround one can downgrade pulumi to `3.94.2`, e.g.:
```
curl -L -O https://raw.githubusercontent.com/Homebrew/homebrew-core/69b97f26bc78cf68eb30eedd0ca874b6e1914b19/Formula/p/pulumi.rb
brew install pulumi.rb
rm pulumi.rb
```

mjeffryes:
Thank you for reporting this; we'll take a look right away.

t0yv0:
Still hitting this in pulumi/registry PR referencing v4.5.6. Has this update been released?

mjeffryes:
This is released on 4.6.x I believe; We did a patch release on v4.5 out of order to help with a bugfix for a customer who could not switch to 4.6.x

t0yv0:
[...]

PR #2685, opened 2023-12-04
Remove unused `urn` output for helm Chart resources

rquitales:
### Proposed changes

Pulumi v3.95 enforces strict validation that reserves `urn` and `id` for their own usage. The `urn` output field in the helm Chart resource is actually unused, and being shadowed by the underlying Pulumi `urn` anyway. Removing this field will have no user facing impacts.

I have confirmed that running `pulumi --logtostderr package get-schema kubernetes` passes using the locally built provider binary from this PR.

### Related issues (optional)

Fixes: #2683


mjeffryes (approved):
Thanks for jumping on this one!

PR #2689, opened 2023-12-05
Update guestbook frontend image used for testing

rquitales:
The `gcr.io/google-samples/gb-frontend:v4` image was deleted from the container registry. The new image was obtained from the GoogleCloudPlatform repository: https://github.com/GoogleCloudPlatform/kubernetes-engine-samples.

Ref: https://cloud.google.com/kubernetes-engine/docs/tutorials/guestbook

### DIFF
--- a/tests/sdk/dotnet/yaml-url/Program.cs
+++ b/tests/sdk/dotnet/yaml-url/Program.cs
@@ -36,7 +36,8 @@ class YamlStack : Stack
     {
         return new ConfigFile(name, new ConfigFileArgs
         {
-            File = "https://raw.githubusercontent.com/pulumi/pulumi-kubernetes/master/tests/sdk/nodejs/examples/yaml-guestbook/yaml/guestbook.yaml",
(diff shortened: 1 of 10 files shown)


tests/sdk/dotnet/yaml-url/Program.cs
```diff
@@ -36,7 +36,8 @@ private ConfigFile MakeConfigFile(string name, string namespaceName, Provider pr
     {
         return new ConfigFile(name, new ConfigFileArgs
         {
-            File = "https://raw.githubusercontent.com/pulumi/pulumi-kubernetes/master/tests/sdk/nodejs/examples/yaml-guestbook/yaml/guestbook.yaml",
+            // TODO: Switch back to master branch once the guestbook.yaml file is merged into mainline branch.
```
rquitales, line 39: CI for this PR presubmit will fail, since the checked in manifest in `master` is not updated yet. I'm not sure what the best way to handle this, but switching to the current commit's manifest then following up with another PR to update it to the latest in master will quickly unblock us for now.

EronWright (approved):

Issue #2628, opened 2023-10-24
Support k8s 1.29 on release


mnlumi:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

k8s 1.29 is in progress. The provider should support 1.29 as soon as its available.

Current tag is an alpha: https://github.com/kubernetes/kubernetes/releases/tag/v1.29.0-alpha.2

mikhailshilkov:
@mnlumi @rquitales do we know the date? Should we open a draft PR?

mnlumi:
I haven't seen a date yet, but based on history, I expect this to come in early December. They only have an alpha out currently, not RC yet.

rquitales:
Final release is December 5. Code freeze is October 31. and test freeze is November 14. It probably makes sense to hold off crafting a draft PR until at least code freeze occurs and when a release candidate is cut.

mnlumi:
https://github.com/kubernetes/kubernetes/releases/tag/v1.29.0-alpha.3 was tagged on 11/2. There are no new updates in https://groups.google.com/g/kubernetes-announce.

PR #2690, opened 2023-12-06
Upgrade Kubernetes schema and libraries to v1.29

rquitales:
### Proposed changes

Update Kubernetes schema and libraries to v1.29.0.

### Related issues (optional)

Fixes: #2628


provider/pkg/provider/provider.go
```diff
@@ -1430,7 +1442,7 @@ func (k *kubeProvider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (
 		}
 
 		// Validate the object according to the OpenAPI schema for its GVK.
-		err = openapi.ValidateAgainstSchema(resources, newInputs)
+		err = openapi.ValidateAgainstSchema(k.getResourcesLazy(), resources, newInputs)
```
EronWright, line 1445: Is this effectively the same as before? Would there be a perf consideration here?

rquitales: This still needs a rework so we don't unnecessarily fetch the OpenAPI spec again, since we already call `k.getResources()`. I've just done it like this for now just to generate a schema to work from. Thanks for the initial review and flagging!

rquitales: Our implementation of OpenAPIResourcesGetter now returns a list of resources that have already been pre-fetched earlier for additional validation outside of the upstream helper function.

rquitales:
The 3 breaking changes should be relatively safe for our users.

- 🟡 "kubernetes:core/v1:PersistentVolumeClaimSpec": properties: "resources" type changed from "#/types/kubernetes:core/v1:ResourceRequirements" to "#/types/kubernetes:core/v1:VolumeResourceRequirements"
- 🟡 "kubernetes:core/v1:PersistentVolumeClaimSpecPatch": properties: "resources" type changed from "#/types/kubernetes:core/v1:ResourceRequirementsPatch" to "#/types/kubernetes:core/v1:VolumeResourceRequirementsPatch"

^ These were updated in Updated in https://github.com/kubernetes/api/commit/8b14183ca44824850577019d995e6f9e811825a2. This seems like a safe change since the previous type is incorrect for volumes as it doesn't require the `Claims` field.

- 🟢 "kubernetes:batch/v1:PodFailurePolicyRule": required: "onPodConditions" property is no longer Required

^ This was marked as optional in this upstream PR: https://github.com/kubernetes/kubernetes/pull/120204, and will be back ported to older minor k8s releases.

rquitales:
Waiting on client libraries to be tagged and released. AFAICT the final release of v1.29.0 should be identical to v1.29.0-rc.2. Will leave the changelog entry check as a merge blocker until this PR is updated to use the final release version of the client libraries.

EronWright (approved):

provider/pkg/openapi/openapi.go
```diff
@@ -48,7 +48,7 @@ import (
 
 // ValidateAgainstSchema validates a document against the given schema.
 func ValidateAgainstSchema(
-	resources openapi.Resources, obj *unstructured.Unstructured,
+	resourcesGetter openapi.OpenAPIResourcesGetter, resources openapi.Resources, obj *unstructured.Unstructured,
```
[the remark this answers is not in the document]
rquitales: This is cleaner and less variable passing. Done.

PR #2691, opened 2023-12-06
Use correct EditDirs in preparation for pulumi/pulumi#14695

thomas11:
The next Pulumi release containing pulumi/pulumi#14695 will fail tests when an `EditDir` is not found (which is the reasonable behavior). Code searched found one such case here.

### DIFF
--- a/tests/sdk/python/get-old/step1/__main__.py
+++ b/tests/sdk/python/get-old/step1/__main__.py
@@ -22,7 +22,7 @@ service = Service.get("kube-api", "kubernetes")
 
 crd = CustomResourceDefinition(
     resource_name="foo",
-    metadata={"name": "gettests.python.test"},
+    metadata={"name": "gettestolds.python.test"},
     spec={
         "group": "python.test",
         "versions": [{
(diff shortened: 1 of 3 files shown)


rquitales (approved):
Thanks for the proactive fix!

thomas11:
> Thanks for the proactive fix!

@rquitales Turns out the fixed test actually fails now that step2 is actually run. Do you know how to fix it?

```
 ~  kubernetes:apiextensions.k8s.io/v1beta1:CustomResourceDefinition foo updating (0s) [diff: ~apiVersion]; error: update of resource gettests.python.test failed because the Kubernetes API server reported that the apiVersion for this resource does not exist. Verify that any required CRDs have been created: no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
```

rquitales:
> > Thanks for the proactive fix!
> 
> @rquitales Turns out the fixed test actually fails now that step2 is actually run. Do you know how to fix it?
> 
> ```
>  ~  kubernetes:apiextensions.k8s.io/v1beta1:CustomResourceDefinition foo updating (0s) [diff: ~apiVersion]; error: update of resource gettests.python.test failed because the Kubernetes API server reported that the apiVersion for this resource does not exist. Verify that any required CRDs have been created: no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
> ```

Ah, this is more likely a test flake actually. Another existing test creates/deletes a cluster scoped CRD also called`GetTest`. We could either update the tests to use namespace scoped CRDs or rename the CRDs to not conflict.

I'm happy to take over this PR and get our tests in the right shape!

PR #2695, opened 2023-12-07
Remove remnant yarn.lock and node_modules when generating nodejs_sdk

rquitales:
### Proposed changes

When re-running `make nodejs_sdk` multiple times locally, we do not update the node packages to be used for the SDK. This could cause drift with CI, since CI always runs without a `yarn.lock` file present.


Makefile
```diff
@@ -74,6 +74,7 @@ nodejs_sdk::
 	$(WORKING_DIR)/bin/$(CODEGEN) -version=${VERSION} nodejs $(SCHEMA_FILE) $(CURDIR)
 	cd ${PACKDIR}/nodejs/ && \
 		echo "module fake_nodejs_module // Exclude this directory from Go tools\n\ngo 1.17" > go.mod && \
+		rm -rf node_modules yarn.lock && \
```
rquitales, line 77: Perhaps we should only just remove the lockfile? This would reduce the number of times we fetch node packages.

blampe:
Do we have a similar problem with other languages like Python?

Issue #2708, opened 2023-12-12
Setting Env Variable `PULUMI_KUBERNETES_MANAGED_BY_LABEL` causes panic


zbuchheit:
### What happened?

When attempting to set the environment variable `PULUMI_KUBERNETES_MANAGED_BY_LABEL` fails with a panic `panic: interface conversion: interface {} is nil, not string`. 

### Example

[This repo](https://github.com/zbuchheit/pulumi-examples/blob/main/aws-eks-daemonPatch-stackReference/go-daemon-patch/main.go) was where I witnessed this behavior with 
```bash
PULUMI_KUBERNETES_MANAGED_BY_LABEL=pulumi-manager pulumi up
```

### Output of `pulumi about`

CLI          
Version      3.92.0
Go Version   go1.21.3
Go Compiler  gc

Plugins
NAME        VERSION
go          unknown
kubernetes  4.5.5

Host     
OS       darwin
Version  14.0
Arch     arm64

This project is written in go: executable='/opt/homebrew/bin/go' version='go version go1.21.0 darwin/arm64'

Current Stack: zbuchheit-pulumi-corp/go-daemon-patch/dev

TYPE                               URN
pulumi:pulumi:Stack                urn:pulumi:dev::go-daemon-patch::pulumi:pulumi:Stack::go-daemon-patch-dev
pulumi:providers:pulumi            urn:pulumi:dev::go-daemon-patch::pulumi:providers:pulumi::default
pulumi:pulumi:StackReference       urn:pulumi:dev::go-daemon-patch::pulumi:pulumi:StackReference::zbuchheit-pulumi-corp/aws-ts-eks/dev
pulumi:providers:kubernetes        urn:pulumi:dev::go-daemon-patch::pulumi:providers:kubernetes::eks-kubernetes
kubernetes:apps/v1:DaemonSetPatch  urn:pulumi:dev::go-daemon-patch::kubernetes:apps/v1:DaemonSetPatch::k8s-daemonset-patch

Found no pending operations associated with dev

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/zbuchheit-pulumi-corp
User           zbuchheit-pulumi-corp
Organizations  zbuchheit-pulumi-corp,
Token type     personal

Dependencies:
NAME                                        VERSION
github.com/pulumi/pulumi-kubernetes/sdk/v4  4.5.5
github.com/pulumi/pulumi/sdk/v3             3.92.0

### Additional context

[...]

PR #2711, opened 2023-12-13
Fix panic when PULUMI_KUBERNETES_MANAGED_BY_LABEL is used on resources without the managed by label

rquitales:
### Proposed changes

This pull request addresses a panic scenario that occurs when attempting to assert a nil object as a string. The issue arises specifically when the `app.kubernetes.io/managed-by` label is absent from the Kubernetes object. This is the case when the object is created using SSA mode. The fix ensures proper handling of this situation to prevent the panic.

In addition to the fix, a new unit test has been introduced to  confirm the resolution of this panic scenario.

### Related issues (optional)

Fixes: #2708

### DIFF
--- a/provider/pkg/metadata/labels.go
+++ b/provider/pkg/metadata/labels.go
@@ -106,12 +106,18 @@ func HasManagedByLabel(obj *unstructured.Unstructured) bool {
 	if isComputedValue(val) {
 		return true
 	}
-	// now we should check to see if the user has specified a label via EnvVar
-	// we should also check to see if that value is the same as what is in the metadata
+
+	// Check if the value is a string/exists.
+	valStr, ok := val.(string)
+	if !ok {
+		return false
+	}
+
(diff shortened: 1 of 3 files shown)


provider/pkg/metadata/labels.go
```diff
@@ -106,12 +106,18 @@ func HasManagedByLabel(obj *unstructured.Unstructured) bool {
 	if isComputedValue(val) {
 		return true
 	}
-	// now we should check to see if the user has specified a label via EnvVar
-	// we should also check to see if that value is the same as what is in the metadata
+
```
EronWright, line 109: An alternative approach to consider:
```go
labelVal := "pulumi"
if env, exists := os.LookupEnv("PULUMI_KUBERNETES_MANAGED_BY_LABEL"); ok {
  labelVal = env
}

str, ok := val.(string)
return ok && str == labelVal
```

EronWright (approved):

Issue #2681, opened 2023-12-01
When a Secret contains a `\n` at the end it constantly tries to replace it


rshade:
### What happened?

We have a couple of secrets that have new lines at the end of them. Pulumi updates the secret, then on the next up, tries to replace it. This didn't happen in the version 3 provider. 

### Example

```
import pulumi
from pulumi_kubernetes.core.v1 import Secret
from pulumi_kubernetes.meta.v1 import ObjectMetaArgs

var_secret: str = "test\n"


new_secret: Secret = Secret(
    resource_name="test-secret",
    metadata=ObjectMetaArgs(
        name="test-secret",
        namespace="kube-system"
    ),
    data={
        "pvt.key": var_secret,
    },
    type="Opaque"
)
```

### Output of `pulumi about`

```
CLI
Version      3.94.2
Go Version   go1.21.4
Go Compiler  gc

Plugins
NAME        VERSION
kubernetes  4.5.5
python      unknown

Host
OS       ubuntu
Version  20.04
Arch     x86_64

This project is written in python: executable='/usr/bin/python3' version='3.8.10'

Current Stack: richard-shade-pulumi/pulumi-kube-secret/dev

TYPE                         URN
pulumi:pulumi:Stack          urn:pulumi:dev::pulumi-kube-secret::pulumi:pulumi:Stack::pulumi-kube-secret-dev
pulumi:providers:kubernetes  urn:pulumi:dev::pulumi-kube-secret::pulumi:providers:kubernetes::default_4_5_5
kubernetes:core/v1:Secret    urn:pulumi:dev::pulumi-kube-secret::kubernetes:core/v1:Secret::test-secret




Dependencies:
NAME               VERSION
pip                23.3.1
pkg_resources      0.0.0
pulumi-kubernetes  4.5.5
setuptools         69.0.2
wheel              0.42.0

Pulumi locates its logs in /tmp by default
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
This is probably related to how we convert `stringData` to `data` in input diff, compared to what Kubernetes does.
Related to #2511.

rquitales:
[...]

PR #2715, opened 2023-12-13
Fix normalization of base64 encoded secrets.data values to strip whitespace

rquitales:
### Proposed changes

This pull request introduces a normalization mechanism for `secret.data` base64 encoded values, specifically stripping whitespace to prevent unintended spurious diffs.

**Changes Made:**
- Created comprehensive integration and unit tests to verify and highlight the broken behavior associated with `secret.data` values containing whitespaces.
- Updated the provider to incorporate the normalization process, ensuring that all related tests pass seamlessly.

### Related issues (optional)

Fixes: #2681

### DIFF
--- a/provider/pkg/clients/unstructured.go
+++ b/provider/pkg/clients/unstructured.go
@@ -17,6 +17,7 @@ package clients
 import (
 	"encoding/base64"
 	"fmt"
+	"strings"
 
 	"github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/kinds"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
@@ -115,10 +116,15 @@ func normalizeSecret(uns *unstructured.Unstructured) *unstructured.Unstructured 
 	contract.Assertf(IsSecret(uns), "normalizeSecret called on a non-Secret resource: %s:%s", uns.GetAPIVersion(), uns.GetKind())
 
 	stringData, found, err := unstructured.NestedStringMap(uns.Object, "stringData")
-	if err != nil || !found {
-		return uns
+	if err != nil || !found || len(stringData) == 0 {
+		// Normalize the .data field if .stringData is not present or empty.
+		return normalizeSecretData(uns)
 	}
 
+	return normalizeSecretStringData(stringData, uns)
+}
+
+func normalizeSecretStringData(stringData map[string]string, uns *unstructured.Unstructured) *unstructured.Unstructured {
 	data, found, err := unstructured.NestedMap(uns.Object, "data")
 	if err != nil || !found {
 		data = map[string]any{}
@@ -134,6 +140,25 @@ func normalizeSecret(uns *unstructured.Unstructured) *unstructured.Unstructured 
 		contract.IgnoreError(unstructured.SetNestedMap(uns.Object, data, "data"))
 		unstructured.RemoveNestedField(uns.Object, "stringData")
 	}
+
+	return uns
+}
+
+// normalizeSecretData normalizes the .data field of a Secret resource by trimming whitespace from string values.
+// This is necessary because the apiserver will trim whitespace from the .data field values, but the provider does not.
+func normalizeSecretData(uns *unstructured.Unstructured) *unstructured.Unstructured {
+	data, found, err := unstructured.NestedMap(uns.Object, "data")
+	if err != nil || !found || len(data) == 0 {
+		return uns
+	}
+
+	for k, v := range data {
+		if s, ok := v.(string); ok {
+			data[k] = strings.TrimSpace(s)
+		}
+	}
+
+	contract.IgnoreError(unstructured.SetNestedMap(uns.Object, data, "data"))
 
 	return uns
 }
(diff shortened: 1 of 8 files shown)


EronWright (commented):
I don't feel comfortable with this fix for a few reasons.
1. I don't understand why one would get spurious diffs if the whitespace were present.
2. I feel it is presumptuous to manipulate the value; could there be a legit use case for trailing whitespace?

rquitales:
> I don't feel comfortable with this fix for a few reasons.
> 
> 1. I don't understand why one would get spurious diffs if the whitespace were present.
> 2. I feel it is presumptuous to manipulate the value; could there be a legit use case for trailing whitespace?

> I have reservations about endorsing this solution for a couple of reasons.
> 
> 1. I'm uncertain about the necessity of eliminating trailing whitespace, as it seems counterintuitive to encounter spurious differences in the absence of whitespace.
> 2. I'm cautious about modifying the value, considering there might be legitimate use cases for trailing whitespace.

1. The occurrence of spurious diffs is attributed to Kubernetes stripping whitespace during the object creation process. Consequently, when comparing the object in its state to our input, which includes whitespace, a diff arises. This can be illustrated with the following YAML manifest:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: dotfile-secret
data:
  secret: "dmFsdWUtMg0KDQo=\n\n\n"
```
When this secret is applied to the cluster using `kubectl apply -f secret.yaml`, the output of `kubectl get secret -f secret.yaml` shows the absence of trailing whitespace in the stored value. Notice the difference between `.data.secret` and the one in `kubectl.kubernetes.io/last-applied-configuration`

```yaml
apiVersion: v1
data:
  secret: dmFsdWUtMg0KDQo=
kind: Secret
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","data":{"secret":"dmFsdWUtMg0KDQo=\n"},"kind":"Secret","metadata":{"annotations":{},"name":"dotfile-secret","namespace":"default"}}
  creationTimestamp: "2023-12-13T21:55:13Z"
  name: dotfile-secret
  namespace: default
  resourceVersion: "54021"
  uid: 1d680c61-327f-4bdd-bc81-6e922fe525e4
type: Opaque
```

2. Since Kubernetes removes trailing whitespace, it seems improbable for it to store whitespace. Furthermore, attempting to store a secret with multiple base64 strings separated by new lines results in rejection by the API server, underscoring the implementation's intolerance for trailing whitespace.

---
Note: A secret with multiple base64 strings separated by a new line would be rejected by the API server, e.g., `secret: "dmFsdWUtMg0KDQo=\n\n\ndmFsdWUtMg0KDQo="` would return the following error: `The request is invalid: patch: Invalid value: "map[data:map[secret:dmFsdWUtMg0KDQo=\n\n\ndmFsdWUtMg0KDQo=] metadata:map[annotations:map[kubectl.kubernetes.io/last-applied-configuration:{\"apiVersion\":\"v1\",\"data\":{\"secret\":\"dmFsdWUtMg0KDQo=\\n\\n\\ndmFsdWUtMg0KDQo=\"},\"kind\":\"Secret\",\"metadata\":{\"annotations\":{},\"name\":\"dotfile-secret\",\"namespace\":\"default\"}}\n]]]": error decoding from json: illegal base64 data at input byte 19`

Trailing whitespace is stripped by Kubernetes due to the underlying implementation of the `encoding/base64` package that the serializer codec uses: [Kubernetes Source Code](https://github.com/kubernetes/kubernetes/blob/41890534532931742770a7dc98f78bcdc59b1a6f/staging/src/k8s.io/apimachinery/pkg/runtime/codec.go#L212-L260).

Refer to this simple Go Playground example ([Go Playground](https://go.dev/play/p/Z6xNA7mBYS8)) that demonstrates Golang's `encoding/base64` stripping trailing whitespace after decoding and re-encoding which is what is done in the Kubernetes codebase.

provider/pkg/clients/unstructured.go
```diff
@@ -138,6 +143,37 @@ func normalizeSecret(uns *unstructured.Unstructured) *unstructured.Unstructured
...
+			// does the decoding and re-encoding to validate the value provided is valid base64.
+			// See: https://github.com/kubernetes/kubernetes/blob/41890534532931742770a7dc98f78bcdc59b1a6f/staging/src/k8s.io/apimachinery/pkg/runtime/codec.go#L212-L260
+			base64Decoded, err := base64.StdEncoding.DecodeString(s)
+			if err != nil {
+				// TODO: propagate error upwards to parent Normalize function to fail early. It is safe to
+				// ignore this error for now, since the apiserver will reject the resource if the value cannot
+				// be decoded.
+				continue
+			}
+			base64ReEncoded := base64.StdEncoding.EncodeToString(base64Decoded)
```
rquitales, line 166: Now decoding and re-encoding to base64 to be closer to what Kubernetes does instead of simply trimming trailing whitespace.

EronWright (approved):

provider/pkg/clients/unstructured.go
EronWright: Thanks for the explanation, it makes sense now.

Issue #2718, opened 2023-12-14
`DaemonSetPatch` taking ownership of fields unexpectedly upon an update operation with a parent set.


zbuchheit:
### What happened?

When attempting to update an DaemonSetPatch resource while a parent component is set, the DaemonSetPatch took ownership of all field previously owned by `kubectl-client-side-apply`. This leads to a broken state that requires manual intervention to repair.

## Notes
If you create the resource, then change a field like the `image` to trigger an update, the behavior will occur.
## Code Example
```ts
export class MyComponent extends pulumi.ComponentResource {
    daemonsetPatch: DaemonSetPatch
    constructor(name: string, opts?: pulumi.ComponentResourceOptions) {
        super("my:component:Resource", name, opts);
        this.daemonsetPatch = new kubernetes.apps.v1.DaemonSetPatch("k8s-daemonset-patch-child", {
            metadata: {
                name: "kube-proxy",
                namespace: "kube-system",
                annotations: {
                    "pulumi.com/patchForce": "true",
                    "pulumi.com/patchFieldManager": "pulumi-daemonset-patch-1",
                },
            },
            spec: {
                template: {
                    spec: {
                        containers: [{
                            name: "kube-proxy",
                            image: "602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/kube-proxy:v1.24.7-minimal-eksbuild.2",
                            command: [
                                "kube-proxy",
                                "--hostname-override=$(NODE_NAME)",
                                "--v=2",
                                "--config=/var/lib/kube-proxy/config.conf",
                            ],
                            env: [{
                                name: "NODE_NAME",
                                valueFrom: {
                                    fieldRef: {
                                        fieldPath: "spec.nodeName",
                                    },
                                },
                            }],
                        }],
                    },
                },
            },
        }, { provider: kubeProvider, retainOnDelete: true, parent: this })
        // Export the Namespace name
        this.registerOutputs({
        });
    }
}

const daemonSet = new MyComponent("k8s-daemonset-patch");
```

### Output of `pulumi about`
```
CLI          
Version      3.92.0
Go Version   go1.21.3
Go Compiler  gc

Plugins
NAME        VERSION
aws         6.13.3
awsx        2.3.0
docker      4.5.0
docker      3.6.1
eks         2.0.0
kubernetes  4.6.0
nodejs      unknown

Host     
OS       darwin
Version  14.0
Arch     arm64

[...]

PR #2719, opened 2023-12-14
Fix resource kind lookup using URN.Type instead of QualifiedType

rquitales:
### Proposed changes

This pull request uses the URN `Type` method, rather than `QualifiedType` to search if resources are a patch or list resource.  Qualified types include the parent resource if there is one, so it would have failed the lookup.

Additionally, comprehensive unit tests have been included to validate this change.

### Related issues (optional)
Fixes: #2718 (manual validation was done to ensure this fix solves this CUJ)

### DIFF
--- a/provider/pkg/kinds/util.go
+++ b/provider/pkg/kinds/util.go
@@ -1,0 +1,45 @@
+package kinds
+
+import (
+	"strings"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+func suffixSearch(urnS string, suffixes []string) bool {
+	for _, suffix := range suffixes {
+		if strings.HasSuffix(urnS, suffix) {
+			return true
+		}
+	}
+	return false
+}
+
+// IsPatchURN returns true if the URN is for a Patch resource.
+func IsPatchURN(urn resource.URN) bool {
+	urnS := urn.QualifiedType().String()
+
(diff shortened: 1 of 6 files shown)


provider/pkg/kinds/util.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// Do a simple O(1) lookup in a set of known patch types.
+	if PatchQualifiedTypes.Has(urnS) {
+		return true
+	}
+
+	// Do a suffix search in case the resource is a component resource containing
+	// a patch resource. Eg. a component resource patch resource could have the following URN:
+	// my:component:Resource$kubernetes:apps/v1:DaemonSetPatch.
+	return suffixSearch(urnS, PatchQualifiedTypes.SortedValues())
```
blampe, line 30: Would `urn.Type()` work here instead?

IIRC `QualifiedType` gives you `parent$child` but `Type` gives you `child`, and it seems like we only want to look up `PatchQualifiedTypes.Has(child)`.

rquitales: Oh nice! Yes, that works too. Wasn't familiar with all the methods of Resource.URN, but yes, `Type` just gives the resource type, whereas `QualifiedType` also returns the parent if there is one. Updated.

blampe (approved):

PR #2759, opened 2024-01-18
WIP: Use status subresource for deployment rollout awaiting

rquitales:
### Proposed changes

Change await to use status subresource for deployments. This WIP PR watches for deployment events such that whenever the status is updated, we log the rollout status. Tests are expected to fail as they require a full re-write. The code structure could also be properly refactored, since this is currently still just using the existing structure which doesn't make too much sense with the level of nesting there currently is.

### DIFF
--- a/provider/pkg/await/deployment.go
+++ b/provider/pkg/await/deployment.go
@@ -16,26 +16,24 @@ package await
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


rquitales:
Closing this PR as the current await logic has been relatively stable for ~3 years and is more status rich. Changing to use the status subresource appears to be something scheduled for a major provider release instead.

Issue #2663, opened 2023-11-10
DiffConfig fails when provider's kubeconfig is set to file path


mikhailshilkov:
### What happened?

Reported in https://github.com/pulumi/pulumi/issues/14529

A CLI version that doesn't ignore errors returned from DiffConfig fails with

> error: rpc error: code = Unknown desc = failed to parse kubeconfig: couldn't get version/kind; json parse error: json: cannot unmarshal string into Go value of type struct { APIVersion string "json:\"apiVersion,omitempty\""; Kind string "json:\"kind,omitempty\"" }

when an explicit kubernetes provider points to a kubeconfig via its path.

I reproduced that the error is returned from this line: https://github.com/pulumi/pulumi-kubernetes/blob/1944a5215b0e12b373547393a6294ef0aee60c28/provider/pkg/provider/provider.go#L376

### Example

```
import { Provider } from '@pulumi/kubernetes'

export const provider = new Provider('***', {
  kubeconfig: join(__dirname, 'kubeconfig.yaml')
})
```

### Output of `pulumi about`

```
CLI          
Version      3.92.0
Go Version   go1.21.3
Go Compiler  gc

Plugins
NAME            VERSION
aws             5.4.0
aws             4.38.1
aws             4.38.1
aws-apigateway  0.0.6
aws-native      0.16.1
command         0.1.0
docker          3.1.0
eks             0.36.0
kubernetes      3.12.1
nodejs          unknown
random          4.3.1

Host     
OS       darwin
Version  14.1
Arch     arm64

This project is written in nodejs: executable='/usr/local/bin/node' version='v18.17.0'

Dependencies:
NAME                              VERSION
@types/node                       17.0.5
husky                             7.0.4
deepmerge                         4.2.2
lodash                            4.17.21
@aws-sdk/client-sts               3.45.0
@pulumi/aws-apigateway            0.0.6
@pulumi/aws-native                0.16.1
@pulumi/eks                       0.36.0
@pulumi/kubernetes                3.12.1
@pulumi/pulumi                    3.18.0
eslint                            8.12.0
typescript                        4.6.3
@aws-sdk/client-kms               3.45.0
@pulumi/aws                       5.4.0
@pulumi/awsx                      0.40.0
@pulumi/command                   0.1.0
@pulumi/random                    4.3.1
@typescript-eslint/eslint-plugin  5.18.0
@types/js-yaml                    4.0.5
@types/lodash                     4.14.178
[...]

Issue #1032, opened 2020-03-15
Panic on non-string `kubeconfig`


lukehoban:
A user reported the panic below.  This appears to be due to the `kubeconfig` property being set to something other than a string, and a call to `.StringValue()` panicing.  

In general, there are a lot of calls to `.StringValue()` and similar in this package that are not guarded with `.IsString()` checks and are thus likely to panic instead of cleanly error on malformed inputs.  We likely should audit (and test) these more holistically.

```
    panic: interface conversion: interface {} is resource.PropertyMap, not string
    goroutine 28 [running]:
    github.com/pulumi/pulumi/pkg/resource.PropertyValue.StringValue(...)
    	/home/travis/gopath/pkg/mod/github.com/pulumi/pulumi@v1.6.1/pkg/resource/properties.go:359
    github.com/pulumi/pulumi-kubernetes/pkg/provider.parseKubeconfigPropertyValue(0x2386280, 0xc0001fea50, 0x2475423, 0xa, 0xc0001b2508)
    	/home/travis/gopath/src/github.com/pulumi/pulumi-kubernetes/pkg/provider/util.go:85 +0x169
    github.com/pulumi/pulumi-kubernetes/pkg/provider.(*kubeProvider).DiffConfig(0xc000014000, 0x26e1260, 0xc0001fe9f0, 0xc0001380e0, 0xc000014000, 0x2275301, 0xc00031a0c0)
    	/home/travis/gopath/src/github.com/pulumi/pulumi-kubernetes/pkg/provider/provider.go:278 +0x61b
    github.com/pulumi/pulumi/sdk/proto/go._ResourceProvider_DiffConfig_Handler.func1(0x26e1260, 0xc0001fe9f0, 0x23a5e80, 0xc0001380e0, 0x23c0a00, 0x33070c0, 0x26e1260, 0xc0001fe9f0)
    	/home/travis/gopath/pkg/mod/github.com/pulumi/pulumi@v1.6.1/sdk/proto/go/provider.pb.go:1504 +0x86
    github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc.OpenTracingServerInterceptor.func1(0x26e1260, 0xc000577200, 0x23a5e80, 0xc0001380e0, 0xc00000cb20, 0xc00000cb40, 0x0, 0x0, 0x26a05e0, 0xc0000cf7b0)
[...]

PR #2771, opened 2024-01-26
Fix: DiffConfig fails when provider's kubeconfig is set to file path

EronWright:
### Proposed changes

This PR refactors the kubeconfig loading code to be more consistent between `DiffConfig` and `Configure`.  Previously the `DiffConfig` method did not support file-based kubeconfigs, and would throw a spurious parser error.

`DiffConfig` loads the kubeconfig to be able to detect cluster replacement, by comparing the active cluster info.  Unlike `Configure`, it doesn't have support for ambient kubeconfigs.  When an ambient config is in use, or the config is invalid (e.g. the `context` or `cluster` properties refer to a non-existent entry), then the replacement logic is skipped.

This PR also fixes a panic in `DiffConfig` that would occur if `cluster` or `context` is a computed value; the provider will conservatively suggest replacement as it does when `kubeconfig` is a computed value.

### Testing
New test cases were added for cluster change detection.

New tests were added for the parsing code that was consolidated into `utils.go`. The tests now cover file-, string-, and object-based resource property values for `kubeconfig`.

### Related issues (optional)

Closes: #2663 
Related: #1032

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -22,10 +22,10 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"os/user"
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -374,13 +374,15 @@ func (k *kubeProvider) DiffConfig(ctx context.Context, req *pulumirpc.DiffReques
 	}
 
 	// We can't tell for sure if a computed value has changed, so we make the conservative choice
-	// and force a replacement.
-	if news["kubeconfig"].IsComputed() {
-		return &pulumirpc.DiffResponse{
-			Changes:  pulumirpc.DiffResponse_DIFF_SOME,
-			Diffs:    []string{"kubeconfig"},
-			Replaces: []string{"kubeconfig"},
-		}, nil
+	// and force a replacement. Note that getActiveClusterFromConfig relies on all three of the below properties.
+	for _, key := range []resource.PropertyKey{"kubeconfig", "context", "cluster"} {
+		if news[key].IsComputed() {
+			return &pulumirpc.DiffResponse{
+				Changes:  pulumirpc.DiffResponse_DIFF_SOME,
+				Diffs:    []string{string(key)},
+				Replaces: []string{string(key)},
+			}, nil
+		}
 	}
 
 	var diffs, replaces []string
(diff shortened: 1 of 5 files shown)


provider/pkg/provider/provider.go
```diff
@@ -419,10 +421,17 @@ func (k *kubeProvider) DiffConfig(ctx context.Context, req *pulumirpc.DiffReques
...
+	activeCluster, found := getActiveClusterFromConfig(newConfig, news)
+	if !oldFound || !found {
+		// The config is either ambient or invalid, so we can't draw any conclusions.
+	} else if !reflect.DeepEqual(oldActiveCluster, activeCluster) {
+		// one of these properties must have changed for the active cluster to change.
+		for _, key := range []string{"kubeconfig", "context", "cluster"} {
+			if slices.Contains(diffs, key) {
+				replaces = append(replaces, key)
+			}
+		}
```
EronWright, line 434: This change makes the replacement logic be slightly less conservative, because it now won't suggest a replacement when the old or new cluster info couldn't be resolved, e.g. when the config is ambient or the `context` doesn't match anything. Given that this PR fixes a problem in `parseKubeconfigPropertyValue`, we're more likely to reach this logic than in the past.

provider/pkg/provider/provider.go
```diff
@@ -374,13 +374,15 @@ func (k *kubeProvider) DiffConfig(ctx context.Context, req *pulumirpc.DiffReques
...
-		}, nil
+	// and force a replacement. Note that getActiveClusterFromConfig relies on all three of the below properties.
+	for _, key := range []resource.PropertyKey{"kubeconfig", "context", "cluster"} {
+		if news[key].IsComputed() {
+			return &pulumirpc.DiffResponse{
+				Changes:  pulumirpc.DiffResponse_DIFF_SOME,
+				Diffs:    []string{string(key)},
+				Replaces: []string{string(key)},
+			}, nil
+		}
```
EronWright, line 385: This fix is to avoid a panic that otherwise occurs in `getActiveClusterFromConfig` if either `context` or `cluster` has a computed value. I think it is safe to be 'more conservative' during preview since it is not determinative as to whether a resource would actually be replaced.

provider/pkg/provider/provider.go
```diff
@@ -660,39 +669,9 @@ func (k *kubeProvider) Configure(_ context.Context, req *pulumirpc.ConfigureRequ
...
-			} else {
-				contents = string(b)
-			}
-		} else { // Assume the contents are a k8s config.
-			contents = pathOrContents
-		}
-
-		// Load the contents of the k8s config.
-		apiConfig, err = clientcmd.Load([]byte(contents))
+		apiConfig, err := parseKubeconfigString(pathOrContents)
```
EronWright, line 674: This change is the essential one that teaches `DiffConfig` to support a file-based `kubeconfig` property, by reusing the logic of `Configure`.

provider/pkg/provider/util.go
```diff
@@ -87,31 +124,31 @@ func parseKubeconfigPropertyValue(kubeconfig resource.PropertyValue) (*clientapi
...
 		if err != nil {
 			return nil, fmt.Errorf("failed to unmarshal kubeconfig: %v", err)
 		}
-		cfg = jsonBytes
+		cfg = string(jsonBytes)
 	} else {
 		return nil, fmt.Errorf("unexpected kubeconfig format, type: %v", kubeconfig.TypeString())
 	}
-	config, err := clientcmd.Load(cfg)
+	config, err := parseKubeconfigString(cfg)
```
EronWright, line 140: This change is the essential one that teaches `DiffConfig` to support a file-based `kubeconfig` property, by reusing the logic of `Configure`.

lblackstone (approved):

Issue #2679, opened 2023-11-29
Helm Release skips preview logic during Update


EronWright:
### Problem description

The kube provider erroneously fails to handoff to the helm provider, in the case of a preview of an update to a Helm Release. The code in `kubeProvider.Update` waits too long to call `isHelmRelease` and complete the handoff; it first converts the inputs to an `Unstructured` and also calls `gvkExists` (which always returns false in this situation), neither of which make sense for Helm Release.

https://github.com/pulumi/pulumi-kubernetes/blob/b045235e4f8a6c239edcf44fb1150b36c36d6c48/provider/pkg/provider/provider.go#L2254-L2263

The code in `Create` has a similar problem:
https://github.com/pulumi/pulumi-kubernetes/blob/b045235e4f8a6c239edcf44fb1150b36c36d6c48/provider/pkg/provider/provider.go#L1753-L1755

The code in `Diff` similarly skips the handoff in some cases:

https://github.com/pulumi/pulumi-kubernetes/blob/b045235e4f8a6c239edcf44fb1150b36c36d6c48/provider/pkg/provider/provider.go#L1577-L1579

To be clear, if the handoff is skipped, the wrong logic is applied, e.g. conversion to `Unstructured`, k8s object defaults applied, etc.

Related:
- https://github.com/pulumi/pulumi-kubernetes/pull/1760
- https://github.com/pulumi/pulumi-kubernetes/pull/1953

PR #2808, opened 2024-02-01
Support for metadata.generateName (CSA)

### LINKED ISSUE #2539: generate_name prefix seems to not be honored

EronWright:
### Proposed changes

This PR implements support for `.metadata.generateName` in CSA mode, based on https://github.com/pulumi/pulumi-kubernetes/pull/2790.
1. Adjust the auto-naming logic to consider `.metadata.generateName` to be a variant of auto-naming. Pulumi will not assign an auto-name to a new resource that has `generateName`, and upon delete will use the replace-then-delete technique.
2. Use outputs rather than inputs to fetch the live object or to obtain the live object's name.
3. Fix an autonaming bug where the old auto-name would erroneously be adopted, overwriting a new computed name. 

### Tests
1. A new integration test is provided to test the use of `metadata.generateName`.  It tests creation, update, replacement, and promotion from `.generateName` to `.name`.  ([ref](https://github.com/pulumi/pulumi-kubernetes/pull/2808/files#diff-7e85ce707283b1d281d971a8e5b0c49b959b0803ba3437dc9dbd422552326835R269))
2. New test cases for the await logic.
3. The existing autonaming test is enhanced to test how autonaming takes precedence over `generateName`, in the update case. This is to ensure backwards compatibility, e.g. in the edge case that an existing object has a `generateName` field (which Pulumi ignored until now).

### Related issues (optional)
Closes #2539

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -1720,12 +1720,14 @@ func (k *kubeProvider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*p
 	deleteBeforeReplace :=
 		// 1. We know resource must be replaced.
 		len(replaces) > 0 &&
-			// 2. Object is NOT autonamed (i.e., user manually named it, and therefore we can't
-			// auto-generate the name on client or server).
-			!(metadata.IsAutonamed(newInputs) || metadata.IsGenerateName(newInputs, newResInputs)) &&
-			// 3. The new, user-specified name is the same as the old name.
+			// 2. Object is named (i.e., not using metadata.generateName).
+			metadata.IsNamed(newInputs, newResInputs) &&
+			// 3. Object is NOT autonamed (i.e., user manually named it, and therefore we can't
+			// auto-generate the name).
+			!metadata.IsAutonamed(newInputs) &&
+			// 4. The new, user-specified name is the same as the old name.
 			newInputs.GetName() == oldLive.GetName() &&
-			// 4. The resource is being deployed to the same namespace (i.e., we aren't creating the
+			// 5. The resource is being deployed to the same namespace (i.e., we aren't creating the
 			// object in a new namespace and then deleting the old one).
 			newInputs.GetNamespace() == oldLive.GetNamespace()
 
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -1294,7 +1294,7 @@ func (k *kubeProvider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (
 	}
 
 	if !k.serverSideApplyMode && kinds.IsPatchURN(urn) {
-		return nil, fmt.Errorf("patch resources require Server-side Apply mode, which is enabled using the " +
+		return nil, fmt.Errorf("patch resources require Server-Side Apply mode, which is enabled using the " +
 			"`enableServerSideApply` Provider config")
 	}
 
@@ -1334,7 +1334,7 @@ func (k *kubeProvider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (
 
 	if k.serverSideApplyMode && kinds.IsPatchURN(urn) {
 		if len(newInputs.GetName()) == 0 {
-			return nil, fmt.Errorf("patch resources require the resource `.metadata.name` to be set")
+			return nil, fmt.Errorf("patch resources require the `.metadata.name` field to be set")
 		}
 	}
 
@@ -1349,10 +1349,9 @@ func (k *kubeProvider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (
 	// needs to be `DeleteBeforeReplace`'d. If the resource is marked `DeleteBeforeReplace`, then
 	// `Create` will allocate it a new name later.
 	if len(oldInputs.Object) > 0 {
-		// NOTE: If old inputs exist, they have a name, either provided by the user or filled in with a
-		// previous run of `Check`.
-		contract.Assertf(oldInputs.GetName() != "", "expected object name to be nonempty: %v", oldInputs)
-		metadata.AdoptOldAutonameIfUnnamed(newInputs, oldInputs)
+		// NOTE: If old inputs exist, they MAY have a name, either provided by the user, or based on generateName,
+		// or filled in with a previous run of `Check`.
+		metadata.AdoptOldAutonameIfUnnamed(newInputs, oldInputs, news)
 
 		// If the resource has existing state, we only set the "managed-by: pulumi" label if it is already present. This
 		// avoids causing diffs for cases where the resource is being imported, or was created using SSA. The goal in
@@ -1375,6 +1374,14 @@ func (k *kubeProvider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (
 				return nil, pkgerrors.Wrapf(err,
 					"Failed to create object because of a problem setting managed-by labels")
 			}
+		}
+	}
+	if metadata.IsGenerateName(newInputs, news) {
+		if k.serverSideApplyMode {
+			return nil, fmt.Errorf("the `.metadata.generateName` field is not supported in Server-Side Apply mode")
+		}
+		if k.yamlRenderMode {
+			return nil, fmt.Errorf("the `.metadata.generateName` field is not supported in YAML rendering mode")
 		}
 	}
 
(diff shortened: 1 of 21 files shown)


provider/pkg/provider/provider.go
```diff
@@ -1552,6 +1559,9 @@ func (k *kubeProvider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*p
 	newInputs := propMapToUnstructured(newResInputs)
 
 	oldInputs, oldLive := parseCheckpointObject(oldState)
+	if !isHelmRelease(urn) {
+		contract.Assertf(oldLive.GetName() != "", "expected live object name to be nonempty: %v", oldLive)
+	}
```
EronWright, line 1564: This is a workaround for https://github.com/pulumi/pulumi-kubernetes/issues/2679 and I'll add a comment before merging.

provider/pkg/provider/provider.go
```diff
@@ -1600,6 +1611,12 @@ func (k *kubeProvider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*p
 	if k.serverSideApplyMode && len(oldLivePruned.GetResourceVersion()) > 0 {
 		oldLivePruned.SetResourceVersion("")
 	}
+	// If a name was specified in the new inputs, be sure that the old live object has the previous name.
+	// This makes it possible to update the program to set `.metadata.name` to the name that was
+	// made by `.metadata.generateName` without triggering replacement.
+	if newInputs.GetName() != "" {
+		oldLivePruned.SetName(oldLive.GetName())
```
rquitales, line 1618: Admittedly, this took me a couple of passes to understand what's happening. Maybe it's worthwhile to expand the comment a bit? Something like, generateName will create and store a name, but the oldLivePruned object drops the name field to match the input, which we now need for diffing.

Though this could just be an artifact of code review only showing small chunks of code without context.

tests/sdk/nodejs/autonaming/step4/index.ts
```diff
@@ -14,17 +14,17 @@
 
 import * as k8s from "@pulumi/kubernetes";
 
-export const namespace = new k8s.core.v1.Namespace("test-namespace");
+const namespace = new k8s.core.v1.Namespace("test-namespace");
 
 //
-// User has now specified `.metadata.name`, so Pulumi should replace the resource, and NOT allocate
-// a name to it.
+// User has now specified `.metadata.generateName`, which Pulumi ignores because autonaming has already occurred,
```
rquitales, line 20: In this scenario, could we emit a warning to the user that we won't honor `generateName`? I believe we just silently continue right?

tests/sdk/nodejs/generatename/step2/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import * as pulumi from "@pulumi/pulumi";
+import * as k8s from "@pulumi/kubernetes";
+
+const config = new pulumi.Config();
+
+const namespace = new k8s.core.v1.Namespace("test-namespace");
+
+//
+// The `.metadata.generateName` field has changed, but Pulumi does NOT automatically replace in that situation.
```
rquitales, line 23: Maybe a warning here too?

rquitales (approved):
LGTM

EronWright:
@vavsab I would guess that your provider is configured to use server-side apply mode, which you can disable:
https://www.pulumi.com/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply/

gunzy83:
> @vavsab I would guess that your provider is configured to use server-side apply mode, which you can disable: https://www.pulumi.com/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply/

Is there a reason this is not supported with SSA? Will it be supported in the future?

EronWright:
The reason that it isn't supported is that, in SSA mode, the provider uses Apply (HTTP PATCH) to both create and update the object (as described [here](https://kubernetes.io/docs/reference/using-api/api-concepts/#patch-and-apply)), and one cannot use `.metadata.generateName` in that case. See https://github.com/pulumi/pulumi-kubernetes/issues/2926 for a discussion about the behavior should be changed to use Create (HTTP POST) even in SSA mode, which would have the nice side-effect of enabling this feature.

gunzy83:
> The reason that it isn't supported is that, in SSA mode, the provider uses Apply (HTTP PATCH) to both create and update the object (as described [here](https://kubernetes.io/docs/reference/using-api/api-concepts/#patch-and-apply)), and one cannot use `.metadata.generateName` in that case. See #2926 for a discussion on whether the behavior should be changed to use Create (HTTP POST) even in SSA mode, which would have the nice side-effect of enabling this feature.

Thanks, I understand now. I will follow the issue over there.

PR #2831, opened 2024-02-16
Use KinD clusters in integration tests

rquitales:
### Related issues (optional)

Fixes: #2243


EronWright:
Some questions:

1. Why is it better to have the test code create the cluster instead of a CI step?
2. How/when will we support the new provider test framework and how many tests are being skipped?
3.

tests/sdk/nodejs/nodejs_test.go
```diff
@@ -449,10 +452,24 @@ func TestCRDs(t *testing.T) {
 			},
 		},
 	})
+	test, _ = testClusters.WrapProviderTestOptions(test)
 	integration.ProgramTest(t, &test)
 }
 
 func TestPod(t *testing.T) {
+	getCondition := func(conditions []any, conditionType string) map[string]any {
```
EronWright, line 460: I suppose that `FindStatusCondition` wasn't useful here?
https://github.com/kubernetes/apimachinery/blob/v0.29.2/pkg/api/meta/conditions.go#L91

tests/sdk/nodejs/nodejs_test.go
```diff
@@ -1511,22 +1543,23 @@ func TestServerSideApplyEmptyMaps(t *testing.T) {
...
 	// Use manual lifecycle management since we need to run external commands in between pulumi up steps, while referencing
 	// the same stack.
+	applyStep, kcfg := testClusters.WrapProviderTestOptions(applyStep)
 	pt := integration.ProgramTestManualLifeCycle(t, &applyStep)
 	err := pt.TestLifeCycleInitAndDestroy()
 	assert.NoError(t, err)
 
 	// Sanity check with kubectl to verify that the ConfigMap was created with the wanted label.
-	out, err := exec.Command("kubectl", "get", "configmap", "-o", "yaml", "-n", ns, cmName).CombinedOutput()
+	out, err := exec.Command("kubectl", "get", "configmap", "--kubeconfig", kcfg, "-o", "yaml", "-n", ns, cmName).CombinedOutput()
```
EronWright, line 1552: Maybe this could be avoided with the `KUBECONFIG` variable.

EronWright (commented):
Nit: Seems like we are exploiting the `--short` flag, not to "tell long-running tests to shorten their run time" but to filter incompatible tests.  
Ideally we'd update the tests to be compatible, e.g. by using a clusterIP.  

My main feedback is that I would prefer we keep cluster creation outside of the go test code, e.g. by using a GH action in the step immediately preceding test execution.

tests/sdk/nodejs/nodejs_test.go
rquitales: The helper function is for typed conditions, whereas the object we are working with is unstructured.

rquitales:
> Some questions:
> 
> 1. Why is it better to have the test code create the cluster instead of a CI step?
> 2. How/when will we support the new provider test framework and how many tests are being skipped?

1. Having a Go test framework handle cluster creation ensures test portability and allows for a unified entrypoint. This removes the requirement for the developer to pre-configure their test environment to target different test environments. It'd be nice if I could simply just run `go test --use-gke` or `go test --use-kind` for example without needing to worry if I had set up these environments manually correctly beforehand.
2. The new provider test framework was supported in the last commit.

After offline discussions, I've decided to opt for cluster creation in the CI job instead for now and parking this approach. We might revisit this in the future.

rquitales:
> Nit: Seems like we are exploiting the `--short` flag, not to "tell long-running tests to shorten their run time" but to filter incompatible tests. Ideally we'd update the tests to be compatible, e.g. by using a clusterIP.
> 
> My main feedback is that I would prefer we keep cluster creation outside of the go test code, e.g. by using a GH action in the step immediately preceding test execution.

Fwiw, most of the tests that are skipped due to not being passable in KinD in their current state are also long running tests.

.github/workflows/run-acceptance-tests.yml
rquitales, line 1: Note: changes in this file will need to be replicated in ci-mgmt. It is included here for faster feedback loop.

EronWright (approved):
LGTM!

PR #2858, opened 2024-03-01
Update version of 'glob' dependency and remove '@types/glob' to resolve imports conflict

mjeffryes:
Pulum 1.108.0 and above introduced introduced a nodejs dependency on the  arborist packaged witch imports a recent version of the "minimatch" package that exports TS types. We were importing an older version version of glob that does not provider TS types and using "@types/glob" for TS types which imports old types for minimatch. 

Updating our dependency on glob to one that exports TS types (and thus does not need the stale @types/glob), resolves the issue.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/2857


rquitales (approved):

boringuy:
I am hitting this exact same bug since last Friday, any chance to cut a release for this soon?

PR #2861, opened 2024-03-05
Ensure nodejs path globs prepend all relative path strings with ./

rquitales:
### Proposed changes

The latest version of the NodeJS `glob` package introduced in #2858 has slightly different behaviour than in the previously used version. This PR sets the options for these glob functions to maintain backwards compatibility and ensure our resource urns do not change.

### Related issues (optional)

Fixes: #2860

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## Unreleased
 
+## 4.9.0 (March 4, 2024)
+
 - Fix SSA ignoreChanges by enhancing field manager path comparisons (https://github.com/pulumi/pulumi-kubernetes/pull/2828)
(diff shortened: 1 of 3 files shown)


mjeffryes (approved):
Thanks for tracing this down! (Does this test not run on PRs? Curious why we didn't see the failure before merge.)

How confident do you feel that this is the only behavior change from the new version of glob? Do we need to do more sanity checking before release?

rquitales:
> Thanks for tracing this down! (Does this test not run on PRs? Curious why we didn't see the failure before merge.)
> 
> How confident do you feel that this is the only behavior change from the new version of glob? Do we need to do more sanity checking before release?

The PR did actually fail the presubmit test (see test results in https://github.com/pulumi/pulumi-kubernetes/pull/2858). Not sure why automerge merged it in despite the test failure.

I think this should be the only required change for `glob` settings. We only have 2 code paths that use the `glob` package, and scanning through the API changes in the changelog indicate we should probably be ok with the use cases in pu/kubernetes.

PR #2888, opened 2024-03-15
Refactor v2/yaml parsing code

EronWright:
### Proposed changes

This PR doesn't make any substantive changes to the provider, it breaks apart the parsing-and-registration code into two independent functions, in anticipation of applying component-specific sorting strategies. 

Specifics:
- split the logic into `Parse` and `Register`
- split the test code accordingly
- enriches the registration data that is collected by the mock resource monitor


provider/pkg/host/fake/monitor.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//nolint:govet
```
rquitales, line 15: Out of curiosity, what linting warning are we suppressing here?

rquitales (approved):
The refactor looks good. Thanks for extending the tests!

provider/pkg/host/fake/monitor.go
EronWright: It is due to direct use of Pulumi RPC types, which contain a lock and trigger this:
```
pkg/host/fake/monitor.go:114:9: copylocks: range var reg copies lock: github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/host/fake.Registration contains github.com/pulumi/pulumi/sdk/v3/proto/go.RegisterResourceRequest contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex (govet)
```

You're right that it is too obscure. I'm going to move the `nolint` directive to be inline on each line that needs it.

PR #2897, opened 2024-03-20
yaml/v2: apply the namespace uniformly to the resource name

### LINKED ISSUE #2870: yaml/v2: apply the namespace uniformly to the resource name
There's a minor wart in how the yaml.v2 package chooses to name the child resources, in that it may or may not incorporate the namespace. It would be more consistent if the namespace were always included (for namespaced resources only, of course). The benefit is improved name stability and regularity.

The underlying reason is that the provider includes the namespace only if it is literally specified, not defaulted. It also doesn't know whether a given kind is namespaced (but could use the dynamic client to discover it).

For example, given this manifest we'd get the following stack:
```yaml
kind: ConfigMap
metadata:
  name: cm1
---
kind: ConfigMap
metadata:
  name: cm2
  namespace: foo
```

```
Current stack resources (6):
    TYPE                                NAME
    pulumi:pulumi:Stack                 issue-1971-nodejs-dev
    ├─ kubernetes:yaml/v2:ConfigGroup   cg
    │  ├─ kubernetes:core/v1:ConfigMap  cg-cm1
    │  └─ kubernetes:core/v1:ConfigMap  cg-foo/cm2
```

I would advocate for the resultant stack to be:
```
Current stack resources (6):
    TYPE                                NAME
    pulumi:pulumi:Stack                 issue-1971-nodejs-dev
    ├─ kubernetes:yaml/v2:ConfigGroup   cg
    │  ├─ kubernetes:core/v1:ConfigMap  cg-default/cm1
    │  └─ kubernetes:core/v1:ConfigMap  cg-foo/cm2
```

There's a further complication: what if the kind is non-existent because a CRD hadn't been applied yet.  In that case, one cannot determine whether the kind is namespaced and thus whether to apply the default namespace.

EronWright:
### Proposed changes

This PR strengthens the normalization of the parsed Kubernetes objects to better interoperate with #2881.
Specifically:
1. Applies the default namespace as necessary on each object, so that they may be reliably targeted by the `depends-on` resource.
2. Stabilizes the resource id (e.g. `default/my-map` versus `my-map`).
3. Validates that the objects are well-formed more eagerly than during `Register`.

For example, given the below manifest and assuming that the default namespace is `default`, we would expect a `DependsOn` option from `Certificate` to `Issuer`. Observe that `namespace` isn't explicitly set on the issuer.

```yaml
name: issue-2870-cert-manager
runtime: yaml
resources:
  example:
    properties:
      yaml: |
        apiVersion: cert-manager.io/v1
        kind: Issuer
        metadata:
          name: test-selfsigned
        spec:
          selfSigned: {}
        ---
        apiVersion: cert-manager.io/v1
        kind: Certificate
        metadata:
          name: selfsigned-cert
          annotations:
            config.kubernetes.io/depends-on: cert-manager.io/namespaces/default/Issuer/test-selfsigned
        spec:
          dnsNames:
            - example.com
          secretName: selfsigned-cert-tls
          issuerRef:
            name: test-selfsigned
```

To apply the default namespace correctly, one must know whether a given kind is namespace-scoped. The provider naturally uses the discovery client, and also looks for a matching CRD amongst the parsed objects (since they're not yet installed).

### Testing
New test cases were added for `Normalize`, and some test cases were moved from `Register` because the validation is performed more eagerly.

New sub-tests were added for `IsNamespacedKind` to cover the local CRD enhancement.

### Related issues (optional)

Closes #2870

### DIFF
--- a/provider/pkg/clients/clients.go
+++ b/provider/pkg/clients/clients.go
@@ -167,7 +167,7 @@ func (dcs *DynamicClientSet) searchKindInGVResources(gvResources *v1.APIResource
 // For GVKs not in the table, attempt to look up the GVK from the API server. If the GVK cannot be found, a
 // NoNamespaceInfoErr is returned.
 func IsNamespacedKind(gvk schema.GroupVersionKind, disco discovery.DiscoveryInterface, objs ...unstructured.Unstructured) (bool, error) {
-	if gvk.Group == "core" {
+	if gvk.Group == "core" { // nolint:goconst
 		gvk.Group = ""
 	}
 
--- a/provider/pkg/clients/clients.go
+++ b/provider/pkg/clients/clients.go
@@ -164,7 +164,8 @@ func (dcs *DynamicClientSet) searchKindInGVResources(gvResources *v1.APIResource
 }
 
 // IsNamespacedKind checks if a given GVK is namespace-scoped. Known GVKs are compared against a static lookup table.
-// For GVKs not in the table, attempt to look up the GVK from the API server. If the GVK cannot be found, a
+// For GVKs not in the table, look at the given objects for a matching CRD.
+// Finally, attempt to look up the GVK from the API server. If the GVK cannot be found, a
 // NoNamespaceInfoErr is returned.
 func IsNamespacedKind(gvk schema.GroupVersionKind, disco discovery.DiscoveryInterface, objs ...unstructured.Unstructured) (bool, error) {
 	if gvk.Group == "core" { // nolint:goconst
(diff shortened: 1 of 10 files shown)


provider/pkg/clients/clients.go
```diff
@@ -257,6 +264,25 @@ func IsCRD(obj *unstructured.Unstructured) bool {
 		strings.HasPrefix(obj.GetAPIVersion(), "apiextensions.k8s.io/")
 }
 
+// FindCRD finds the CRD for a given kind amongst a list of unstructured objects.
+func FindCRD(objs []unstructured.Unstructured, kind schema.GroupKind) (*unstructured.Unstructured, bool) {
```
rquitales, line 268: Looks like we don't use the returned bool value, should we simplify this function's return signture?

rquitales (approved):
LGTM with 1 minor non-blocking feedback.

EronWright:
Unfortunately there's a problem with this, in preview mode, the CRDs aren't actually installed and this breaks multi-CG use cases.  The solution will either be to eat the error in preview mode, or to somehow resolve the CRDs across components.

PR #2904, opened 2024-03-22
Fix: Missing Kind in preview mode (yaml/v2)

EronWright:
### Proposed changes

The yaml/v2 package eagerly resolves kinds to apply a default namespace.  This PR addresses a problem with the use case
where a kind is installed by one `ConfigGroup` and then used by another.  In preview mode, the kind cannot be found by the latter group because the expected side-effect (i.e. installation of a CRD) doesn't occur.

The solution to the issue is to maintain a cache in the provider of new CRDs.  When the provider executes `Check` on  a `CustomResourceDefinition` resource, it adds the CRD to the cache.  Later, when a component resource must resolve a given kind, it checks the cache.  In this way, the cache compensates for the lack of side-effects (i.e. API server calls) that would normally allow the kind to be resolved.  Overall ordering is established using `DependsOn`.

It is reasonable to populate the cache during `Check` because  a)`Check` produces the "planned" state that is needed later without actuating it, and b) `Update` is called conditionally and thus isn't a reliable alternative.

### Implementation Details
The provider has three overlapping modes of operation -  preview-mode, `clusterUnreachable` and `yamlRenderMode` - that affect the initialization and use of `clientSet`. Previously when `clusterUnreachable` is true then the `clientSet` is left null. Now, we new-up the `clientSet` itself but leave its various fields null.  This change makes it possible to use the CRD cache in all modes, as was necessary to support `yamlRenderMode`.

### Testing
A new integration test is added called `yamlv2` that exercises this scenario.

Manual testing of "cluster unreachable mode" was done by misconfiguring my local environment (to have an invalid kube context).  Tested in combination with "yaml rendering mode" and found that the latter works as expected even when the cluster is unreachable.


provider/pkg/provider/resource/provider.go
```diff
@@ -12,27 +12,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package register
+package resource
```
EronWright, line 15: Fixing a typo here.

blampe (commented):
Mostly just some small questions and clarifications.

Higher level -- I definitely agree this makes for a better user experience, and how you've implemented things matches my understanding of the provider lifecycle (a single provider per operation etc.). I can't think of a better way to accomplish the desired UX given the current engine behavior.

For completeness, the strongest argument I can think of _against_ this is that it's an experience the user will only see once when a CRD is first installed. Does one preview over the lifetime of a stack warrant complexity that bumps into things like the engine/provider contract?

I don't have enough context to feel strongly one way or the other. I do prefer the UX, but my default position is to usually err on the side of simpler-is-better when brushing up against system boundaries like this. If folks from the Platform side are on board with this then it seems like a no brainer.

provider/pkg/provider/resource/provider.go
```diff
@@ -12,27 +12,27 @@
...
+type ResourceProviderOptions struct { // nolint:golint
 	ClientSet        *clients.DynamicClientSet
 	DefaultNamespace string
 }
 
-type ResourceProviderFactory func(*ResourceProviderOptions) ResourceProvider
+type ResourceProviderFactory func(*ResourceProviderOptions) ResourceProvider // nolint:golint
 
-type ResourceProviderFuncs struct {
+type ResourceProviderFuncs struct { // nolint:golint
```
blampe, line 35: Why do we need to disable lint here? You can use a format like below to make exceptions like this more self-explanatory:
```
//nolint:golint // Explanation.
```

(Note no space after the slashes since it's a machine-readable directive. `nolintlint` is helpful for that one.)

EronWright: Good idea, in this case for stuttering. I'll add commentary in future.

provider/pkg/clients/clients_test.go
```diff
@@ -116,3 +128,12 @@ func TestIsNamespacedKind(t *testing.T) {
 		})
 	}
 }
+
+type simpleDiscovery struct {
+	discoveryfake.FakeDiscovery
+}
```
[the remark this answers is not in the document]
EronWright: This is test code for `TestIsNamespacedKind` in `clients` package, and unfortunately cannot use `clients/fake` or else we get a circular dependency.  `clients/fake` is intended for provider unit testing, and it provides a "complete" fake client.

provider/pkg/provider/provider.go
```diff
@@ -763,15 +762,14 @@ func (k *kubeProvider) Configure(_ context.Context, req *pulumirpc.ConfigureRequ
 		}
 	}
 
+	var err error
+	k.clientSet, k.logClient, err = k.makeClient(k.canceler.context, config)
```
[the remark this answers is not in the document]
EronWright: Yes, though I don't know why the log client wasn't integrated more closely into the `DynamicClientSet` which contains a whole set of "clients".

provider/pkg/provider/invoke_helm_template.go
```diff
@@ -257,7 +257,7 @@ func (c *chart) template(clientSet *clients.DynamicClientSet) (string, error) {
 		installAction.APIVersions = c.opts.APIVersions
 	}
 
-	if clientSet != nil {
+	if clientSet != nil && clientSet.DiscoveryClientCached != nil {
```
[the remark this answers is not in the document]
EronWright: No, the `clientSet` is a _set of clients_ including a discovery client (that happens to have a built-in cache), a Kubernetes API client, and a REST Mapper for translating Kinds to API Resources. AFAIK the provider doesn't cache any k8s objects but does cache the discovered type information.

provider/pkg/clients/clients_test.go
```diff
@@ -81,31 +81,43 @@ var fakeResources = []*metav1.APIResourceList{
 }
 
 func TestIsNamespacedKind(t *testing.T) {
-	// coverage: in-built kinds, discoverable kinds, and kinds based on the supplied CRDs.
+	// coverage: in-built kinds, discoverable kinds, cached kinds, and kinds based on the supplied CRDs.
```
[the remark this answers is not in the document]
EronWright: That would be cool but no, I was simply explaining the sub-test cases.

EronWright:
I had a discussion with @Frassle where he expressed support but advised me to submit some tests into pu/pu to harden the contract.

Frassle:
> I don't have enough context to feel strongly one way or the other. I do prefer the UX, but my default position is to usually err on the side of simpler-is-better when brushing up against system boundaries like this. If folks from the Platform side are on board with this then it seems like a no brainer.

There are a couple of arguments against this (or adding state anywhere in providers for that matter).
1. Is it makes debug attach trickier because you probably now need a clear "Reset" signal to tell the long running attached provider process that it's time to clear its state and start again because a new deployment cycle is happening.
2. It's less performant because we can't share the provider process across the preview and up phases of `up` (well I suppose we could if we had a supported Reset method like for 1).

But I suspect providers can do _much_ smarter previews if they are stateful, and that's probably worth the downsides of the above.

provider/pkg/clients/cache.go
```diff
@@ -0,0 +1,83 @@
...
+
+// RemoveCRD removes a CRD from the cache.
+func (c *CRDCache) RemoveCRD(crd *unstructured.Unstructured) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	kind, err := groupKindFromCRD(crd)
+	if err != nil {
+		return err
+	}
+	if c.crds == nil {
```
rquitales, line 64: nit: we don't really need a nil check here since `delete()` can work on a nil map.

provider/pkg/clients/cache.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil
+	}
+	delete(c.crds, kind)
+	return nil
+}
+
+func groupKindFromCRD(crd *unstructured.Unstructured) (schema.GroupKind, error) {
+	contract.Requiref(IsCRD(crd), "crd", "expected a CRD")
+	crdGroup, _, err := unstructured.NestedString(crd.Object, "spec", "group")
+	if err != nil {
```
rquitales, line 74: Should we also ensure that NestedString was able to find the nested field? The function only errors if there are type errors, not if the field is not found.

EronWright:
Just want to mention that the `Cancel` RPC method could maybe serve as a reset signal.  Today, `Cancel` is not called unless the user cancels the deployment, but it is probably harmless to call at the end of every run, and could reset the provider state for another run. To the point about attaching to a long-running provider, cancellation is a further complication to that story unless the provider takes pains to 'reset' (p-k doesn't; it simply closes a channel).

provider/pkg/clients/cache.go
```diff
@@ -74,10 +80,16 @@ func groupKindFromCRD(crd *unstructured.Unstructured) (schema.GroupKind, error)
 	if err != nil {
 		return schema.GroupKind{}, err
 	}
+	if crdGroup == "" {
+		return schema.GroupKind{}, fmt.Errorf("expected .spec.group")
```
rquitales, line 84: nit: I usually prefer `errors.New` when I don't need to format the err message.

rquitales (approved):
LGTM

PR #2943, opened 2024-04-11
Ensure deplyoment await logic uses the latest deployment object

rquitales:
### Proposed changes

This PR ensures that we compare deployment revisions based on the old live object, and the most current deployment object. 

This PR does the following:

- Add a repro test that fails before the change is applied
- Only watch for deployment events until the deployment object has been reconciled by the deployment controller
- Update existing unit tests to account for blocking unbuffered channels, and faked objects that did not have `.status.observedGeneration`

### Related issues (optional)

Fixes: #2941

### DIFF
--- a/provider/pkg/await/deployment.go
+++ b/provider/pkg/await/deployment.go
@@ -335,6 +335,13 @@ func (dia *deploymentInitAwaiter) await(
 	timeout,
 	aggregateErrorTicker <-chan time.Time,
 ) error {
+
+	// Before we start processing any ReplicaSet, PVC or Pod events, we need to wait until the Deployment controller
+	// has seen and updated the status of the Deployment object.
+	if err := dia.waitUntilDeploymentControllerReconciles(deploymentEvents, timeout); err != nil {
+		return err
+	}
+
 	for {
 		if dia.checkAndLogStatus() {
 			return nil
@@ -365,6 +372,45 @@ func (dia *deploymentInitAwaiter) await(
 			dia.processPodEvent(event)
 		case event := <-pvcEvents:
 			dia.processPersistentVolumeClaimsEvent(event)
+		}
+	}
+}
+
+func (dia *deploymentInitAwaiter) waitUntilDeploymentControllerReconciles(deploymentEvents <-chan watch.Event, timeout <-chan time.Time) error {
+	for {
+		select {
+		case <-dia.config.ctx.Done():
+			return &cancellationError{
+				object:    dia.deployment,
+				subErrors: dia.errorMessages(),
+			}
+		case <-timeout:
+			return &timeoutError{
+				object:    dia.deployment,
+				subErrors: dia.errorMessages(),
+			}
+		case event := <-deploymentEvents:
+			deployment, isUnstructured := event.Object.(*unstructured.Unstructured)
+			if !isUnstructured {
+				logger.V(3).Infof("Deployment watch received unknown object type %T",
+					event.Object)
+				continue
+			}
+
+			// Do nothing if this is not the Deployment we're waiting for.
+			if deployment.GetName() != dia.deployment.GetName() {
+				continue
+			}
+
+			observedGeneration, found, err := unstructured.NestedInt64(deployment.Object, "status", "observedGeneration")
+			if err != nil || !found {
+				continue
+			}
+
+			if deployment.GetGeneration() == observedGeneration {
+				dia.processDeploymentEvent(event)
+				return nil
+			}
 		}
 	}
 }
--- a/provider/pkg/await/deployment_test.go
+++ b/provider/pkg/await/deployment_test.go
@@ -607,10 +607,10 @@ func Test_Apps_Deployment_With_PersistentVolumeClaims(t *testing.T) {
 				// Controller creates ReplicaSet, and it tries to progress, but
 				// it fails when there are no PersistentVolumes available to fulfill the
 				// PersistentVolumeClaim.
-				pvcs <- watchAddedEvent(
-					persistentVolumeClaimInput(inputNamespace, pvcInputName))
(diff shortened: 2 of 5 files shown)


provider/pkg/await/deployment.go
```diff
@@ -341,6 +344,10 @@ func (dia *deploymentInitAwaiter) await(
 		}
 
 		// Else, wait for updates.
+		// We run the event handlers in goroutines to avoid blocking the main event loop.
+		// We also need to ensure that only one of each event processes can run at a time.
+		var depMutex, rsMutex, podMutex, pvcMutex sync.Mutex
```
blampe, line 349: I'm a little worried this introduces room for more race conditions on `dia`, because we can still be handling multiple events at once. Re: not blocking the main loop, do you have an idea of where we're getting bottlenecked?

I wonder if there's an alternative where we simply ignore replicaset events until we see a deployment event with a generation change. Something like:

```go
select {
case event := <-deploymentEvents:
    deploying := dia.processDeploymentEvent(event)
case event := <-replicaSetEvents:
    if deploying {
        dia.processReplicaSetEvent(event)
    }
}
```

provider/pkg/await/deployment.go
```diff
@@ -689,7 +712,23 @@ func (dia *deploymentInitAwaiter) changeTriggeredRollout() bool {
...
-	return dia.deployment.GetAnnotations()[revision] != dia.config.lastOutputs.GetAnnotations()[revision]
+	// There is a race condition where the deployment object's revision and status has not been updated yet. We have to keep polling
+	// until the status is updated to ensure that the deployment controller has started the rollout. This function is called from
+	// within the replicaset event watcher code, and the deployment object is updated by the deployment event watcher code, which is why
+	// we also need to do this check here as well to ensure that the deployment object has been updated from the event log.
+	for {
+		generation := dia.deployment.GetGeneration()
+		observedGeneration, _, _ := unstructured.NestedInt64(dia.deployment.Object, "status", "observedGeneration")
+
+		if generation == observedGeneration {
```
blampe, line 723: Related to my comment above, I don't _think_ we need this polling if we ignore replica events until the deployment has started -- but this code is gnarly and there's a good chance I'm wrong!

rquitales: Agreed, the await logic is somewhat gnarly!

provider/pkg/await/deployment_test.go
```diff
@@ -1537,7 +1540,7 @@ func deploymentUpdated(namespace, name, revision string) *unstructured.Unstructu
         }
     },
     "status": {
-        "observedGeneration": 1,
+        "observedGeneration": 2,
```
rquitales, line 1543: note: the unit test had the incorrect observedGeneration for this object.

provider/pkg/await/deployment_test.go
```diff
@@ -664,12 +665,12 @@ func Test_Apps_Deployment_Without_PersistentVolumeClaims(t *testing.T) {
 				// The Deployment specifically does not reference the PVC in
 				// its spec. Therefore, the Deployment should succeed no
 				// matter what phase the PVC is in as it does not have to wait on it.
+				deployments <- watchAddedEvent(
```
rquitales, line 668: The event is reordered since we're using unbuffered channels which will block until there is a value receiver.

provider/pkg/await/deployment.go
rquitales: Yes, the initial approach in this PR may be risky. We also shouldn't be discarding replicaset events since we need those events to inform us of status updates so I'm hesitant to go down the approach you suggested.

I've reworked this PR so that we continuously poll for deployment events first, and ensure that the deployment controller has informed us if a rollout was triggered or not before we proceed onto processing other events. This ensures that when we check the replicaset, the deployment controller would have informed us if a roll-out has been triggered.

provider/pkg/await/deployment.go
```diff
@@ -369,6 +376,45 @@ func (dia *deploymentInitAwaiter) await(
...
+			if deployment.GetName() != dia.deployment.GetName() {
+				continue
+			}
+
+			observedGeneration, found, err := unstructured.NestedInt64(deployment.Object, "status", "observedGeneration")
+			if err != nil || !found {
+				continue
+			}
+
+			if deployment.GetGeneration() == observedGeneration {
```
blampe, line 410: Would `generation <= observedGeneration` be technically safer here? I don't know if that's actually observable in practice.

EronWright (approved):
LGTM

provider/pkg/await/deployment.go
```diff
@@ -335,6 +335,13 @@ func (dia *deploymentInitAwaiter) await(
 	timeout,
 	aggregateErrorTicker <-chan time.Time,
 ) error {
+
+	// Before we start processing any ReplicaSet, PVC or Pod events, we need to wait until the Deployment controller
+	// has seen and updated the status of the Deployment object.
+	if err := dia.waitUntilDeploymentControllerReconciles(deploymentEvents, timeout); err != nil {
+		return err
+	}
```
blampe, line 343: 👍

blampe (approved):

provider/pkg/await/deployment.go
rquitales: We need the `observedGeneration` to match exactly the `generation` of the deployment object. `generation <= observedGeneration` is currently why this bug exists.

EronWright: I too stared at the comparison for a few minutes.  Rollout code will sometimes have a 'planned' generation that they got from the call to `Apply`, and await an observed generation that is greater-or-equal to the planned generation.  This code is not doing that; it is simply comparing generations within a given object and has no planned generation.

PR #2952, opened 2024-04-15
Fix a data race in watcher.go

blampe:
Ran into a small data race while testing await logic.

This PR fixes the underlying race condition as well as a race-y test.

### DIFF
--- a/provider/pkg/watcher/watcher_test.go
+++ b/provider/pkg/watcher/watcher_test.go
@@ -17,10 +17,10 @@ package watcher
 import (
 	"context"
 	"strings"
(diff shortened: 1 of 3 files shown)


provider/pkg/watcher/watcher_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"testing"
 	"time"
 
+	"go.uber.org/atomic"
```
rquitales, line 23: Could we use `sync/atomic` instead?

rquitales (approved):

Issue #609, opened 2019-06-24
Add await logic for DaemonSets


RichardWLaub:
We are deploying the sysdig helm chart that creates a daemonSet with a rollingUpdate updateStrategy. We immediately run tests to make sure the pods are running correctly afterward. With `deployment`s and `service`s our tests always pass because pulumi waits to verify that those kinds deployed successfully. Our daemonSet tests sometimes fail because there is no await logic for daemonsets. 

Simple repro:
```
import * as k8s from '@pulumi/kubernetes';

new k8s.helm.v2.Chart('simple-nginx-local', {
  path: 'ds-chart',
  values: {
    image: {
      tag: '1.16'
    }
  }
});
```
With a chart with template `ds-chart/templates/daemonset.yaml`:
```
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  name: test
spec:
  selector:
    matchLabels:
      name: nginx-ds
  template:
    metadata:
      labels:
        name: nginx-ds
      name: nginx-ds
    spec:
      containers:
        - name: {{ .Chart.Name }}
          image: "nginx:{{ .Values.image.tag }}"
          imagePullPolicy: {{ .Values.image.pullPolicy }}
  updateStrategy:
    rollingUpdate:
      maxUnavailable: 1
    type: RollingUpdate
```
Toggle the tag above to `1.17` to see that the pulumi program finishes before the new pods are ready.

casey-robertson:
Raising awareness.  We had a sysdig outage that could have been mitigated sooner if this was in place.

blampe:
@RichardWLaub @casey-robertson @rquitales I wanted to share the current approach I'm taking with this in case you have any concerns or suggestions.

For DaemonSets with a RollingUpdate strategy, we will essentially follow the same behavior as [kubectl rollout status](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_rollout/kubectl_rollout_status/). Pulumi will wait until the rollout status is ready unless the DS has a skipAwait annotation.

It's less clear how to handle the OnDelete update strategy. For background, we currently handle StatefulSets with OnDelete update strategies by waiting for all pods to be manually removed (https://github.com/pulumi/pulumi-kubernetes/pull/2473); if the user doesn't want this behavior they must annotate the OnDelete StatefulSet with skipAwait.

I'm preserving this behavior for OnDelete DaemonSets for consistently, although I see an argument that Pulumi should not wait for this update strategy at all.

Issue #2799, opened 2024-01-31
Unittests for await logic (Update)


EronWright:
Develop unit test cases for the `Update` function of the `await` package.

PR #2953, opened 2024-04-15
Add await logic for DaemonSets

blampe:
### Proposed changes

This adds await logic for DaemonSets with RollingUpdate or OnDelete update strategies.

The implementation is largely based on the existing StatefulSet logic with two high-level simplifications:

1. We use [kstatus](https://pkg.go.dev/sigs.k8s.io/cli-utils/pkg/kstatus/status) to decide when a DaemonSet is ready.
2. We use a `PodAggregator` to handle reporting pod statuses.

Importantly, unlike StatefulSet this means we do not currently inspect pods to decide readiness -- we only use them for informational purposes. I _think_ this is sufficient but I could easily be missing something. I haven't been able to simulate situations where this logic doesn't fully capture readiness and we would need to inspect pod statuses.

A failing e2e test was added in YAML. I noticed we didn't have any YAML tests yet, and 
I've found YAML to be quick and easy to crank out for very simple test cases like this.

Unit tests were added around the public `Creation`, `Update`, etc. methods in order to more fully exercise timeouts and retries. To that end I introduced a mock clock package which might be controversial. IMO Go doesn't have a great OSS mock clock but something like this can be very helpful for testing.

I'm still somewhat confused by the role of `await.Read` since it doesn't actually await anything, but it's implemented similar to StatefulSet as a one-shot read + readiness check.

### Related issues

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/609
Refs https://github.com/pulumi/pulumi-kubernetes/issues/2800
Refs https://github.com/pulumi/pulumi-kubernetes/issues/2799
Refs https://github.com/pulumi/pulumi-kubernetes/issues/2798


rshade:
@blampe thanks for the await logic, this will be a big help. Did you happen to test the two following scenarios: 
1. Updating to a new image(Working)
2. Updating to a new image(Broken). 
I ask because #2 specifically has caused us problems before the waiter.

blampe:
> @blampe thanks for the await logic, this will be a big help. Did you happen to test the two following scenarios:
> 
> 1. Updating to a new image(Working)
> 2. Updating to a new image(Broken).
>    I ask because 2 specifically has caused us problems before the waiter.

@rshade I added a failing e2e test [here](https://github.com/pulumi/pulumi-kubernetes/pull/2953/files#diff-0801c1d6adc734510cc1aca3fcaeda5f9171f11dc27da277f519aac73b9004d1R27) which should cover both of these scenarios.

[This step](https://github.com/pulumi/pulumi-kubernetes/pull/2953/files#diff-b8ac385b3c4a6f4c25271f89dcbf3b12175c5b00529458f200560458431584fa) in particular captures scenario (2) by updating to an invalid image. Previously the operation would succeed, but after this PR it will fail after it times out waiting for the broken pods to come up.

Let me know if there are any other test cases you have in mind!

provider/pkg/await/daemonset.go
```diff
@@ -0,0 +1,356 @@
...
+
+// rolloutComplete checks whether we've succeeded, and logs the result as a
+// status message to the provider.
+func (dsa *dsAwaiter) rolloutComplete() bool {
+	dssv := polymorphichelpers.DaemonSetStatusViewer{}
+
+	msg, done, err := dssv.Status(dsa.ds, 0)
+	if done && err != nil {
+		// If done=true and err is non-nil then this is an OnDelete rollout.
+		msg, done, err = dsa.onDeleteRolloutStatus()
```
EronWright, line 286: This is a pretty awkward handoff, and given how little code is in `DaemonSetStatusViewer`, maybe we shouldn't use it.

tests/sdk/yaml/yaml_test.go
```diff
@@ -0,0 +1,73 @@
...
+			{
+				// Await unsuccessful update -- should fail.
+				Dir:           filepath.Join(dir, "step3"),
+				Additive:      true,
+				ExpectFailure: true,
+			},
+		},
+	}
+
+	integration.ProgramTest(t, &test)
```
EronWright, line 72: Nit: I would advocate to use the new provider test framework as seen in `tests/provider/yamlv2_test.go`.

tests/sdk/yaml/await-daemonset/Pulumi.yaml
```diff
@@ -0,0 +1,35 @@
+name: await-daemonset
+runtime: yaml
+plugins:
+  providers:
+    - name: kubernetes
+      path: ../../../../bin
```
EronWright, line 6: I thought the provider linking happened automatically but I frankly don't know how.  And why don't we see it in the other steps?

tests/sdk/yaml/await-daemonset/step2/Pulumi.yaml
EronWright, line 1: Could you annotate the lines that are changing from steps 1-2-3?

EronWright (commented):
I don't know the awaiters too well but everything seemed good to me.

Makefile
```diff
@@ -132,6 +132,7 @@ test_fast::
 	cd tests/sdk/python && $(GO_TEST_FAST) ./...
 	cd tests/sdk/dotnet && $(GO_TEST_FAST) ./...
 	cd tests/sdk/go && $(GO_TEST_FAST) ./...
+	cd tests/sdk/yaml && $(GO_TEST_FAST) ./...
```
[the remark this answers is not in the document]
blampe: Perfect, totally gets around the lack of `test_yaml` on CI. Will do!

provider/pkg/await/daemonset.go
blampe: I switched this over to use kstatus since it's essentially the same logic and doesn't distinguish between Rolling or OnDelete rollouts.

tests/sdk/yaml/await-daemonset/Pulumi.yaml
blampe: Yeah it happens automatically in tests, although sometimes I will manually run examples in which case this helps. I took it out since I don't need to test it manually anymore.

tests/sdk/java/testdata/await/daemonset/step3/Pulumi.yaml
```diff
@@ -0,0 +1,31 @@
...
+            labels:
+              "foo": "bar"
+          spec:
+            containers:
+              - name: nginx
+                image: nginx:invalid-tag # Invalid tag -- this will never become ready.
+        minReadySeconds: 5
+    options:
+      customTimeouts:
+        create: 30s
```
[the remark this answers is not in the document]
blampe: Nope, just forgot to update this one - good catch!

rquitales (approved):

Issue #2731, opened 2023-12-26
Empty array value still ignored in Helm Chart


sebastien-prudhomme:
### What happened?

I'm trying to empty a list in the Velero Helm chart.

I expect that the resulting value used to instanciate the Helm chart is also an empty list. Nothing is take into account by the Pulumi provider, even when I use the "allow_null_values" parameter.

I've tried to replace the empty list by None (Python) or null (Typescript) without any success.

The result is that the corresponding YAML key is ignored by Pulumi and the value configured in the Helm chart is always used.

### Example

I've tested this Python code:

```
import pulumi
import pulumi_kubernetes as kubernetes

velero_ns = kubernetes.core.v1.Namespace(
    "velero",
    metadata=kubernetes.meta.v1.ObjectMetaArgs(
        name="velero"
    )
)

velero = kubernetes.helm.v3.Release(
    "velero",
    allow_null_values=True,
    chart="velero",
    name="velero",
    namespace=velero_ns.metadata.name,
    repository_opts=kubernetes.helm.v3.RepositoryOptsArgs(
        repo="https://vmware-tanzu.github.io/helm-charts/",
    ),
    values={
        "configuration": {
            "backupStorageLocation": [],
            "volumeSnapshotLocation": []
        }
    },
    version="5.2.0"
)
```

and also this Typescript code:

```
import * as pulumi from "@pulumi/pulumi";
import * as kubernetes from "@pulumi/kubernetes";

const veleroNs = new kubernetes.core.v1.Namespace("velero", {metadata: {
    name: "velero",
}});

const velero = new kubernetes.helm.v3.Release("velero", {
    chart: "velero",
    allowNullValues: true,
    name: "velero",
    namespace: veleroNs.metadata.name,
    repositoryOpts: {
        repo: "https://vmware-tanzu.github.io/helm-charts/",
    },
    values: {
        configuration: {
            backupStorageLocation: [],
            volumeSnapshotLocation: [],
        },
    },
    version: "5.2.0",
});
```

### Output of `pulumi about`

```
CLI
Version      3.99.0
Go Version   go1.21.5
Go Compiler  gc

Plugins
NAME        VERSION
kubernetes  4.6.1
python      unknown

Host
OS       ubuntu
Version  22.04
Arch     x86_64

This project is written in python: executable='/usr/bin/python3' version='3.10.12'

Current Stack: organization/pulumi-bug/dev

TYPE                           URN
[...]

PR #2963, opened 2024-04-18
Fix deep merge of ValuesYamlFiles

### LINKED ISSUE #2958: Incorrect merge of `ValuesYamlFile` for `helmv3.NewRelease`
### What happened?

Helm values are not merged as expected when providing an array with several yaml files to `ValueYamlFiles`. 
For instance with the following values:
(1)
```
image:
  repository: bitnami/nginx
```
and 
(2)
```
image:
  tag: 1.25.0
```

We'd expect the following values to be applied
```
image:
  tag: 1.25.0
  repository: bitnami/nginx
```
However only (1) or (2) is applied depending on the order you put them in the list.

Note: The behavior described above is the current Helm behavior, when specifying several values file, Helm will deep merge them together. https://github.com/helm/helm/issues/1620

The following lines seem to be responsible for the unexpected behavior: https://github.com/pulumi/pulumi-kubernetes/blob/ee61f79a97bf5e4d62ddedfa2ad52f76445cfdca/provider/pkg/provider/helm_release.go#L284-L321
The merge of the values rely on `yaml.Unmarshall` which does not handle correctly the nested fields.

Related PR: https://github.com/pulumi/pulumi-kubernetes/pull/2963

Let me know if I should add anything to the issue

### Example

```
package main

import (
	corev1 "github.com/pulumi/pulumi-kubernetes/sdk/v4/go/kubernetes/core/v1"
	helmv3 "github.com/pulumi/pulumi-kubernetes/sdk/v4/go/kubernetes/helm/v3"
	metav1 "github.com/pulumi/pulumi-kubernetes/sdk/v4/go/kubernetes/meta/v1"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi/config"
)

func main() {
		cfg := config.New(ctx, "")
		k8sNamespace, err := cfg.Try("k8sNamespace")
		if err != nil {
			k8sNamespace = "default"
		}
		appLabels := pulumi.StringMap{
			"app": pulumi.String("nginx-ingress"),
		}

		// Create a new namespace (user supplies the name of the namespace)
		ingressNs, err := corev1.NewNamespace(ctx, "ingressns", &corev1.NamespaceArgs{
			Metadata: &metav1.ObjectMetaArgs{
				Labels: pulumi.StringMap(appLabels),
				Name:   pulumi.String(k8sNamespace),
			},
		})
		if err != nil {
			return err
		}
		helmValues1 := `
image:
  repository: bitnami/nginx
  `
		helmValues2 := `
image:
  tag: 1.25.0  
`
		var helmValues pulumi.AssetOrArchiveArray
		helmValues = append(helmValues, pulumi.NewStringAsset(helmValues2), pulumi.NewStringAsset(helmValues1))
		// Use Helm to install the Nginx ingress controller
		ingresscontroller, err := helmv3.NewRelease(ctx, "ingresscontroller", &helmv3.ReleaseArgs{
			Chart:     pulumi.String("nginx"),
			Namespace: ingressNs.Metadata.Name(),
			RepositoryOpts: &helmv3.RepositoryOptsArgs{
				Repo: pulumi.String("https://charts.bitnami.com/bitnami"),
			},
			SkipCrds:       pulumi.Bool(true),
			ValueYamlFiles: helmValues,
			Version:        pulumi.String("16.0.3"),
		})
		if err != nil {
			return err
		}

		// Export some values for use elsewhere
		ctx.Export("name", ingresscontroller.Name)
		return nil
	})
}
```
Running the following code with `kubectl` properly configured to a Kubernetes cluster can show the issue.

You will find the deployed resource with different tag on the image depending on the order of the helmValues:
`append(helmValues, pulumi.NewStringAsset(helmValues2), pulumi.NewStringAsset(helmValues1))`
or 
`append(helmValues, pulumi.NewStringAsset(helmValues1), pulumi.NewStringAsset(helmValues2))`

### Output of `pulumi about`

```
CLI          
Version      3.103.1
Go Version   go1.21.5
Go Compiler  gc

Plugins
NAME        VERSION
go          unknown
kubernetes  4.0.3

Host     
OS       darwin
Version  13.5
Arch     arm64

This project is written in go: executable='/usr/local/go/bin/go' version='go version go1.21.6 darwin/arm64'

Current Stack: organization/pulumi-helm-debug/dev

pulumi:pulumi:Stack            urn:pulumi:dev::pulumi-helm-debug::pulumi:pulumi:Stack::pulumi-helm-debug-dev
pulumi:providers:kubernetes    urn:pulumi:dev::pulumi-helm-debug::pulumi:providers:kubernetes::default
kubernetes:core/v1:Namespace   urn:pulumi:dev::pulumi-helm-debug::kubernetes:core/v1:Namespace::ingressns
kubernetes:helm.sh/v3:Release  urn:pulumi:dev::pulumi-helm-debug::kubernetes:helm.sh/v3:Release::ingresscontroller


OPP TYPE  URN
creating  urn:pulumi:dev::pulumi-helm-debug::kubernetes:helm.sh/v3:Release::ingresscontroller
creating  urn:pulumi:dev::pulumi-helm-debug::kubernetes:helm.sh/v3:Release::ingresscontroller


Backend        
Name           
URL            file://~
User          
Organizations  
Token type     personal

Dependencies:
github.com/pulumi/pulumi-kubernetes/sdk/v4  4.0.3
github.com/pulumi/pulumi/sdk/v3             3.96.1

Pulumi locates its logs in /var/folders/_b/sjpwr13931b95__1m3thg78h0000gp/T/ by default
```

### Additional context

_No response_

### Contributing

KevinFairise2:
### Proposed changes

Stop relying on `yaml.Unmarshal` to merge `ValuesYaml` as it cannot do a deep merge. Instead we Unmarshal the values in a temporary map and use the `mergeMap` function that already exists to merge.

With this the behavior of the provider is aligned with the deep merge made by Helm https://github.com/helm/helm/issues/1620

Note: This is a breaking change

### Related issues (optional)
Should close #2958


rquitales:
@KevinFairise2 Thanks for identifying this issue and providing a fix. The overall code looks good to me. It'd be nice to have some tests to exercise this new logic. Would you be able to add a simple test case for this?

provider/pkg/provider/helm_release.go
```diff
@@ -299,9 +299,15 @@ func decodeRelease(pm resource.PropertyMap, label string) (*Release, error) {
 					if err != nil {
 						return nil, err
 					}
-					if err = yaml.Unmarshal(b, &values); err != nil {
+					valuesMap := map[string]any{}
+					if err = yaml.Unmarshal(b, &valuesMap); err != nil {
 						return nil, err
 					}
+					mergedValues, err := mergeMaps(values, valuesMap, false)
```
blampe, line 306: I notice we might still deviate from upstream's behavior here -- this is Helm's logic:

https://github.com/helm/helm/blob/14d0c13e9eefff5b4a1b511cf50643529692ec94/pkg/cli/values/options.go#L108-L125

But we're using [github.com/imdario/mergo.Merge](https://github.com/darccio/mergo/blob/v0.3.13/merge.go#L345-L367). I haven't looked at this closely enough to see if there are behavioral differenced -- it's quite a bit more complex so I imagine there must be?

I would feel a lot better if we followed upstream's `mergeMaps` logic directly. It's not public but I wonder if we should just vendor the logic to stay closer to upstream. Or maybe this is an edge case not worth worrying about?

EronWright: I would definitely advocate for using Helm itself as much as possible, copying as necessary.

KevinFairise2: Thanks for your fast feedbacks on that PR!

Using the same function as Helm to merge the maps sounds look I good idea.
`mergo.Merge` can do a deep merge on any struct, so the code is indeed way more complex. In our case we only need to deep merge two `map[string]any` so using the same function as Helm should be enough.
I can replace the call to `mergo.Merge` [here](https://github.com/pulumi/pulumi-kubernetes/blob/master/provider/pkg/provider/helm_release.go#L1331) with a call to a copy of the Helm merge maps. If that sounds good to you. As you said it is hard to anticipate any behavior changes seeing how complect the `mergoM.Merge` implementation is.
Do you think your test would catch any changes in the behavior of `mergeMaps`?

Let me know how I should handle License things if I copy the function from Helm. The code they use seems to be standard way to deep merge two `map[string]any` that can be found elsewhere on the internet

blampe: I pushed a change to use upstream's merge logic. Both projects are Apache 2.0 and they don't include a `NOTICE` so I think it's sufficient to preserve attribution.

I left the existing `mergeMaps` implementation alone because I'd like to keep the scope of this change as small as possible for now. I'm not sure how well our tests cover that behavior.

provider/pkg/provider/helm_release_test.go
```diff
@@ -138,3 +140,41 @@ func Test_MergeMaps(t *testing.T) {
...
+func TestDecodeRelease(t *testing.T) {
+	tests := []struct {
+		name  string
+		given resource.PropertyMap
+		want  *Release
+	}{
+		{
+			name: "valueYamlFiles layering",
+			given: resource.PropertyMap{
+				"valueYamlFiles": resource.NewArrayProperty([]resource.PropertyValue{
```
EronWright, line 153: Maybe this test should also include a literal value, to test how the value files combine with literals.  e.g. a literal value of `repository: bitnami/nginx-patched`.

This test should check the relative priority of the files, by setting `tag: latest` into the first file (and would be superseded by the second file).

provider/pkg/provider/helm_release.go
```diff
@@ -299,9 +300,11 @@ func decodeRelease(pm resource.PropertyMap, label string) (*Release, error) {
 					if err != nil {
 						return nil, err
 					}
-					if err = yaml.Unmarshal(b, &values); err != nil {
+					valuesMap := map[string]any{}
+					if err = yaml.Unmarshal(b, &valuesMap); err != nil {
 						return nil, err
 					}
+					values = helm.MergeMaps(values, valuesMap)
```
EronWright, line 307: Should we also merge the literal values using `helm.MergeMaps` rather than our own `mergeMaps` (see line 319)?   Perhaps use `helm.MergeMaps` when `AllowNullValues` is true (since the flag was intended to improve compatibility). Are there any differences in the logic in that case? Or we could leave it for the v4 implementation of `Release`.

EronWright (approved):

provider/pkg/provider/helm_release.go
blampe: This is a good question. I added some test cases and I think this basically boils down to whether we also want to fix https://github.com/pulumi/pulumi-kubernetes/issues/2731 as part of this PR, because I think that's essentially what your suggestion would do.

I'm OK to leave line 319 as-is just to keep scope and risk down. I can follow up with #2731.

provider/pkg/provider/helm_release.go
KevinFairise2: PR looks good to me like that. I think everything is in your hand to have it merged now. Let me know if I can do anything more to help with that.
What will be the release plan for that? I think this is a breaking change that can impact current users of the provider.

blampe: > I think this is a breaking change that can impact current users of the provider.

It's breaking in the sense that behavior will change, yes. In this case it's a bug that this field doesn't agree (and has never agreed) with helm's behavior despite claiming to. I'm comfortable fixing the behavior instead of forcing users to live with it until a much later date for v5 -- or introducing another awkward field `valueYamlFilesThisIsTheOneYouReallyWant`.

Correctness is more important since the current behavior can potentially cause cascading issues by unexpectedly dropping values. If someone has been impacted by this, they might not realize it (in which case they could be sitting on a time bomb) or they might have worked around it by providing a union of values (in which case the change in behavior is less impactful). 

In any case, previews will show a diff with any impacted values and users can choose to modify their values accordingly.

Issue #1970, opened 2022-04-19
CustomResource panic when changing name field in Tekton Pipeline


jsravn:
### What happened?

I'm trying to modify a Tekton Pipeline via `kubernetes.apiextensions.CustomResource`. If I try changing a step name, I get a panic.

```
pulumi up
Previewing update (extended-test):
     Type                                       Name                         Plan     Info
     pulumi:pulumi:Stack                        extended-test-extended-test           44 messages     └─ kubernetes:tekton.dev/v1beta1:Pipeline  test-pipeline                         1 error
 
Diagnostics:
  kubernetes:tekton.dev/v1beta1:Pipeline (test-pipeline):
    error: transport is closing
 
  pulumi:pulumi:Stack (extended-test-extended-test):
    panic: fatal: An assertion has failed: path: [spec tasks 0 taskSpec spec]  |  v: <nil>  | old: <nil>  |  oldInput: <nil>
    goroutine 308 [running]:
    github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.failfast(...)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.27.0/go/common/util/contract/failfast.go:23
    github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.Assertf(0x80, {0x2204c9c, 0x8}, {0xc003129cc8, 0x203000, 0x203000})
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.27.0/go/common/util/contract/assert.go:33 +0xf4
    github.com/pulumi/pulumi-kubernetes/provider/v3/pkg/provider.(*patchConverter).addPatchValueToDiff(0xc00312b398, {0xc003422c80, 0x1db06a0, 0x8}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, ...)
    	/home/runner/work/pulumi-kubernetes/pulumi-kubernetes/provider/pkg/provider/provider.go:2887 +0x188
    github.com/pulumi/pulumi-kubernetes/provider/v3/pkg/provider.(*patchConverter).addPatchMapToDiff(0xc003219540, {0xc003219540, 0x4, 0x4}, 0x2, 0x46acb9, 0x1d55fe0, 0x1f79ea0, 0x1)
    	/home/runner/work/pulumi-kubernetes/pulumi-kubernetes/provider/pkg/provider/provider.go:2989 +0x3f4
[...]

PR #2987, opened 2024-05-01
Fix a panic when diffing nil values against missing keys

blampe:
Users can run into a panic because we invoke `addPatchValueToDiff` like so:

```go
pc.addPatchValueToDiff(append(path, k), v, old[k], newInput[k], oldInput[k], inArray)
```

These are `map[string]any` types and will return `nil` when the `[k]` lookup returns `nil` _or_ when the key `k` is missing.

Then in `addPatchValueToDiff` we assert:
```go
contract.Assertf(v != nil || old != nil || oldInput != nil || newInput != nil)
```

but we don't have enough information to distinguish the "all values were `nil`" case from the "one value was `nil` and the other was missing" case. The assertion is only valid if all keys were present, hence we panic when a key is removed when it previously had a value of `nil`.

The easiest fix is to relax the assertion (b886cfb8d) -- allow the all-`nil` case since it can be valid. The downside is a poorer user experience, because we don't have enough information to say whether the update is an add or a delete; we could call it an update (odd) or a no-op (lossy). The upside is it's a much smaller change with less risk.

Alternatively, we can have a better user experience by making `addPatchValueToDiff` aware of whether keys were missing. This PR does this by introducing a `missing` sentinel to distinguish from `nil` values. This is riskier, but fortunately our test coverage is comprehensive.

A failing unit test was added, along with another test to cover a small bit of logic we weren't testing.

I spent a bit of time trying to cook up an E2E repro but ran into some roadblocks so I put that on the back burner for now.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/1970.

### DIFF
--- a/provider/pkg/provider/diff_test.go
+++ b/provider/pkg/provider/diff_test.go
@@ -16,9 +16,9 @@ import (
 )
 
 type (
-	object   map[string]any
-	list     []any
-	expected map[string]*pulumirpc.PropertyDiff
+	object   = map[string]any
+	list     = []any
+	expected = map[string]*pulumirpc.PropertyDiff
 )
 
 func TestPatchToDiff(t *testing.T) {
--- a/provider/pkg/provider/diff_test.go
+++ b/provider/pkg/provider/diff_test.go
@@ -16,9 +16,9 @@ import (
 )
 
 type (
-	object   = map[string]any
-	list     = []any
-	expected = map[string]*pulumirpc.PropertyDiff
+	object   map[string]any
+	list     []any
+	expected map[string]*pulumirpc.PropertyDiff
 )
 
 func TestPatchToDiff(t *testing.T) {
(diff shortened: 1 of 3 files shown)


EronWright (approved):

provider/pkg/provider/diff_test.go
```diff
@@ -15,9 +15,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-type object = map[string]any
-type list = []any
-type expected = map[string]*pulumirpc.PropertyDiff
+type (
+	object   = map[string]any
```
rquitales, line 19: nit: should we omit `=` to align with how we define types elsewhere.

rquitales (approved):

provider/pkg/provider/diff_test.go
blampe: Ah, sorry, should have realized these are type aliases to make the test assertions easier to work with. We should leave these as-is.

PR #2990, opened 2024-05-02
Remove deprecated pkg/errors package

blampe:
`errors.Wrap` can mask bugs by returning `nil` errors where a non-`nil` error was intended. This is most common in situations where multiple errors are being checked and the wrong (`nil`) error is wrapped:

```go
if err != nil {
    return err
}
if _, otherErr := f(...); otherErr != nil {
    return errors.Wrap(err, ...) // bug
}
```
From spot checking our usage I don't _think_ we have any cases of this, but we should stop using the deprecated package to be safe. Additionally the native errors package now has `Join` and other helpers we should be using instead.

Most of this PR was programmatically generated.


provider/pkg/provider/provider.go
```diff
@@ -2364,19 +2366,21 @@ func (k *kubeProvider) Update(
...
 		if getErr != nil {
 			// Object update/creation failed.
-			return nil, pkgerrors.Wrapf(
-				awaitErr, "update of resource %q failed because the Kubernetes API server "+
-					"reported that it failed to fully initialize or become live", urn)
+			return nil, fmt.Errorf(
+				"update of resource %q failed because the Kubernetes API server "+
+					"reported that it failed to fully initialize or become live: %w", urn, getErr)
+
 		}
```
blampe, line 2384: This was returning `awaitErr` instead of `getErr`.

provider/pkg/provider/helm_release.go
```diff
@@ -348,7 +347,7 @@ func (r *helmReleaseProvider) Check(ctx context.Context, req *pulumirpc.CheckReq
 		KeepSecrets:  true,
 	})
 	if err != nil {
-		return nil, pkgerrors.Wrapf(err, "check failed because malformed resource inputs: %+v", err)
+		return nil, fmt.Errorf("check failed because malformed resource inputs: %+v: %w", err, err)
```
EronWright, line 350: Seems unintentional that `err` is used twice.

provider/pkg/provider/helm_release.go
```diff
@@ -392,7 +391,7 @@ func (r *helmReleaseProvider) Check(ctx context.Context, req *pulumirpc.CheckReq
...
```
EronWright, line 394: Seems unintentional that `err` is used twice.

provider/pkg/provider/helm_release.go
blampe: Good catches! This was duplicating the error and my script preserved the behavior because it's just another formatting arg. The original author might have forgotten or not realized `Wrapf` already did this automatically.

provider/pkg/provider/provider.go
rquitales: While we're cleaning up errs, the original error message also seems like it could be improved upon as well. The code comment directly below: `but failed to initialize` conflicts with the error message we're returning: `reported that it failed to fully initialize or become live`. Something along the lines of object creation failed seems to be more fitting, since we send the patch update, but the object still does not exist on the cluster.

rquitales (approved):

Issue #2034, opened 2022-06-18
Unable to delete a Chart default key


aureq:
### What happened?

In a Helm Chart, it's possible to delete a default key (as provided by the chart defaults) using the `null` value. On the command line, this looks like this.

```bash
helm install stable/drupal --set image=my-registry/drupal:0.1.0 --set livenessProbe.httpGet=null
```

This feature is also [documented](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key).

However, when using this provider and the [descheduler](https://github.com/kubernetes-sigs/descheduler/tree/master/charts/descheduler) Chart, it doesn't seem to be possible to achieve the same by providing a `null`. 

Here are the details below:

In the `values.yaml` default values we find [`RemovePodsViolatingNodeAffinity`](https://github.com/kubernetes-sigs/descheduler/blob/master/charts/descheduler/values.yaml#L72-L76).

```yaml
    RemovePodsViolatingNodeAffinity:
      enabled: true
      params:
        nodeAffinityType:
        - requiredDuringSchedulingIgnoredDuringExecution
```

Of this item, we want to remove the `params` keys. In TypeScript we construct the following object for the `deschedulerPolicy` and we set `params: null`.

```
deschedulerPolicy: {
            strategies: {
              RemoveDuplicates: {
                enabled: false,
                params: null,
              },
              LowNodeUtilization: {
                enabled: false,
                params: null,
              },
              HighNodeUtilization: {
                enabled: false,
                params: null,
              },
              RemovePodsViolatingInterPodAntiAffinity: {
                enabled: false,
                params: null,
              },
              RemovePodsViolatingNodeAffinity: {
                enabled: false,
                params: null,
              },
              RemovePodsViolatingNodeTaints: {
                enabled: false,
                params: null,
              },
              RemovePodsViolatingTopologySpreadConstraint: {
                enabled: false,
                params: null,
              },
              RemovePodsHavingTooManyRestarts: {
                enabled: false,
                params: null,
              },
              PodLifeTime: {
                enabled: false,
                params: null,
              },
              RemoveFailedPods: {
                enabled: false,
                params: null,
              },
            },
          },
```
Unfortunately, the resulting configMap is like this and still contains the `params` key.

```yaml
strategies:
  HighNodeUtilization:
    enabled: false
  LowNodeUtilization:
    enabled: false
    params:
      nodeResourceUtilizationThresholds:
        targetThresholds:
          cpu: 50
[...]

PR #2995, opened 2024-05-03
Fix value merging for empty lists

blampe:
This fixes our behavior to match upstream's when merging empty lists. We now produce an empty list instead of a no-op.

The bug was due to [this logic](https://github.com/pulumi/pulumi-kubernetes/blob/13c87c632b4aa5f36c5b9d119acffada769b4a9e/provider/pkg/provider/helm_release.go#L1385-L1391) which initialized a slice as `nil`. This was then passed to `mergo.Merge` which ignores zero values by default, hence why the original value was not overridden.

This PR changes our `allowNullValues` logic to use upstream's merge behavior more directly since the intention behind that flag is to match upstream as closely as possible.

Importantly, this _doesn't_ fix https://github.com/pulumi/pulumi-kubernetes/issues/2034 because we're still unmarshaling inputs with `SkipNulls: true`. In practice `allowNullValues` behaves more like _`allowEmptyValues`_ because our unmarshaling strips away all of the meaningful `nil` values. That fix will be in another PR.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/2731.
Refs https://github.com/pulumi/pulumi-kubernetes/issues/2034.
Refs https://github.com/pulumi/pulumi-kubernetes/pull/2963.


rquitales (approved):
Seems reasonable that we'd want to match as close to upstream as possible.

Issue #2948, opened 2024-04-13
Named resources get deleted from cluster after pulumi resource name change, even though indicated otherwise by pulumi


sfc-gh-jlangefeld:
### What happened?

When changing the pulumi resource name of a kubernetes resource that was named via 
```go
	Metadata: metav1.ObjectMetaArgs{
		Name: pulumi.String("test"),
	},
```
and then running `pulumi up`, pulumi indicates that the resource gets deleted and created (as expected), but in reality the resource only gets deleted.

This is a risk for us in the case of resource name changes. Especially that the preview and up indicate that the resource should be created again, but it never was in reality.

We rely on resource naming via `metadata.name` because that's the only way to adopt existing resource, a common use case for us.

### Example

Here are some good steps to reproduce and to exemplify the behavior:

1. Create kind cluster via `kind create cluster`
2. Create the quickstart project as described in https://www.pulumi.com/docs/clouds/kubernetes/get-started/create-project/
3. Add a Kubernetes `metadata.name` in the `appsv1.DeploymentArgs` struct via:
    ```go
	Metadata: metav1.ObjectMetaArgs{
		Name: pulumi.String("test"),
	},
    ```
5. Run `pulumi up` and make sure the deployment was created:
    ```yaml
    ╰─ kubectl get deploy test
    NAME   READY   UP-TO-DATE   AVAILABLE   AGE
    test   1/1     1            1           10s
    ```
6. Change the name of the pulumi resource to something new:
    ```diff
    -deployment, err := appsv1.NewDeployment(ctx, "app-dep", &appsv1.DeploymentArgs{
    +deployment, err := appsv1.NewDeployment(ctx, "app-dep1", &appsv1.DeploymentArgs{
    ```
7. Run `pulumi preview` and it shows correctly that it wants to delete `test` and create `test1`:
    ```
    ╰─ pulumi preview
    Previewing update (dev)
    
    View in Browser (Ctrl+O): https://app.pulumi.com/jonny-langefeld/quickstart3/dev/previews/eb071e68-ace0-4ce3-a36e-fecf4f91f419
    
         Type                              Name             Plan
         pulumi:pulumi:Stack               quickstart3-dev
     +   ├─ kubernetes:apps/v1:Deployment  app-dep1         create
     -   └─ kubernetes:apps/v1:Deployment  app-dep          delete
    
    Resources:
        + 1 to create
        - 1 to delete
        2 changes. 1 unchanged
    ```
[...]

Issue #2926, opened 2024-04-01
Fail-fast when a program attempts to create a namespace that already exists.


mjeffryes:
I was helping a user get started with an EKS project and they encountered an error on destroy that was confusing to them: `namespaces "kube-system" is forbidden: this namespace may not be deleted.`

What had happened was that in their program, they we creating a `Namespace` resource to manage the built in `kube-system` namespace like this: 
``` python
# This is the namespace where the AWS Load Balancer Controller will be installed.
namespace = kubernetes.core.v1.Namespace('aws-loadbalancer',
                                   metadata={
                                       'name': 'kube-system'  # typically installed in the kube-system namespace
                                   },
                                   opts=pulumi.ResourceOptions(provider=eks_provider))
```

This succeeded on create, probably as an SSA-mode “upsert” on create since the namespace already exists. However, since this is a built-in namespace, the destroy fails.

We should probably have failed fast on the initial `pulumi up`, warning the user that the namespace already exists and suggesting that they need to either import the resource if they want to manage it or just use a `NamespacePatch` if they just want to modify some properties of it.

EronWright:
I would advocate for fail-fast behavior with any kind, not just namespace.

mjeffryes:
As noted on #2948, the upsert behavior can also cause pulumi to silently delete a resource the user was intending to rename

EronWright:
Another variation: https://github.com/pulumi/pulumi-kubernetes/issues/2227#issuecomment-1482511794

EronWright:
To summarize, there's various ways to have two Pulumi resources pointing to the same Kubernetes object. Whether by changing the type of resource (e.g. `Chart` to `Release`) or the name of the resource (e.g. a change to `resourcePrefix`), a replacement resource is created, the old resource is deleted, and the object is then gone.

Normal resource replacement is not relevant here, because the before/after resources have a different URN.  That is, unless one applies an alias (and that would help with the name change only).

The problem happens specifically in SSA mode because the creation of the replacement doesn't fail as it would in CSA mode. #2926 suggests one possible solution, that is to fail-fast during create, but users do like the upsert feature (see [comment](https://github.com/pulumi/pulumi-kubernetes/issues/2948#issuecomment-2059929691)). Another solution based on reference counting is discussed in https://github.com/pulumi/pulumi/issues/15982.

PR #2999, opened 2024-05-07
Annotate resources with URNs to prevent accidental deletion

blampe:
### Proposed changes

User feedback [here](https://github.com/pulumi/pulumi-kubernetes/issues/2948) indicates that upsert behavior is actually desirable, and the concern is more along the lines of preventing accidental resource deletions. I agree with all the concerns -- while it's good to be aware of aliases and deleteBeforeReplace, the fact remains that our default behavior is still very foot-shooty.

> What if Pulumi wrote the pulumi resource name as pulumi.com/something annotation into the k8s resource. Upon deletion it checks if the pulumi resource name matches the value of this annotation, and only if that's the case it can go ahead and actually delete the resource.

I liked this proposal and wanted to see if it was feasible, because it could potentially provide a stricter/safer execution mode without requiring any additional effort from the user.

To quickly recap, #2948 describes a situation where a resource is renamed in Pulumi but unchanged in the cluster.
```diff
-const name1 = new k8s.apps.v1.Deployment("name1", { metadata: { name: "foo" }, ...})
+const name2 = new k8s.apps.v1.Deployment("name2", { metadata: { name: "foo" }, ...})
```

The engine plans this as create `default/foo` followed by delete `default/foo`, so Pulumi thinks it's created a new resource but the end result is actually a deletion. Yikes!!

The proposal is to annotate the object in the cluster with its URN and then only perform deletion when the URN matches what we expect to be deleting. This is not hard to implement, but there are a couple caveats:

* Resources shared by multiple stacks. I don't know how common this is, but we have to assume some people are using multiple stacks to operate on the same cluster resource(s). In those situations this proposal would refuse to delete a resource if it was previously manipulated by another stack. If we want that behavior to continue to work we can treat URNs belonging to different projects/stacks as valid deletions.
* Field managers. The create step currently performs an `Apply` that doesn't actually patch the resource, but it _does_ use a different field manager from what's already recorded in the cluster. With this proposal, the create step _would_ modify the resource to update its ARN annotation and this generates a SSA conflict. We could handle this a few ways:
  1. Surface the SSA conflict. This is the easiest but also the least user-friendly option, since they want the rename to "just work".
  2. Modify how we assign field managers. Instead of giving every resource its own unique field manager, make the field manager consistent for every physical resource in the stack. This allows us to upsert a new URN annotation without a conflict. Patch resources still need unique managers. The way this is implemented, existing field managers would be unchanged since we prefer to use whatever's already in state.

In any case, this is a big enough change that it would need to be opt-in -- what would we call the config to enable this?

Refs https://github.com/pulumi/pulumi-kubernetes/issues/2926
Refs https://github.com/pulumi/pulumi-kubernetes/issues/2948


blampe:
https://github.com/pulumi/pulumi-kubernetes/pull/2999

PR #3000, opened 2024-05-08
Expose config to disable upsert behavior

blampe:
This exposes opt-in provider configuration to disable upsert behavior. When this is set we will use CSA for creation and surface "already exists" errors when we attempt to overwrite something.

I called the config `enableUpsert` to be consistent with some of our other configs (although I'd prefer `disableUpsert` since that makes it more obvious what the default behavior is).

An E2E test is included and I'll try to add some units around this as well.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/2926.

### DIFF
--- a/sdk/dotnet/Config/Config.cs
+++ b/sdk/dotnet/Config/Config.cs
@@ -98,6 +98,16 @@ namespace Pulumi.Kubernetes
             set => _enableServerSideApply.Set(value);
         }
 
+        private static readonly __Value<bool?> _enableUpsert = new __Value<bool?>(() => __config.GetBoolean("enableUpsert") ?? Utilities.GetEnvBoolean("PULUMI_K8S_ENABLE_UPSERT") ?? true);
+        /// <summary>
+        /// If present and set to false, the provider will surface errors if a create operation would overwrite existing resources in the cluster.
+        /// </summary>
+        public static bool? EnableUpsert
+        {
+            get => _enableUpsert.Get();
+            set => _enableUpsert.Set(value);
+        }
+
         private static readonly __Value<string?> _kubeconfig = new __Value<string?>(() => __config.Get("kubeconfig"));
         /// <summary>
         /// The contents of a kubeconfig file or the path to a kubeconfig file. If this is set, this config will be used instead of $KUBECONFIG.
--- a/sdk/dotnet/Provider.cs
+++ b/sdk/dotnet/Provider.cs
@@ -83,6 +83,12 @@ namespace Pulumi.Kubernetes
         public Input<bool>? EnableServerSideApply { get; set; }
 
         /// <summary>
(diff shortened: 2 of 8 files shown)


provider/pkg/await/await_test.go
```diff
@@ -325,6 +325,7 @@ func Test_Creation(t *testing.T) {
 					DedupLogger:       logging.NewLogger(context.Background(), host, urn),
 					Resources:         resources,
 					ServerSideApply:   tt.args.serverSideApply,
+					EnableUpsert:      tt.args.serverSideApply,
```
EronWright, line 328:
```suggestion
					EnableUpsert:      tt.args.enableUpsert,
```

provider/pkg/gen/schema.go
```diff
@@ -90,6 +90,16 @@ func PulumiSchema(swagger map[string]any) pschema.PackageSpec {
 					Description: "BETA FEATURE - If present and set to true, allow ConfigMaps to be mutated.\nThis feature is in developer preview, and is disabled by default.\n\nThis config can be specified in the following ways using this precedence:\n1. This `enableConfigMapMutable` parameter.\n2. The `PULUMI_K8S_ENABLE_CONFIGMAP_MUTABLE` environment variable.",
 					TypeSpec:    pschema.TypeSpec{Type: "boolean"},
 				},
+				"enableUpsert": {
```
EronWright, line 93: Should we reverse the polarity of this field, so that it is false by default?

provider/pkg/gen/schema.go
```diff
@@ -177,6 +187,15 @@ func PulumiSchema(swagger map[string]any) pschema.PackageSpec {
...
+					Description: "If present and set to false, the provider will surface errors if a create operation would overwrite existing resources in the cluster.",
```
EronWright, line 191: Maybe should mention that this is for ssa mode only.

EronWright (commented):
In an earlier experiment along these lines (i.e. using Create rather than Apply), I observed that `metadata.managedFields` is significantly different.  I was concerned that some migration logic might be needed.
 
We have CSA-to-SSA logic elsewhere in the provider. Would this be involved on a subsequent update?

EronWright:
Let me state my opinion here, that Kubernetes SSA is designed for multi-party authoring experience (where multiple parties apply their intentions to a given object, but design-wise says nothing about object lifecycle. For example, it doesn't answer the question of "who should delete this object, my party or your party?". An auto-scaler, for example, would not expect to create or to delete a given object, merely to control the replicas field while the object exists.

Pulumi has features like "import" and "retainOnDelete" to control an object's lifecycle, and there's no good reason that SSA should complicate that, in my opinion.

EronWright:
An alternative way to describe this is "automatic import" - less technocratic.  When the object already exists, I think it is fair to say that the object is being imported.  In v4, the default value would be true, and in v5 it would assumedly be false.

PR #3010, opened 2024-05-16
Use upstream's MergeValues with assets

blampe:
Followup to my comment [here](https://github.com/pulumi/pulumi-kubernetes/pull/2947#discussion_r1588192130), this removes our custom `ValueOpts` in favor of upstream's `value.Options` to handle merging.

Upstream expects to work with files on disk, so we fetch any assets in our provided values and then write them to disk temporarily.

### DIFF
--- a/provider/pkg/helm/tool.go
+++ b/provider/pkg/helm/tool.go
@@ -35,6 +35,7 @@ import (
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/cli/values"
 	"helm.sh/helm/v3/pkg/downloader"
 	"helm.sh/helm/v3/pkg/getter"
 	"helm.sh/helm/v3/pkg/registry"
@@ -98,7 +99,7 @@ type TemplateOrInstallCommand struct {
 	Chart string
 
 	// Values to be applied to the chart.
-	Values ValueOpts
+	Values values.Options
 
 	tool         *Tool
 	actionConfig *action.Configuration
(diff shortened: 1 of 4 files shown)


provider/pkg/provider/helm/v4/values_test.go
```diff
@@ -121,7 +122,7 @@ image:
 			},
 			want: map[string]interface{}{
 				"configuration": map[string]any{
-					"backupStorageLocation": []map[string]any{},
+					"backupStorageLocation": []any{},
```
blampe, line 125: A subtle difference between the previous `MergeValues` and upstream's.

provider/pkg/helm/tool.go
```diff
@@ -130,17 +131,9 @@ func (cmd *TemplateOrInstallCommand) addInstallFlags() {
 	client.SubNotes = false
 	client.Labels = nil
 	client.EnableDNS = false
-	cmd.addValueOptionsFlags()
 	cmd.addChartPathOptionsFlags()
 }
 
-func (cmd *TemplateOrInstallCommand) addValueOptionsFlags() {
```
EronWright, line 137: I would still keep `addValueOptionsFlags` to set defaults on `Values` and to further resemble:
https://github.com/helm/helm/blob/14d0c13e9eefff5b4a1b511cf50643529692ec94/cmd/helm/flags.go#L45-L51

provider/pkg/provider/helm/v4/values.go
```diff
@@ -93,14 +113,14 @@ func readAsset(p getter.Providers, asset pulumi.Asset) ([]byte, error) {
 	}
 }
 
-// marshalValues converts Pulumi values to Helm values.
-// - Expands assets to their content (to support --set-file).
-func marshalValues(p getter.Providers, a map[string]interface{}) (map[string]interface{}, error) {
+// readAssets converts Pulumi values to Helm values, hydrating Asset values
+// along the way.
+func readAssets(p getter.Providers, a map[string]interface{}) (map[string]interface{}, error) {
```
EronWright, line 118: I feel like `marshalValues` is a better name because it isn't just dealing with assets; imagine that resource references were marshaled in some way.

provider/pkg/provider/helm/v4/values_test.go
```diff
@@ -143,30 +144,29 @@ image:
 	for _, tt := range tests {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
-			merger := &ValueOpts{
-				ValuesFiles: tt.valuesFiles,
-				Values:      tt.values,
-			}
+			p := getter.All(cli.New())
+			opts, cleanup, err := readValues(p, tt.values, tt.valuesFiles)
```
EronWright, line 148: Can you use `getter.Providers{}` for `p`?

EronWright (commented):
Nice!

EronWright:
On one hand, this approach allows us to leverage more of the Helm library, specifically related to loading values files.  On the other hand, the need to write out the adhoc values to a file seems a bit contrived. Let's vote on it, in description.

EronWright:
@blampe would you agree to closing this now?

PR #3011, opened 2024-05-16
new annotation: `deletionPropagationPolicy`

### LINKED ISSUE #1831: Add ability to delete a statefulset with the cascade orphan option
## Hello!

## Issue details

For certain statefulset update situations, pulumi needs to delete the statefulset and then create it again. One such case would be when modifying the storage size of a volume in the claim template. The issue here is that all of the statefulset pods will be terminated at once, and hence there won't be any pods to receive traffic. 

K8s provides a `cascade` option which allows it to delete the statefulset without it deleting the dependent pods. https://kubernetes.io/docs/tasks/run-application/delete-stateful-set/

Currently I need to run manually kubectl with the cascade flag, pulumi refresh, and then pulumi up. Providing a way to configure the cascade option in code would allow Pulumi to delete the statefulset while keeping the pods, and then recreate it with the new changes.

Perhaps setting a new annotation on the statefulset would be the way to solve this?

This request has some relation to: https://github.com/pulumi/pulumi-kubernetes/issues/304

EronWright:
### Proposed changes

Introduces a new Pulumi annotation to set the [deletion propagation policy](https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/), e.g. to support non-cascading delete on `StatefulSet` to preserve pods during replacement (see [walkthrough](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#non-cascading-delete)). 

Note that the policy annotation must be set on the old resource before deleting or replacing it; setting it on the replacement or on the live object is ineffective.

### Related issues (optional)

Closes #1831 

### Example

This example serves to show how the 'orphan' propagation policy allows for non-disruptive replacement of
a StatefulSet, e.g. by touching the `volumeClaimTemplates`.

```yaml
name: issue-1831
runtime: yaml
description: A StatefulSet to demonstrate the `pulumi.com/deletionPropagationPolicy` annotation.
config:
  # disable SSA for this demonstration
  kubernetes:enableServerSideApply: false
resources:
  nginx:
    type: kubernetes:apps/v1:StatefulSet
    properties:
      metadata:
        name: nginx
        annotations:
          pulumi.com/deletionPropagationPolicy: "orphan"
      spec:
        replicas: 1
        selector:
          matchLabels:
            app: nginx
        serviceName: nginx
        template:
          metadata:
            labels:
              app: nginx
          spec:
            containers:
            - image: nginx:1.19.9
              name: nginx
              ports:
              - containerPort: 80
                name: web
        volumeClaimTemplates:
        - metadata:
            name: nginx
          spec:
            accessModes:
            - ReadWriteOnce
            resources:
              requests:
                storage: 1Gi
```

Following the initial deployment, we have these objects:
```
❯ kubectl get statefulset,pod,pvc -ocustom-columns='KIND:.kind,NAME:.metadata.name,UID:.metadata.uid'
KIND                    NAME            UID
StatefulSet             nginx           b1aa144d-3f16-448e-8e15-02d2c2d4b61a
Pod                     nginx-0         c00d97cc-39d7-4a95-b839-0910f911dbca
PersistentVolumeClaim   nginx-nginx-0   2c624ff9-e856-4d2d-bfaf-527c6b770bc7
```

To provoke a replacement, we change the PVC template:
```
              requests:
-                storage: 1Gi
+                storage: 2Gi
```
Let's also increase the replicas:
```
      spec:
-        replicas: 1
+        replicas: 2
```

And deploy:
```
❯ pulumi up -f
Updating (dev)

     Type                               Name            Status            Info
     pulumi:pulumi:Stack                issue-1831-dev                    4 warnings; 2 messages
 +-  └─ kubernetes:apps/v1:StatefulSet  nginx           replaced (2s)     [diff: ~spec]

Resources:
    +-1 replaced
    1 unchanged
```

Looking again at the objects:
```
❯ kubectl get statefulset,pod,pvc -ocustom-columns='KIND:.kind,NAME:.metadata.name,UID:.metadata.uid'
KIND                    NAME            UID
StatefulSet             nginx           135d9142-460c-4f64-82a6-8ce23427f52b
Pod                     nginx-0         c00d97cc-39d7-4a95-b839-0910f911dbca
Pod                     nginx-1         8c80932f-7051-4fc7-baeb-00dae4b07b64
PersistentVolumeClaim   nginx-nginx-0   2c624ff9-e856-4d2d-bfaf-527c6b770bc7
PersistentVolumeClaim   nginx-nginx-1   e4b4fd18-28b2-454b-8c6b-9fa06435d3d6
```

We see the expected result: the StatefulSet was replaced, the existing pod was adopted, and a new pod was added w/ a PVC. 

In more detail, the StatefulSet controller uses the selector to identify existing pods, then chooses to delete or adopt based on suitability.  Note the updated owner reference on `nginx-0`:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-0
  uid: c00d97cc-39d7-4a95-b839-0910f911dbca
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: StatefulSet
    name: nginx
    uid: 135d9142-460c-4f64-82a6-8ce23427f52b
```

To demonstrate how the StatefulSet controller might choose to delete the existing pod rather than adopting it, let's change the image rather than the replicas:
```
            containers:
-            - image: nginx:1.19.9
+            - image: nginx:1.19.10
 
              requests:
-                storage: 2Gi
+                storage: 3Gi
```

We deploy again and see that all pods were replaced.
```
KIND                    NAME            UID
StatefulSet             nginx           ead53943-abc9-47c8-9393-326f845c7f42
Pod                     nginx-0         74752b8c-3979-478b-9be4-ff3ca1b0aa6f
Pod                     nginx-1         b6c2f0f6-f5ff-4e04-a1da-66966b8d697c
PersistentVolumeClaim   nginx-nginx-0   2c624ff9-e856-4d2d-bfaf-527c6b770bc7
PersistentVolumeClaim   nginx-nginx-1   e4b4fd18-28b2-454b-8c6b-9fa06435d3d6
```

Note that PVC `nginx-nginx-0` was not replaced and still has `storage: 1Gi`.

### DIFF
--- a/provider/pkg/metadata/annotations.go
+++ b/provider/pkg/metadata/annotations.go
@@ -35,7 +35,7 @@ const (
 	AnnotationPatchForce        = AnnotationPrefix + "patchForce"
 	AnnotationPatchFieldManager = AnnotationPrefix + "patchFieldManager"
 
-	AnnotationDeletionPropagation = AnnotationPrefix + "deletionPropagation"
+	AnnotationDeletionPropagation = AnnotationPrefix + "deletionPropagationPolicy"
 
 	AnnotationHelmHook = "helm.sh/hook"
 )
--- a/provider/pkg/metadata/annotations.go
+++ b/provider/pkg/metadata/annotations.go
@@ -35,6 +35,8 @@ const (
 	AnnotationPatchForce        = AnnotationPrefix + "patchForce"
 	AnnotationPatchFieldManager = AnnotationPrefix + "patchFieldManager"
 
+	AnnotationDeletionPropagation = AnnotationPrefix + "deletionPropagation"
+
 	AnnotationHelmHook = "helm.sh/hook"
 )
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fix a panic that could occur when a missing field became `null`. (https://github.com/pulumi/pulumi-kubernetes/issues/1970)
 - Add field manager's name to server-side apply conflict errors. (https://github.com/pulumi/pulumi-kubernetes/pull/2983)
 - Helm Chart V4 (https://github.com/pulumi/pulumi-kubernetes/pull/2947)
+- New annotation: deletionPropagationPolicy (https://github.com/pulumi/pulumi-kubernetes/pull/3011)
 
 ## 4.11.0 (April 17, 2024)
 
--- a/provider/pkg/await/await.go
+++ b/provider/pkg/await/await.go
@@ -790,7 +790,12 @@ func Deletion(c DeleteConfig) error {
 		return nilIfGVKDeleted(err)
 	}
 
-	err = deleteResource(c.Context, c.Name, client)
+	// delete the specified resource (using foreground cascading delete by default).
+	deletePolicy := metadata.DeletionPropagation(c.Inputs)
+	deleteOpts := metav1.DeleteOptions{
+		PropagationPolicy: &deletePolicy,
+	}
+	err = client.Delete(c.Context, c.Name, deleteOpts)
 	if err != nil {
 		return nilIfGVKDeleted(err)
 	}
@@ -878,16 +883,6 @@ func Deletion(c DeleteConfig) error {
 	}
 
 	return nil
-}
-
-// deleteResource deletes the specified resource using foreground cascading delete.
-func deleteResource(ctx context.Context, name string, client dynamic.ResourceInterface) error {
-	fg := metav1.DeletePropagationForeground
-	deleteOpts := metav1.DeleteOptions{
-		PropagationPolicy: &fg,
-	}
-
-	return client.Delete(ctx, name, deleteOpts)
 }
 
 // checkIfResourceDeleted attempts to get a k8s resource, and returns true if the resource is not found (was deleted).
--- a/provider/pkg/metadata/overrides.go
+++ b/provider/pkg/metadata/overrides.go
@@ -16,8 +16,10 @@ package metadata
 
 import (
 	"strconv"
+	"strings"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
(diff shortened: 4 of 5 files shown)


provider/pkg/metadata/annotations.go
```diff
@@ -35,6 +35,8 @@ const (
 	AnnotationPatchForce        = AnnotationPrefix + "patchForce"
 	AnnotationPatchFieldManager = AnnotationPrefix + "patchFieldManager"
 
+	AnnotationDeletionPropagation = AnnotationPrefix + "deletionPropagation"
```
blampe, line 38: What about `deleteCascade` to better match the k8s terminology?

EronWright: My rationale was to match the terminology in [apimachinery](https://github.com/kubernetes/apimachinery/blob/1da46c3f5a5b4a0cc756cb6050df0cf6f06b64c2/pkg/apis/meta/v1/types.go#L479-L495) but you're right that `kubectl` uses `--cascade`. I like your suggestion too and will change it unless others have stronger opinions.

```
// DeletionPropagation decides if a deletion will propagate to the dependents of
// the object, and how the garbage collector will handle the propagation.
type DeletionPropagation string
```

rquitales: I am strongly aligned with using "propagationPolicy," which is the terminology used in the API server and embraced by other GitOps tools such as [ArgoCD](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#resources-prune-deletion-propagation-policy) and [ConfigSync](https://cloud.google.com/anthos-config-management/docs/how-to/managing-objects#deletion_propagation_options).

Furthermore, "orphan" isn't really a cascading deletion type, reinforcing the preference for "propagationPolicy." When kubectl introduced the `cascade` flag, its purpose was to toggle between orphan (`cascade=false`) and background cascading delete (`cascade=true`), as indicated in this upstream issue: https://github.com/kubernetes/kubernetes/issues/93364. I suggest adopting `propagationPolicy` to also accommodate potential non-cascading strategies in future Kubernetes versions.

EronWright: OK let's go with `deletionPropagationPolicy`.

blampe (approved):
Love it!

Issue #3022, opened 2024-05-23
Job with unreachable cluster causes a panic


sinhalvi:
### What happened?

Hi, I suddenly see this error when I do a pulumi up on a stack
```
panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x2daee4d]
    goroutine 103 [running]:
    github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/clients.(*DynamicClientSet).ResourceClient(0xc0005f3300, {{0xc000c2ddd8, 0x5}, {0xc000c2ddde, 0x2}, {0xc000c2ddf4, 0x3}}, {0xc000a96068, 0x4})
        /home/runner/work/pulumi-kubernetes/pulumi-kubernetes/provider/pkg/clients/clients.go:95 +0x10d
    github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/clients.(*DynamicClientSet).ResourceClientForObject(0xc0005f3300, 0xc00087fd40)
        /home/runner/work/pulumi-kubernetes/pulumi-kubernetes/provider/pkg/clients/clients.go:118 +0x11f
    github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/provider.(*kubeProvider).readLiveObject(0xc000628000, 0xc00087fd40)
        /home/runner/work/pulumi-kubernetes/pulumi-kubernetes/provider/pkg/provider/provider.go:2680 +0xf8
    github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/provider.(*kubeProvider).Diff(0xc000628000, {0x59c0b50, 0xc0009902a0}, 0xc000581400)
        /home/runner/work/pulumi-kubernetes/pulumi-kubernetes/provider/pkg/provider/provider.go:1749 +0x13fe
    github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Diff_Handler.func1({0x59c0b50?, 0xc0009902a0?}, {0x5364c60?, 0xc000581400?})
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.114.0/proto/go/provider_grpc.pb.go:575 +0xcb
[...]

PR #3024, opened 2024-05-24
Fix panic with Job resource diffing against an unreachable cluster

rquitales:
### Proposed changes
This PR ensures that we do not make a k8s API request during the provider's diff if there is an unreachable cluster. This currently occurs when the Pulumi program contains a Job resource with the `replaceUnready` annotation set to true. A panic would occur if we attempt to make the API call since our clients are nil.

#### Testing done:
1. Created a repro test case that fails with a panic (https://github.com/pulumi/pulumi-kubernetes/actions/runs/9228447658/job/25392833842?pr=3024)
2. Added logic to prevent the panic, and test passes subsequently without intervention (https://github.com/pulumi/pulumi-kubernetes/actions/runs/9228685506/job/25393667599?pr=3024)
3. Manual validation to ensure panic isn't trigerred.

### Related issues (optional)

Fixes: #3022

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Update to pulumi-java v0.12.0 #3025 (https://github.com/pulumi/pulumi-kubernetes/pull/3025)
+- Fixed a panic that occurs when diffing Job resources containing `replaceUnready` annotations and an unreachable cluster connection. (https://github.com/pulumi/pulumi-kubernetes/pull/3024)
 
 ## 4.12.0 (May 21, 2024)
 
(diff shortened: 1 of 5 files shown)


EronWright (approved):
LGTM

blampe (approved):
Makes me wonder if we should also check if the cluster is unreachable inside `readLiveObject`, seems like an easy mistake to make otherwise.

rquitales:
> Makes me wonder if we should also check if the cluster is unreachable inside `readLiveObject`, seems like an easy mistake to make otherwise.

Sonuds good! Added an assert as a precheck for `readLiveObject` in https://github.com/pulumi/pulumi-kubernetes/pull/3024/files#diff-418e6e7d93aeacebe7c22d5fb674b48bfa15c6a514eb3ae4744b86d447595d8eR2684.

PR #3036, opened 2024-05-30
Develop Component resource for: kubernetes:kustomize:Directory

### LINKED ISSUE #2786: Develop Component resource for: kubernetes:kustomize:Directory
```[tasklist]
### Resolves
- [ ] https://github.com/pulumi/pulumi-kubernetes/issues/1966
- [ ] https://github.com/pulumi/pulumi-kubernetes/issues/1884
```

```[tasklist]
### Tasks
- [ ] docs for Kustomize MLC
- [x] Kustomize MLC
```

EronWright:
## Proposed changes

This PR implements the Kustomize `Directory` resource (v2) as a multi-language component resource.

The `Directory` resource creates child resources based on a kustomization directory. The input is a path to a directory containing 'kustomization.yaml', or a git repository URL with a path suffix and other qualifiers.

### Related issues (optional)

Closes #2786 

### API

|property|description|
|-|-| 
| `directory` | The directory or git URL containing the kustomization to apply. |
| `namespace` | Overrides the default namespace. |
| `skipAwait` | Applies the skipAwait annotation. |
| `resourcePrefix` | Prefix for child resources, defaults to the component name. |

### Feature: Remote Targets

This implementation supports remote kustomization targets as described [here](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md).

Note: the v1 implementation also supports git references, but resolves them using the Pulumi SDK ([`RetrieveGitFolder`](https://github.com/pulumi/pulumi/blob/v3.117.0/sdk/go/common/workspace/templates.go#L392)) rather than using Kustomize's own implementation.

### Feature: Enable Alpha Plugins

This implementation always enables Kustomize's plugin support, akin to `kustomize build --enable-alpha-plugins`.

Plugins are drawn from `KUSTOMIZE_PLUGIN_HOME` (default: `~/.config/kustomize/plugin/`). Note: the kustomize library doesn't allow for easy customization of the plugin home.

### Feature: Unrestricted Loading

Kustomize has a strict and a relaxed mode with respect to path references outside the kustomization base directory (`--load-restrictor`, default is `strict`). The feature seems intended to encourage portability, similarly to paths in Dockerfiles. This implementation simply enables the relaxed mode.

### Feature: Enable Helm Charts

This implementation enables [Helm chart support in Kustomize](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/helmcharts/), which is an experimental feature of Directory v1 (see: https://github.com/pulumi/pulumi-kubernetes/issues/2470). Note that chart support in Kustomize is limited and evolving; see https://github.com/kubernetes-sigs/kustomize/issues/4401 for the long-term support plan.

The `helm` binary is assumed to be on the path.

### Feature: Namespace Override

New to v2 is support for overriding the default namespace (default is from provider configuration), as a convenience. Kustomize itself has a similar facility (see: [namespace](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/namespace/)) but it makes sense to support it natively.

### Feature: Resource Ordering

The Directory resource uses Pulumi's engine to install the resources produced by kustomize. It automatically detects dependencies between some resources. For example, it knows to install namespaces and Custom Resource Definitions (CRDs) first.

Use the `config.kubernetes.io/depends-on` annotation to declare an explicit resource dependency. See [blog post](https://www.pulumi.com/blog/kubernetes-yaml-v2/#resource-ordering) for more info.

### Limitation: Kubernetes-Style Transformations

The older v1 implementation provides the ability to transform the Kubernetes objects produced by kustomize before being sent to the server. That feature isn't available in Directory v2 at this time. You may still use Pulumi's [transform option](https://www.pulumi.com/docs/concepts/options/transformations/) to modify the child resources.

### Tests

The PR includes unit tests covering the provider implementation code.

Also includes an integration test that is similar to the below example.

### Example

Here's an example of deploying a couple of kustomizations. The `local` kustomization (see: [helloWorld](https://github.com/kubernetes-sigs/kustomize/tree/master/examples/helloWorld) for details) shows how to use a variable to select a kustomization overlay.

```yaml
name: issue-2786-yaml
runtime: yaml
description: |
  Demonstrates the Directory resource.
  Reference: https://github.com/kubernetes-sigs/kustomize/tree/master/examples/helloWorld
config:
  variant:
    type: string
    default: staging
outputs:
  name: ${local.resources}
resources:
  ns:
    type: kubernetes:core/v1:Namespace
  local:
    type: kubernetes:kustomize/v2:Directory
    properties:
      namespace: ${ns.metadata.name}
      directory: ./config/${variant}
      skipAwait: true
  remote:
    type: kubernetes:kustomize/v2:Directory
    properties:
      namespace: ${ns.metadata.name}
      directory: https://github.com/kubernetes-sigs/kustomize//examples/helloWorld/?ref=v3.3.1
      skipAwait: true
```


provider/pkg/gen/overlays.go
```diff
@@ -1230,6 +1230,55 @@ var kustomizeDirectoryResource = pschema.ResourceSpec{
...
+			TypeSpec: pschema.TypeSpec{
+				Type: "string",
+			},
+			Description: "The directory containing the kustomization to apply. The value can be a local directory or a folder in a\ngit repository.\nExample: ./helloWorld\nExample: https://github.com/kubernetes-sigs/kustomize/tree/master/examples/helloWorld",
+		},
+		"namespace": {
+			TypeSpec: pschema.TypeSpec{
+				Type: "string",
+			},
+			Description: "The default namespace. Defaults to the provider's namespace.",
```
rquitales, line 1262: `The default namespace` may not be super clear. Perhaps something like the namespace to apply manifests?

provider/pkg/provider/kustomize/v2/directory_test.go
```diff
@@ -0,0 +1,238 @@
...
+	resmap kresmap.ResMap
+	err    error
+}
+
+var _ kustomizer = &fakeKustomizer{}
+
+func (k *fakeKustomizer) Run(fSys kfilesys.FileSystem, path string) (kresmap.ResMap, error) {
+	return k.resmap, k.err
+
+	// var rmF = resmap.NewFactory(rf)
```
rquitales, line 49: Let's remove these commented out code.

rquitales:
Second pass of the PR looks good. I'd prefer if we could also include a integration test flow to ensure that we can actually create these resources.

rquitales (approved):

PR #3046, opened 2024-06-04
Fix: Chart v4 fails on update

### LINKED ISSUE #3045: Chart v4 fails on update
### What happened?

I tried to use `Chart` v4 to install cert-manager and saw an unexpected error during a subsequent update.

```
error: an unhandled error occurred: waiting for RPCs: rpc error: code = Unknown desc = Unable to continue with install: NetworkPolicy "certman-oci" in namespace "cert-manager" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "certman-oci"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "cert-manager"
```

### Example

```yaml
name: example
runtime: yaml
resources:
  certman-ns:
    type: kubernetes:core/v1:Namespace
    properties:
      metadata:
        name: cert-manager
  certman-oci:
    type: kubernetes:helm.sh/v4:Chart
    properties:
      namespace: ${certman-ns.metadata.name}
      chart: oci://registry-1.docker.io/bitnamicharts/cert-manager
      version: "1.1.0"
```

### Output of `pulumi about`

```
CLI          
Version      3.117.0
Go Version   go1.22.3
Go Compiler  gc

Plugins
KIND      NAME        VERSION
resource  kubernetes  unknown
language  yaml        unknown

Host     
OS       darwin
Version  14.4.1
Arch     arm64
```

### Additional context

_No response_

### Contributing

EronWright:
### Proposed changes

This PR fixes a problem with how Chart v4 uses the Helm library. The design goal is to allow for connectivity during template rendering, to support the lookup function (see https://github.com/helm/helm/pull/9426) and to provide an accurate [Capabilities object](https://helm.sh/docs/chart_template_guide/builtin_objects/). Unfortunately we were slightly too aggressive and caused some of Helm's "non-template" code to execute.

This fix works by turning off the `helm template --validation` flag, so that the internal `ClientOnly` flag is true thus avoiding [this block of code](https://github.com/helm/helm/blob/6f32a8f9d338bacc3c6a1c0c3610012b01edb3d1/pkg/action/install.go#L345-L350) that causes the unexpected error. A side-effect of `ClientOnly` being true is that the capabilities aren't automatically set, and so we set them using the provider's kube client (akin to using `--kube-version`).

Detailed changes:
- (chart.go) use ClientOnly mode, set KubeVersion and APIVersions
- (tool.go) remove redundant KubeVersion and ExtraAPIs 
- (testdata/reference) add a version check
- (chart_test.go) unit tests for `.Capabilities`
- integration test to install cert-manager

### Related issues (optional)

Closes #3045

### DIFF
--- a/provider/pkg/helm/tool.go
+++ b/provider/pkg/helm/tool.go
@@ -33,7 +33,6 @@ import (
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
-	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/downloader"
 	"helm.sh/helm/v3/pkg/getter"
@@ -167,8 +166,6 @@ type TemplateCommand struct {
 	Validate    bool
 	IncludeCRDs bool
 	SkipTests   bool
-	KubeVersion string
-	ExtraAPIs   []string
 }
 
 // Template returns a new `helm template` command.
@@ -191,8 +188,6 @@ func (t *Tool) Template() *TemplateCommand {
 	cmd.IncludeCRDs = false
 	cmd.SkipTests = false
 	cmd.Install.IsUpgrade = false
-	cmd.KubeVersion = ""
-	cmd.ExtraAPIs = []string{}
 	cmd.Install.UseReleaseName = false
 	return cmd
 }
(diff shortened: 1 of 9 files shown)


provider/pkg/helm/tool.go
```diff
@@ -167,8 +166,6 @@ type TemplateCommand struct {
 	Validate    bool
 	IncludeCRDs bool
 	SkipTests   bool
-	KubeVersion string
-	ExtraAPIs   []string
```
EronWright, line 171: These two properties were unnecessary wrappers over properties in the embedded type (`action.Install`), and made it harder than necessary to setup the capabilities.

rquitales (commented):
First pass on this PR looks reasonable. 👍

provider/pkg/provider/helm/v4/chart.go
```diff
@@ -173,10 +177,20 @@ func (r *ChartProvider) Construct(ctx *pulumi.Context, typ, name string, inputs
 	p := tool.AllGetters()
 	cmd := tool.Template()
 
-	cmd.Validate = true
+	// connectivity: use server-side dry-run to enable the lookup function.
+	// i.e. `helm template --dry-run=server --validate=false`.
+	cmd.DisableOpenAPIValidation = true
+	cmd.Validate = false
+	cmd.ClientOnly = true
```
EronWright, line 184: Emulates:
```
helm template --dry-run=server --validate=false --disable-openapi-validation \
  --kube-version $VERSION --api-versions $APIVERSIONS
```

rquitales (approved):

tests/sdk/java/chartv4_test.go
```diff
@@ -0,0 +1,20 @@
...
+// TestChartV4 deploys a complex stack using chart/v4 package.
+func TestChartv4(t *testing.T) {
+	test := pulumitest.NewPulumiTest(t, "testdata/chartv4")
+	t.Logf("into %s", test.Source())
+	t.Cleanup(func() {
+		test.Destroy()
+	})
+	test.Preview()
+	test.Up()
+	test.Up(optup.ExpectNoChanges())
```
rshade, line 19: ❤️

Issue #2575, opened 2023-09-21
replaceOnChange doesn't work for immutable Fields


bernhardloos:
### What happened?

Making changes to the storage class field of a PersistentVolumeClaim fails with an error during preview:
```
error: Preview failed: resource ... was not successfully created by the Kubernetes API server : PersistentVolumeClaim "..." is invalid: spec: Forbidden: spec is immutable after creation except resources.requests for bound claims
      core.PersistentVolumeClaimSpec{
        ... // 2 identical fields
        Resources:        {Requests: {s"storage": {i: {...}, s: "1", Format: "DecimalSI"}}},
        VolumeName:       "...",
    -   StorageClassName: &"no-provisioner",
    +   StorageClassName: &"no-provisionerf",
        VolumeMode:       &"Filesystem",
        DataSource:       nil,
        DataSourceRef:    nil,
      }
```

Setting replaceOnChange doesn't seem to have any effect, no matter what I put there. I suspect the provider sends the data to the API server for server side apply before the replaceOnChange option is processed. replaceOnChange works if use it for example for annotations.

### Example

any PVC should exhibit the behavior: https://www.pulumi.com/registry/packages/kubernetes/api-docs/core/v1/persistentvolumeclaim/

### Output of `pulumi about`

```
CLI          
Version      3.81.0
Go Version   go1.21.0
Go Compiler  gc

Plugins
NAME             VERSION
kubernetes       4.1.1
kubernetes_crds  4.1.1
python           unknown
random           4.13.2
tls              4.10.0
vault            5.15.0

Host     
OS       fedora
Version  38
Arch     x86_64

This project is written in python: executable='/usr/bin/python3' version='3.11.4


Found no pending operations associated with main

Backend        
Name           lap
URL            file://.
User           bl
Organizations  

Pulumi locates its logs in /tmp by default
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Hi @bernhardloos, thanks for reporting this issue, and apologies you're facing it. Your hypothesis is on the right track. If there is a diff detected, Pulumi attempts to also do a server side apply dry run. Since an update on immutable fields is prohibited, the dry run apply will fail which results in what you observed. I'll dig into what we can do to make this experience better, but I believe the fix will probably require us to NOT do a dry-run if `replaceOnChange` is specified.

[...]

PR #3053, opened 2024-06-10
Fix previews in SSA mode when there are replacements due to immutable fields

rquitales:
### Proposed changes

This PR adds special handling for previews with SSA when we have a replacement (delete before create) operation. There is no easy way to tell the API server that the dry-run to be run is an object replacement. As such, a workaround that we can do is to send a different object name so the API server thinks we are creating a new object, so a already exists type error doesn't occur.

Note, because of how SSA creates work currently with upserts, we'll never face the `AlreadyExists` error, so we do not have to account for that in this PR.

As an aside, `kubectl` also supports delete before create replacements with the `replace` sub-command. A user would need to run `kubectl replace --force -f .` to recreate any resources that update an immutable field. `kubectl` does not support `-dry-run=server` with the `--force` flag and the follow error is shown `--dry-run can not be used when --force is set` since the API server does not support previewing this type of operation.

### Related issues (optional)

Fixes: #2575

### DIFF
--- a/provider/pkg/await/await.go
+++ b/provider/pkg/await/await.go
@@ -46,6 +46,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/client-go/dynamic"
 	k8sopenapi "k8s.io/kubectl/pkg/util/openapi"
 	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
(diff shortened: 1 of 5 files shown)


provider/pkg/await/await.go
```diff
@@ -213,6 +213,25 @@ func Creation(c CreateConfig) (*unstructured.Unstructured, error) {
...
 
+				// Handle the preview scenario where we need to re-create the object due to immutable fields.
+				// To avoid the immutable field error reported by the API server, we append "-preview" to the name
+				// so that the dry-run operation can succeed. This is a workaround for the limitation of the
+				// server-side apply API, which does not support replacement dry-run operations.
+				//
+				// Note that the AlreadyExists error does not occur since we always send a PATCH request during
+				// creates with SSA enabled. This means that we currently always upsert the object.
+				if c.Preview &&
+					(apierrors.IsInvalid(err) && strings.Contains(err.Error(), "field is immutable")) {
```
blampe, line 224: You could use `apierrors.FieldImmutableErrorMsg` here but it doesn't really matter.

blampe (approved):
Reasonable! It's too bad we can't know upfront whether we're about to replace or update. Another consequence of the engine not expecting upserts.

provider/pkg/await/await.go
```diff
@@ -213,6 +214,37 @@ func Creation(c CreateConfig) (*unstructured.Unstructured, error) {
...
+				// server-side apply API, which does not support replacement dry-run operations.
+				//
+				// Note that the AlreadyExists error does not occur since we always send a PATCH request during
+				// creates with SSA enabled. This means that we currently always upsert the object.
+				if c.Preview &&
+					(apierrors.IsInvalid(err) && strings.Contains(err.Error(), apivalidation.FieldImmutableErrorMsg)) {
+					const previewSuffix = "-preview"
+					objName := c.Inputs.GetName()
+
+					// Set the name to have the "-preview" suffix. Ensure we don't have a string over 253 characters, which is
```
rquitales, line 229: fyi: added this chunk in the last commit.

provider/pkg/await/await.go
```diff
@@ -213,6 +214,37 @@ func Creation(c CreateConfig) (*unstructured.Unstructured, error) {
...
+					// the maximum length for a name in Kubernetes.
+					var previewName string
+					if len(objName) > 253-len(previewSuffix) {
+						previewName = objName[:253-len(previewSuffix)] + previewSuffix
```
EronWright, line 233: Should the string be converted to runes before truncating it?  See https://stackoverflow.com/a/74700627/3026310

EronWright (approved):
I like this because it is targeted. Down the road I would advocate for simply using client-side validation during previews, to avoid the problems inherent to server-side dry-run such as missing CRDs, missing namespaces, missing admission webhooks, inability to preview a replacement, etc.

provider/pkg/await/await.go
rquitales: Updated to use [upstream K8's name generator](https://github.com/jpbetz/kubernetes/blob/4b5ee802e847a3ea8243946aeae588906bd6f79f/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L49-L54) instead.

rquitales:
Update changelog status is frozen. Force merging - we do have a changelog entry here.

Issue #2745, opened 2024-01-10
Alternative way of detecting cluster changes instead of `kubeconfig`


eplightning:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently the provider relies on passed `kubeconfig` input to decide if it needs replacement or not.
This is not always correct and has a chance for both false negatives and false positives, resulting either in potentially destructive replacement of resources (volumes, ArgoCD applications...) or errors when an expected resource doesn't exist.

Therefore I'd like to request an alternative to doing that for users that have this information by other means (for example ID of a managed K8s cluster).

Simplest way I can think of to achieve this would be adding a new optional provider config property `clusterId` and changing the detection logic as follows:
1. If `clusterId` is passed, report provider for replacement whenever new `clusterId` is computed or differs from previous `clusterId`
2. Else use current kubeconfig logic

Then this could be used this way in Pulumi code:
```typescript
const aks = new containerservice.KubernetesCluster('cluster', ...);
const k8sProvider = new kubernetes.Provider('cluster', {
  clusterId: aks.id,
  kubeconfig: aks.kubeAdminConfigRaw
});
```

### Affected area/feature

Provider

EronWright:
Another alternative may be to use some server information, such as the UID of the default namespace.  I don't know whether the provider resource could do the detection, and vary its URN (identity) accordingly.  Or whether each object resource would do the detection and state tracking.

EronWright:
I think we should prioritize a simple fix that gives the user control, like a `clusterid` property that causes replacement whenever it changes but has no other significance, and fall back to existing behavior if `clusterid` is not set.

blampe:
> I think we should prioritize a simple fix that gives the user control, like a `clusterid` property that causes replacement whenever it changes but has no other significance, and fall back to existing behavior if `clusterid` is not set.

Agree!

Roughly speaking, I think we want `DiffConfig` to have behavior like:

* Setting or unsetting a clusterid on an existing provider → update the provider.
* Other config changes if clusterid is set and unchanged, or being set/unset → update the provider.
* Changing a non-null clusterid to a different non-null clusterid → replace the provider.

Otherwise fall back to existing `DiffConfig` logic when clusterid is unset.

[...]

PR #3068, opened 2024-06-24
Add `clusterIdentifier` configuration to control provider lifecycle

blampe:
This adds a new `clusterIdentifier` provider configuration to control replacement behavior.

If this config has been previously set, then we will only replace the provider when `clusterIdentifier` changes. Additionally, changes to other properties like `kubeconfig` which currently trigger replacements will now only be updates.

An E2E test is included along with unit tests. The units use [txtar](https://pkg.go.dev/golang.org/x/tools/txtar) to more conveniently express before/after state in YAML -- a sort of middle ground between massive table tests and full-blown integration tests.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/2745.


EronWright (commented):
Looks good to me, have 2 points:

1. there's already some tests for DiffConfig in `provider_diffconfig_test.go`; can we add a section to outline the behavior of the clusteridentifer?
2. The term `clusterIdentifier` seems a bit long and imprecise, I would advocate for `clusterKey`.

EronWright (approved):
LGTM but we should do some manual testing for good measure.

rquitales (approved):

PR #3070, opened 2024-06-25
Enable TypedDict input types for the Python SDK

julienp:
### Proposed changes

This PR requires pulumi/pulumi 3.121.0 and is stacked on top of https://github.com/pulumi/pulumi-kubernetes/pull/3066.

Enable TypedDict input types for the Python SDK. The only manual change is to set `inputTypes` flag in the python language features of the schema. This is done here https://github.com/pulumi/pulumi-kubernetes/commit/20314f5451f9fc7aec4fc8a1e0d649398159d3ec. This flag is currently opt in, but will default to `classes-and-dicts` in the future.

I ran `make build` afterwards.

### Related issues (optional)

Epic https://github.com/pulumi/pulumi/issues/12689

### DIFF
--- a/sdk/dotnet/Pulumi.Kubernetes.csproj
+++ b/sdk/dotnet/Pulumi.Kubernetes.csproj
@@ -9,7 +9,7 @@
     <PackageProjectUrl>https://pulumi.com</PackageProjectUrl>
     <RepositoryUrl>https://github.com/pulumi/pulumi-kubernetes</RepositoryUrl>
     <PackageIcon>logo.png</PackageIcon>
-    <Version>4.14.0-alpha.1718665781</Version>
+    <Version>4.0.0-alpha.0+dev</Version>
 
     <TargetFramework>net6.0</TargetFramework>
     <Nullable>enable</Nullable>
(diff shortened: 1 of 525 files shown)


sdk/dotnet/Pulumi.Kubernetes.csproj
```diff
@@ -9,7 +9,7 @@
     <PackageProjectUrl>https://pulumi.com</PackageProjectUrl>
     <RepositoryUrl>https://github.com/pulumi/pulumi-kubernetes</RepositoryUrl>
     <PackageIcon>logo.png</PackageIcon>
-    <Version>4.14.0-alpha.1718665781</Version>
+    <Version>4.0.0-alpha.0+dev</Version>
```
julienp, line 12: Should I do something in CI here? Will this be done by some release process?

julienp:
How are tests run, are they triggered manually?

sdk/dotnet/Pulumi.Kubernetes.csproj
rquitales: We can ignore this change for now - but we should fix it in the future to prevent it from always changing. The final release version will be embedded in the release process, so it's irrelevant here.

rquitales:
> How are tests run, are they triggered manually?

It should be automatic. I rebased this PR to latest change in `master` and that seems to have triggerred CI.

blampe (approved):

PR #3078, opened 2024-06-27
Propagate RetainOnDelete to child resources

### LINKED ISSUE #3065: RetainOnDelete NOT Work ON chart sub resource
### What happened?

create new chart with options `pulumi.RetainOnDelete(true)`.
when i  delete the chart, only retain the chart resource, all its sub resources have been delete.

```
     Type                                                              Name                                       Plan               Info
     pulumi:pulumi:Stack                                               ob-component-qa                                               6 messages
 ~   ├─ pulumi:providers:kubernetes                                    example-provider                           update             [diff: ~kubeconfig,version
 -   └─ kubernetes:helm.sh/v3:Chart                                    agent-demo                                 delete[retain]
 -      ├─ kubernetes:rbac.authorization.k8s.io/v1:ClusterRoleBinding  agent-demo-scale-agent-demo-rolebinding    delete
 -      ├─ kubernetes:core/v1:Service                                  mo-ob/agent-demo-scale-agent-demo          delete
 -      ├─ kubernetes:rbac.authorization.k8s.io/v1:ClusterRole         agent-demo-scale-agent-demo                delete
 -      ├─ kubernetes:core/v1:ServiceAccount                           mo-ob/agent-demo-scale-agent-demo          delete
 -      └─ kubernetes:apps/v1:Deployment                               mo-ob/agent-demo-scale-agent-demo          delete
 ```

### Example

```golang
import (
        "github.com/pulumi/pulumi-kubernetes/sdk/v3/go/kubernetes"
	helmv3 "github.com/pulumi/pulumi-kubernetes/sdk/v3/go/kubernetes/helm/v3"
        "github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

	if _, err := helmv3.NewChart(ctx, "myChart", helmv3.ChartArgs{
		Path:      pulumi.String(chartPath),
		Namespace: pulumi.String(namespace),
		Values:    values,
	}, pulumi.Provider(provider), pulumi.RetainOnDelete(true)); err != nil {
		return err
	}
```

### Output of `pulumi about`

CLI
Version      3.103.0
Go Version   go1.21.5
Go Compiler  gc

Plugins
NAME        VERSION
alicloud    3.37.0
go          unknown
kubernetes  4.13.1
kubernetes  3.29.0

Host
OS       darwin
Version  14.4.1
Arch     arm64

### Additional context

_No response_

### Contributing

blampe:
WIP - needs tests.

Closes #3065

### DIFF
--- a/provider/pkg/provider/helm/v4/chart.go
+++ b/provider/pkg/provider/helm/v4/chart.go
@@ -260,9 +260,11 @@ func (r *ChartProvider) Construct(ctx *pulumi.Context, typ, name string, inputs 
 		Objects:         objs,
 		ResourcePrefix:  *chartArgs.ResourcePrefix,
 		SkipAwait:       chartArgs.SkipAwait,
+		ParentOptions:   options,
 		ResourceOptions: []pulumi.ResourceOption{pulumi.Parent(comp)},
 		PreRegisterF: func(ctx *pulumi.Context, apiVersion, kind, resourceName string, obj *unstructured.Unstructured,
-			resourceOpts []pulumi.ResourceOption) (*unstructured.Unstructured, []pulumi.ResourceOption) {
+			resourceOpts []pulumi.ResourceOption,
+		) (*unstructured.Unstructured, []pulumi.ResourceOption) {
 			return preregister(ctx, comp, obj, resourceOpts)
 		},
 	}
@@ -276,8 +278,8 @@ func (r *ChartProvider) Construct(ctx *pulumi.Context, typ, name string, inputs 
 }
 
 func preregister(ctx *pulumi.Context, comp *ChartState, obj *unstructured.Unstructured,
-	resourceOpts []pulumi.ResourceOption) (*unstructured.Unstructured, []pulumi.ResourceOption) {
-
+	resourceOpts []pulumi.ResourceOption,
+) (*unstructured.Unstructured, []pulumi.ResourceOption) {
 	// Implement support for Helm resource policies.
 	// https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource
 	policy, hasPolicy, err := unstructured.NestedString(obj.Object, "metadata", "annotations", helmkube.ResourcePolicyAnno)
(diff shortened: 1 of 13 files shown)


provider/pkg/provider/helm/v4/chart.go
```diff
@@ -260,9 +260,11 @@ func (r *ChartProvider) Construct(ctx *pulumi.Context, typ, name string, inputs
 		Objects:         objs,
 		ResourcePrefix:  *chartArgs.ResourcePrefix,
 		SkipAwait:       chartArgs.SkipAwait,
+		ParentOptions:   options,
```
EronWright, line 263: It seems awkward to have both `ParentOptions` and `ResourceOptions`. I think you should drop `ParentOptions` and introduce a yamlv2.GetChildOptions helper function to compute the value for line 264, the `ResourceOptions`.  In other words, why not compute the options eagerly at this point.

EronWright (commented):
Looks pretty good, but can we introduce a GetChildOptions function for the MLCs, for consistency with the overlays?  It seems like an omission (of mine) that the plugin url/version isn't propagated.

EronWright:
Here's a thought-experiment to argue against taking this change.

Imagine you are deploying a third-party Chart and, over time, the chart version changes and, over time, the chart's resources change.  For example, a `RoleBinding` isn't needed anymore.

Now, imagine the user has marked the chart as `protect` and/or `retainOnDelete`.  Should the user get errors over time about how the child resources cannot be deleted? Should the old `RoleBinding` be "retained"?  I think no.

I believe that the engine is in a better position to 'get it right'.  Imagine the children aren't marked as `retainOnDelete`, but the chart is, and the user attempts to delete the chart. If the engine did propagation, and if the propagation logic was applied only at deletion time, Pulumi would correctly retain all the children. In other words, an inherited retain may have a slightly different semantic, e.g. inherited only at deletion time.  This would allow the internal structure of the component to vary over time without any UX issues.

EronWright:
@blampe can we close this PR?

blampe:
Recommendation will be to use transforms to apply this to child resources.

Issue #3094, opened 2024-07-09
`pulumi preview` patching issues


alayshia:
### What happened?

`pulumi preview` **updates** Kubernetes custom resource definitions while it only supposed to **patch** them in dry-run mode. 

### Example

```
import * as eks from '@pulumi/eks'; // 1.0.4
import * as k8s from '@pulumi/kubernetes'; // 4.7.1
import * as pulumi from '@pulumi/pulumi'; // 3.119.0

const cluster = new eks.Cluster('eks-cluster', {});

const k8sProvider = new k8s.Provider('eks-provider', {
    context: 'aws',
    kubeconfig: cluster.kubeconfig.apply(JSON.stringify),
    enableServerSideApply: false, // CRDs get updated regardless of the value
});

new k8s.helm.v3.Chart('karpenter-crds', {
    chart: '',
    version: '0.36.2', // alternate between '0.36.2' and '0.37.0'
    namespace: 'kube-system',
    transformations: [
        function (obj: any, _opts: pulumi.CustomResourceOptions): void {
            if (obj.kind !== "CustomResourceDefinition") {
                obj.apiVersion = "v1";
                obj.kind = "List";
            }
        }
    ],
}, { provider: k8sProvider });
```

1. `pulumi up` using `0.37.0`
2. `kubectl` to get manifest
3. `pulumi preview` with version `0.36.2`
4. `kubectl` to check if manifest has changed

### Output of `pulumi about`

'@pulumi/eks'; // 1.0.4
 '@pulumi/kubernetes'; // 4.7.1
 '@pulumi/pulumi'; // 3.119.0

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Our provider is not specifying the `dryRun` option here: https://github.com/pulumi/pulumi-kubernetes/blob/b6ede27fbf54aa70018cf3cca67a1110480265d8/provider/pkg/await/await.go#L499 resulting in previews always applying the resource on cluster. Thanks for reporting this issue!

PR #3096, opened 2024-07-09
Fix: prevent CRDs from always applying during preview

rquitales:
### Proposed changes

This PR ensures the correct `metav1.UpdateOptions` are set for CRD resources. The FieldManager and DryRun options need to be propagated to ensure we are updating CRD resources correctly.

This PR adds a regression test that fails without the added changes in this PR.

### Related issues (optional)

Fixes: #3094

### DIFF
--- a/provider/pkg/await/await.go
+++ b/provider/pkg/await/await.go
@@ -496,7 +496,16 @@ func updateResource(c *UpdateConfig, liveOldObj *unstructured.Unstructured, clie
 		// results in the immediate replacement of the CRD without deleting it, or any CustomResources that depend on
 		// it. The PUT operation is still validated by the api server, so a badly formed request will fail as usual.
 		c.Inputs.SetResourceVersion(liveOldObj.GetResourceVersion())
-		currentOutputs, err = client.Update(c.Context, c.Inputs, metav1.UpdateOptions{})
+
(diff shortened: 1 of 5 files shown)


provider/pkg/await/await.go
```diff
@@ -496,7 +496,16 @@ func updateResource(c *UpdateConfig, liveOldObj *unstructured.Unstructured, clie
 		// results in the immediate replacement of the CRD without deleting it, or any CustomResources that depend on
 		// it. The PUT operation is still validated by the api server, so a badly formed request will fail as usual.
 		c.Inputs.SetResourceVersion(liveOldObj.GetResourceVersion())
-		currentOutputs, err = client.Update(c.Context, c.Inputs, metav1.UpdateOptions{})
+
+		options := metav1.UpdateOptions{
+			FieldManager: c.FieldManager,
```
EronWright, line 501: Any potential for conflicts here during upgrade?

EronWright (approved):

provider/pkg/await/await.go
rquitales: The CRD is applied using `client.Update`, so conflicts will never happen.

blampe (approved):

Issue #3023, opened 2024-05-24
CustomResourcePatch resource doesn't behave like a normal Patch resource


EronWright:
### What happened?

I applied a patch to an object using the `CustomResourcePatch` resource. I later destroyed the stack, and was surprised to find that the object was unexpectedly deleted.  Patch resources aren't supposed to delete the object.

The problem seems to exist in all SDKs.

### Example

Apply a manifest directly to your Kubernetes cluster to create an issuer that is not lifecycle-managed by Pulumi:
```yaml
apiVersion: cert-manager.io/v1
kind: Issuer
metadata:
  name: issuer1
spec:
  selfSigned: {}
```
 
Patch the issuer using a Pulumi program and the `CustomResourcePatch` resource. Note how the patch resource isn't fully-specified; there's no `spec` in this case.
```ts
import * as kubernetes from "@pulumi/kubernetes";

const issuerPatch = new kubernetes.apiextensions.CustomResourcePatch("issuer1", {
    kind: "Issuer",
    apiVersion: "cert-manager.io/v1",
    metadata: {
        name: "issuer1",
        annotations: {
            "example.com": "patched",
        },
    }
});
```

Deploy the stack and observe the `example.com: patched` annotation, then destroy the stack. One expects the annotation to be removed and for the object to still exist. Unexpectedly, the object is gone.

### Output of `pulumi about`
```
CLI          
Version      3.115.1-dev.0
Go Version   go1.22.1
Go Compiler  gc

Plugins
KIND      NAME        VERSION
resource  kubernetes  4.4.0
language  nodejs      unknown

Host     
OS       darwin
Version  14.4.1
Arch     arm64

Dependencies:
NAME                VERSION
@pulumi/kubernetes  4.4.0
@pulumi/pulumi      3.116.1

```

### Additional context

I believe the root cause is that the provider uses the "patch" logic only for a fixed list of resource types, [here](https://github.com/pulumi/pulumi-kubernetes/blob/4e4e91de7bcb74088a546c21945bff23dff41366/provider/pkg/kinds/util.go#L8). In the example, the resource type is `kubernetes:cert-manager.io/v1:IssuerPatch`, and the provider simply doesn't recognize it as a patch resource type. 

If one were to use `CustomResourcePatch` to patch a `ConfigMap`, it would probably work as expected. I don't think the problem is at the SDK level.

### Contributing

Vote on this issue by adding a 👍 reaction. 
[...]

PR #3102, opened 2024-07-12
Enhance detection logic for patch variants

rquitales:
### Proposed Changes

This PR improves our Patch resource detection logic by removing the reliance on a hardcoded list. Hardcoded lists are fragile because new Patch resources introduced in future SDK versions, or those generated dynamically by tools like `crd2pulumi`, would **not** be recognized as Patch variants by our provider. This issue is already present with `CustomResourcePatch` resources, as CustomResource kinds will never be included in a hardcoded list.

The updated logic now compares the resource URN and the object's Kind. Since the token within the URN of a patch variant always follows the pattern `kind + "Patch"`, we can use this for comparison. Previously, the logic simply checked if the URN ended in `Patch`, which was inaccurate due to the possibility of a Custom Resource (CR) Kind ending in `Patch` (e.g., "MyResourcePatch"). This would lead to the Pulumi patch variant URN being `MyResourcePatchPatch`.

#### Testing

- Added integration tests for the YAML, Node.js, and Go languages since CustomResources are overlays and not supported in YAML. Existing tests did not ensure that deleted Patch resources only removed the fields applied, and not the whole object. This behavior is now exercised as well.
- Expanded existing unit tests.

Note: I did not expand the test coverage for Python, Java, and .NET at this time, as porting the test programs for these languages requires additional effort. We may revisit and add tests for these languages in the future.

### Related Issues (Optional)

Fixes: #3023


provider/pkg/provider/provider_test.go
```diff
@@ -155,44 +155,6 @@ func Test_equalNumbers(t *testing.T) {
 	}
 }
 
-func Test_isPatchURN(t *testing.T) {
```
rquitales, line 158: This is duplicated in `provider/pkg/kinds/util_test.go`.

provider/pkg/kinds/util.go
```diff
@@ -4,11 +4,12 @@ import (
...
-// IsPatchURN returns true if the URN is for a Patch resource.
-func IsPatchURN(urn resource.URN) bool {
-	urnS := urn.Type().String()
+// IsPatchResource returns true if it is a Patch resource and also ensures
+// that we don't return true for CustomResources that end with "Patch".
+func IsPatchResource(urn resource.URN, kind string) bool {
+	resourceName := urn.Type().Name().String()
 
-	return PatchQualifiedTypes.Has(urnS)
+	return kind+"Patch" == resourceName
```
rquitales, line 12: The crux of the change.

blampe (approved):
Brilliant!

EronWright (changes requested):
I don't understand how a non-patch resource of kind `MyResourcePatch` would work.

rquitales:
> I don't understand how a non-patch resource of kind `MyResourcePatch` would work.

For a non-patch resource with Kind `MyResourcePatch`, the urn for the non-patch variant would be something like `kubernetes:gv:MyResourcePatch`. The urn for the patch variant will be `kubernetes:gv:MyResourcePatchPatch`. The `kind` for both patch, and non-patch variant is **always** `MyResourcePatch`.

So, we can determine if a resource is the patch variant or not simply by checking that: `kind+"Patch == pulumiResourceName`.

This PR includes both unit and integration tests to showcase this behavior:
Unit test: https://github.com/pulumi/pulumi-kubernetes/blob/d795fadc001b2d55b1ee87de7b8dfe19968cd262/provider/pkg/kinds/util_test.go#L24-L34
Integration test (nodejs example): https://github.com/pulumi/pulumi-kubernetes/blob/d795fadc001b2d55b1ee87de7b8dfe19968cd262/tests/sdk/java/testdata/patch-resources/nodejs/step2/index.ts#L120-L148

EronWright (approved):

PR #3107, opened 2024-07-15
Define Supported Languages for Overlay Resources

flostadler:
### Proposed changes

docsgen supports defining the supported languages of an overlay now.
This will ensure only the supported languages show up in the language choosers
in the registry.

Example for Chart v3:
Screenshot 2024-07-01 at 16 11 23

### Related issues (optional)

Relates to https://github.com/pulumi/pulumi/issues/13231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Added
 
-- Define Supported Languages for Chart v3 (https://github.com/pulumi/pulumi-kubernetes/pull/3107)
+- Define Supported Languages for Overlays (https://github.com/pulumi/pulumi-kubernetes/pull/3107)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Added
+
+- Define Supported Languages for Chart v3 (https://github.com/pulumi/pulumi-kubernetes/pull/3107)
+
 ### Fixed
 
 - Updated logic to accurately detect if a resource is a Patch variant (https://github.com/pulumi/pulumi-kubernetes/pull/3102)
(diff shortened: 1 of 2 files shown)


rquitales:
I understand the scope of this PR is currently limited to Chart v3, but should we also include specify the languages supported for the other overlays in this provider as well? `apiextensions.CustomResource`, `kustomize` (v1) and `yaml` (v1).

Ref: https://github.com/pulumi/pulumi-kubernetes/blob/cf02cf4a059cf77e7a2fe9dd13b280d3a2be6344/provider/cmd/pulumi-gen-kubernetes/main.go#L228-L232

flostadler:
> I understand the scope of this PR is currently limited to Chart v3, but should we also include specifying the languages supported for the other overlays in this provider as well? `apiextensions.CustomResource`, `kustomize` (v1) and `yaml` (v1).
> 
> Ref:
> 
> https://github.com/pulumi/pulumi-kubernetes/blob/cf02cf4a059cf77e7a2fe9dd13b280d3a2be6344/provider/cmd/pulumi-gen-kubernetes/main.go#L228-L232

@rquitales Good point, I updated the PR!

rquitales (approved):

EronWright:
@flostadler I should mention that the Java SDK does in fact have CustomResource support; see https://github.com/pulumi/pulumi-kubernetes/pull/3020.  If you have a bit of time, could you open a PR to fix the `apiextensionsCustomResource` and `apiextensionsCustomResourcePatch` entries to include Java.

I would also argue that "csharp" is NOT a supported language for CustomResource.  Yes we have a CustomResource class in the .NET SDK, but it doesn't have a public constructor because it is intended to be subclassed, not used directly as is the case with the other SDKs. 

Thanks!

PR #3128, opened 2024-07-24
Fix pod info reporting

blampe:
This fixes 3 issues:

1. The biggest problem was that `relatedResource` only considered two resources related if they had equal generations. As far as I can tell this is almost never the case, at least not where we're currently using `PodAggregator` (Job/Pod and DaemonSet/Pod relationships) because  Pods usually don't have a generation. This also doesn't make sense for ReplicaSet/Deployment relationships because the generations can differ. This PR relaxes that restriction, so we now surface informational messages that were previously being discarded during updates.
2. `isOwnedBy` was using heuristics on GVK and names to determine ownership but this can return false positives. This PR makes the logic exact by matching on UIDs. This is relevant to generic await logic, because we'll want to accurately report status messages for arbitrary Parent→Child relationships (including custom resources) so we need something more reliable than the name.
3. `relatedResource` and `isOwnedBy` were using different logic to determine whether one resource owns another. We can safely remove `relatedResource` since it is equivalent to `isOwnedBy` without the constraint on generation. 

Along the way I also removed `ResourceID` since it wasn't providing much value over plain unstructured resources.

`statefulset_test.go` and `deployment_test.go` have large diffs but these are just mechanical changes to simply set UIDs correctly on the test resources.

Example of DaemonSet output before this change:
```
 +  kubernetes:apps/v1:DaemonSet ds creating (0s) DaemonSet status.observedGeneration not found
 +  kubernetes:apps/v1:DaemonSet ds creating (0s) Current: 0/1
 +  kubernetes:apps/v1:DaemonSet ds creating (0s) Available: 0/1
@ updating.........
 +  kubernetes:apps/v1:DaemonSet ds creating (5s) All replicas scheduled as expected. Replicas: 1
```

After this change the pod's status is reported:
```
 +  kubernetes:apps/v1:DaemonSet ds creating (0s) DaemonSet status.observedGeneration not found
 +  kubernetes:apps/v1:DaemonSet ds creating (0s) Current: 0/1
 +  kubernetes:apps/v1:DaemonSet ds creating (0s) Available: 0/1
 +  kubernetes:apps/v1:DaemonSet ds creating (0s) warning: [Pod await-daemonset-xww9v]: containers with unready status: [nginx]
@ updating.........
 +  kubernetes:apps/v1:DaemonSet ds creating (6s) All replicas scheduled as expected. Replicas: 1
```


provider/pkg/await/util.go
```diff
@@ -98,37 +98,18 @@ func is404(err error) bool {
...
-		switch owner.Kind {
-		case "Deployment":
-			ownerAPIVersion = canonicalizeDeploymentAPIVersion(owner.APIVersion)
-		case "StatefulSet":
-			ownerAPIVersion = canonicalizeStatefulSetAPIVersion(owner.APIVersion)
-		}
-
-		if ownerAPIVersion == possibleOwnerAPIVersion &&
-			possibleOwner.GetKind() == owner.Kind && possibleOwner.GetName() == owner.Name {
+		if owner.UID == possibleOwner.GetUID() {
```
blampe, line 110: The crux of the change.

rquitales:
>This is relevant to generic await logic, because we'll want to accurately report status messages for arbitrary Parent→Child relationships (including custom resources) so we need something more reliable than the name.

Switching to match on UID makes sense to me. For my understanding however, shouldn't we still be able to accurately find Parent→Child relationships based on the existing logic we have to match GVK and object name in the `ownerReference`?

As reference, here is an example upstream code that matches on all coordinates: https://github.com/kubernetes/kubernetes/blob/9edabd617945cd23111fd46cfc9a09fe37ed194a/pkg/controller/garbagecollector/graph.go#L204-L206

rquitales (approved):

blampe:
> Switching to match on UID makes sense to me. For my understanding however, shouldn't we still be able to accurately find Parent→Child relationships based on the existing logic we have to match GVK and object name in the ownerReference?

@rquitales it's subtle, but our existing logic usually discards the GVK and only matches on name! Note `ownerAPIVersion` and `possibleOwnerAPIVersion` remain uninitialized (`""`) unless the owner is a Deployment/StatefulSet.

https://github.com/pulumi/pulumi-kubernetes/blob/53c18349cee59d1a1233f9284c56a162d300a655/provider/pkg/await/util.go#L107-L131

It would be straightforward to fix this, but I'm assuming UID is simple/reliable enough.

Issue #895, opened 2019-11-26
Cannot use PVC (eternal wait)


brandonkal:
### Problem description

1. Pulumi Creates PVC
2. PVC reports:
```
Status: Pending
WaitForFirstConsumer
waiting for first consumer to be created before binding
```
3. Pulumi waits 5 minutes and then times out
```
resource pvc was successfully created, but the Kubernetes API server reported that it failed to fully initialize or become live: Timeout occurred polling for 'pvc'
```

Pulumi never creates the deployment because `pvc.mount('/data')` from `kubernetesx` is used. However, I do not believe this bug is specific to kx.

The problem is the pvc waits for a pod to request it while the pulumi controller waits for the pvc to no longer be pending, which of course won't happen unless pulumi is smart enough to create that deployment while the PVC is still pending.

gregnr:
We are encountering this as well on AKS who's storage classes default to:
```yaml
volumeBindingMode: WaitForFirstConsumer
```

It's a chicken-and-egg scenario, where Pulumi doesn't want to create the `Deployment` until the `PersistentVolumeClaim` has been created (because the PVC name is a dependency), but Kubernetes doesn't want to create the `PersistentVolumeClaim` until a Pod is created, because of the above `volumeBindingMode`.

Our workaround was to create a custom `StorageClass` with 
```yaml
volumeBindingMode: Immediate
```

and manually set each PVC's storage class to the above custom one.

An alternative workaround would just be to ensure your `Deployment` doesn't depend on your `PVC` at all from Pulumi's perspective, ie. don't reference your `pvc.metadata.name` and instead hardcode the name. This of course assumes that the PVC name is known at build time.

rvowles:
I'm hitting this on KinD as well.

FP-Guitar:
Hitting this aswell on AKS

FP-Guitar:
Found a feasible solution:
https://www.pulumi.com/blog/improving-kubernetes-management-with-pulumis-await-logic/
tl:dr
_To disable await logic for a selected resource, add the following annotation to the resource’s metadata: pulumi.com/skipAwait: "true"_
Add this to your PVC and pulumi will NOT wait for the resource to be ready getting you out of this Catch22

isaac-mcfadyen:
Is there any update on this? This is a *major* issue to have happen by default and seems like a simple fix. Even Terraform has something more standardized: it's a dedicated field on the PVC instead of an ad-hoc annotation.

[...]

PR #3130, opened 2024-07-24
Fix endless PVC await

blampe:
This fixes an issue where `WaitForFirstConsumer` PVCs could cause deadlocks due to the downstream consumer waiting for the PVC to be bound.

We now attempt to fetch the PVC's storage class and the corresponding `volumeBindMode`. If we can find the bind mode, then we consider the PVC ready once it's `Pending`. The existing behavior is preserved if we can't find the bind mode, or if it's `Immediate`. 

An e2e test is included. I'm omitting unit tests as this might end up getting refactored in some later PRs.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/895.


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 
 - Updated logic to accurately detect if a resource is a Patch variant. (https://github.com/pulumi/pulumi-kubernetes/pull/3102)
 - Added java as supported language for CustomResource overlays. (https://github.com/pulumi/pulumi-kubernetes/pull/3120)
+- `PersistentVolumeClaims` with a bind mode of `WaitForFirstConsumer` will no
+  long hang indefinitely. (https://github.com/pulumi/pulumi-kubernetes/pull/3130)
```
rquitales, line 16:
```suggestion
  longer hang indefinitely. (https://github.com/pulumi/pulumi-kubernetes/pull/3130)
```

provider/pkg/await/awaiters.go
```diff
@@ -483,19 +483,57 @@ func untilCoreV1PersistentVolumeInitialized(c createAwaitConfig) error {
...
+		// Lookup the PVC's storage class once it's available.
+		if bindMode == "" {
+			b, err := pvcBindMode(c.ctx, c.clientSet, pvc)
+			if err != nil {
+				c.logStatus(diag.Warning, err.Error())
+			}
+			bindMode = b
+		}
+
+		phase, _ := openapi.Pluck(pvc.Object, "status", "phase")
```
rquitales, line 498: nit: prefer using `unstructured.NestedString`

provider/pkg/await/awaiters.go
```diff
@@ -483,19 +483,57 @@ func untilCoreV1PersistentVolumeInitialized(c createAwaitConfig) error {
...
+	ctx context.Context,
+	clientset *clients.DynamicClientSet,
+	pvc *unstructured.Unstructured,
+) (string, error) {
+	gvk := storagev1.SchemeGroupVersion.WithKind("StorageClass")
+	scClient, err := clientset.ResourceClient(gvk, "")
+	if err != nil {
+		return "", fmt.Errorf("getting storageclass client: %w", err)
+	}
+	name, _, err := unstructured.NestedString(pvc.Object, "spec", "storageClassName")
```
rquitales, line 527: Should we also validate that `name` is not empty?

rquitales (approved):

PR #3133, opened 2024-07-26
Unify await logic for deletes

blampe:
NB: This is a larger change that's easier to review as separate commits. The
first commit introduces some new types and interfaces; the second one hooks up
those types and removes a lot of existing code.

We already have "generic" await logic for deletion -- if we don't have an
explicit delete-awaiter defined for a particular GVK, then we always run some
logic that waits for the resource to 404. 

As it turns out, all of our custom delete-awaiters do essentially the same 404
check, modulo differences in the messages we log. This makes the deletion code
flow a good starting place to introduce more generic/unified await logic.

If you look at the code deleted in the second commit you get a good sense of
the current issues with our awaiters: each custom awaiter is responsible for
establishing its own watchers; determining its default timeout; performing the
same 404 check; etc. There's a lot of duplication and subtle differences in
behavior that leads to issues like
https://github.com/pulumi/pulumi-kubernetes/issues/1232.

As part of this change we start decomposing our await logic into more
composable pieces. Note that I'm replacing our deletion code path with these
new pieces because I don't think we lose much by changing the logged messages,
but for the create/update code paths we'll need some glue to preserve existing
await behavior.

The relevant interfaces are:
```go
// Observer acts on a watch.Event Source. Range is responsible for filtering
// events to only those relevant to the Observer, and Observe optionally
// updates the Observer's state.
type Observer interface {
	// Range iterates over all events visible to the Observer. The caller is
	// responsible for invoking Observe as part of the provided callback. Range
	// can be used to customize setup and teardown behavior if the Observer
	// wraps another Observer.
	Range(func(watch.Event) bool)

	// Observe handles events and can optionally update the Observer's state.
	// This should be invoked by the caller and not during Range.
	Observe(watch.Event) error
}

// Satisfier is an Observer which evaluates the observed object against some
// criteria.
type Satisfier interface {
	Observer

	// Satisfied returns true when the criteria is met.
	Satisfied() (bool, error)

	// Object returns the last-known state of the object being observed.
	Object() *unstructured.Unstructured
}

// Source encapsulates logic responsible for establishing
// watch.Event channels.
type Source interface {
	Start(context.Context, schema.GroupVersionKind) (<-chan watch.Event, error)
}
```

At a high level: 
1. We determine what condition (`Satisfier`) to wait for during deletion. There is
   always a condition even if it's a no-op. Deletion is simple because there
   are only two possibilities -- "skip" and "wait for 404" -- but with
   create/update and user-defined conditions it will get more interesting. 
2. We wait for the condition `Satisfier` and can combine it with arbitrary
   `Observers`. This lets us do things like log additional information while
   we're waiting, e.g. https://github.com/pulumi/pulumi-kubernetes/pull/3135.

The underlying machinery responsible for handling timeouts, informers, etc. is
all hidden behind the `Source`. Implementing new await logic is essentially
just a matter of defining a new `Satisifer` which understands how to evaluate
an unstructured resource.

A number of unit tests are included as well as an E2E regression test to ensure we respect the `skipAwait` annotation. The existing delete-await tests are mostly unchanged except for tweaks to inject a `Condition` instead of an `awaitSpec`. Some watcher-specific tests were no longer relevant and were removed, however the functionality is still implemented/tested as part of `Awaiter`. 

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/3157.
Fixes https://github.com/pulumi/pulumi-kubernetes/issues/1418.

### DIFF
--- a/provider/pkg/await/await_test.go
+++ b/provider/pkg/await/await_test.go
@@ -767,7 +767,7 @@ func TestDeletion(t *testing.T) {
 				name:    "foo",
 				objects: []runtime.Object{validPodUnstructured},
 				inputs:  withSkipAwait(validPodUnstructured),
-				outputs: withSkipAwait(validPodUnstructured),
+				outputs: validPodUnstructured,
 			},
 			reaction: []reactionF{suppressDeletion}, // suppress deletion to safeguard that the built-in watcher is not used.
 			expect:   []expectF{succeeded()},
--- a/provider/pkg/await/await_test.go
+++ b/provider/pkg/await/await_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/await/condition"
 	"github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/clients"
 	"github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/clients/fake"
 	"github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/cluster"
@@ -663,23 +664,16 @@ func TestDeletion(t *testing.T) {
 
 	// awaiters
 
-	awaitNoop := func(t *testing.T, ctx testCtx) deletionAwaiter {
-		return func(dac deleteAwaitConfig) error {
-			return nil
-		}
+	awaitNoop := func(t *testing.T, ctx testCtx) condition.Satisfier {
+		return condition.NewImmediate(ctx.config.DedupLogger, ctx.config.Inputs)
 	}
 
-	awaitError := func(t *testing.T, ctx testCtx) deletionAwaiter {
-		return func(dac deleteAwaitConfig) error {
-			return serviceUnavailableErr
-		}
+	awaitError := func(t *testing.T, ctx testCtx) condition.Satisfier {
+		return condition.NewFailure(serviceUnavailableErr)
 	}
 
-	awaitUnexpected := func(t *testing.T, ctx testCtx) deletionAwaiter {
-		return func(dac deleteAwaitConfig) error {
-			require.Fail(t, "Unexpected call to awaiter")
-			return nil
-		}
+	awaitUnexpected := func(t *testing.T, ctx testCtx) condition.Satisfier {
+		return condition.NewFailure(fmt.Errorf("unexpected call to await"))
 	}
 
 	// expectations
@@ -705,13 +699,13 @@ func TestDeletion(t *testing.T) {
 	}
 
 	tests := []struct {
-		name     string
-		client   client
-		args     args
-		expect   []expectF
-		reaction []reactionF
-		watcher  *watch.RaceFreeFakeWatcher
-		awaiter  func(t *testing.T, ctx testCtx) deletionAwaiter
+		name      string
+		client    client
+		args      args
+		expect    []expectF
+		reaction  []reactionF
+		watcher   *watch.RaceFreeFakeWatcher
+		condition func(*testing.T, testCtx) condition.Satisfier
 	}{
 		{
 			name: "ServiceUnavailable",
@@ -738,8 +732,8 @@ func TestDeletion(t *testing.T) {
 				inputs:  validPodUnstructured,
 				outputs: validPodUnstructured,
 			},
-			awaiter: awaitNoop,
-			expect:  []expectF{succeeded(), deleted("default", "foo")},
+			condition: awaitNoop,
+			expect:    []expectF{succeeded(), deleted("default", "foo")},
 		},
 		{
 			name: "NonNamespaced",
@@ -750,8 +744,8 @@ func TestDeletion(t *testing.T) {
 				inputs:  validClusterRoleUnstructured,
 				outputs: validClusterRoleUnstructured,
 			},
-			awaiter: awaitNoop,
-			expect:  []expectF{succeeded(), deleted("default", "foo")},
+			condition: awaitNoop,
+			expect:    []expectF{succeeded(), deleted("default", "foo")},
 		},
 		{
 			name: "Gone",
@@ -762,8 +756,8 @@ func TestDeletion(t *testing.T) {
 				inputs:  validPodUnstructured,
 				outputs: validPodUnstructured,
 			},
-			awaiter: awaitUnexpected,
-			expect:  []expectF{succeeded()},
+			condition: awaitUnexpected,
+			expect:    []expectF{succeeded()},
 		},
 		{
 			name: "SkipAwait",
@@ -772,10 +766,9 @@ func TestDeletion(t *testing.T) {
 				name:    "foo",
 				objects: []runtime.Object{validPodUnstructured},
 				inputs:  withSkipAwait(validPodUnstructured),
-				outputs: validPodUnstructured,
+				outputs: withSkipAwait(validPodUnstructured),
 			},
 			reaction: []reactionF{suppressDeletion}, // suppress deletion to safeguard that the built-in watcher is not used.
-			awaiter:  awaitUnexpected,
 			expect:   []expectF{succeeded()},
 		},
 		{
@@ -787,8 +780,8 @@ func TestDeletion(t *testing.T) {
 				inputs:  validPodUnstructured,
 				outputs: validPodUnstructured,
 			},
-			awaiter: awaitError,
-			expect:  []expectF{failed(serviceUnavailableErr)},
+			condition: awaitError,
+			expect:    []expectF{failed(serviceUnavailableErr)},
 		},
 		{
 			name: "Deleted",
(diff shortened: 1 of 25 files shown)


blampe:

- #3133 ◀

provider/pkg/await/await_test.go
```diff
@@ -501,10 +495,9 @@ func Test_Deletion(t *testing.T) {
 				name:    "foo",
 				objects: []runtime.Object{validPodUnstructured},
 				inputs:  withSkipAwait(validPodUnstructured),
-				outputs: validPodUnstructured,
+				outputs: withSkipAwait(validPodUnstructured),
```
blampe, line 498: The test was previously using `c.Inputs` to check whether to skip the await. We're now using `c.Output` which is equivalent for the deletion code path.

provider/pkg/await/await_test.go
```diff
@@ -528,64 +521,8 @@ func Test_Deletion(t *testing.T) {
...
-				inputs:  validPodUnstructured,
-				outputs: validPodUnstructured,
-			},
-			reaction: nil,
-			awaiter:  nil,
-			watcher:  withWatchClosed(watch.NewRaceFreeFake()),
-			expect:   []expectF{succeeded()},
-		},
-		{
-			name: "WatchError",
```
blampe, line 563: These are no longer needed because we're using an Informer which takes care of these errors for us.

provider/pkg/await/condition/deleted.go
```diff
@@ -0,0 +1,140 @@
...
+func (dc *Deleted) Satisfied() (bool, error) {
+	if dc.deleted.Load() {
+		return true, nil
+	}
+
+	uns := dc.Object()
+	r, _ := status.Compute(uns)
+	if r.Message != "" {
+		dc.logger.LogMessage(checkerlog.StatusMessage(r.Message))
+	}
```
blampe, line 117: We will still emit messages related to the object's status, if possible.

provider/pkg/await/internal/awaiter.go
```diff
@@ -0,0 +1,171 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO:
+// event aggregator
+// thing to spawn a bunch of observers
+// legacy condition to bridge shit
```
rquitales, line 18: Rephrase?

provider/pkg/await/internal/awaiter_test.go
```diff
@@ -0,0 +1,181 @@
...
+		WithObservers(condition.NewNever(nil)),
+	)
+	require.NoError(t, err)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	err = awaiter.Await(ctx)
+
+	assert.Error(t, err)
```
rquitales, line 44: Does it make sense to be more explicit with our error assertion here?

provider/pkg/await/internal/awaiter.go
```diff
@@ -0,0 +1,171 @@
...
+	// Before returning we attempt to re-evaluate the condition a final time,
+	// and we wrap our error with our object's last known state so it can be
+	// checkpointed.
+	defer func() {
+		if err == nil {
+			// Nothing to do.
+			return
+		}
+		// Make sure Observers are all done.
+		wg.Wait()
```
rquitales, line 95: In the event of context deadline being exceeded, wouldn't this mean that the awaiter still continues waiting/observing? Shouldn't we just skip waiting for the wait group to be done?

provider/pkg/await/await_test.go
blampe: Fair! I'll try switching this back - I think I ran into a failure somewhere, but it might have been due to a quirk of the test or something else.

.github/workflows/run-acceptance-tests.yml
```diff
@@ -6,10 +6,6 @@ on:
     types:
     - run-acceptance-tests-command
   pull_request:
-    branches:
```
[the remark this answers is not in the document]
blampe: Nope! This was temporary to get tests running until https://github.com/pulumi/ci-mgmt/pull/1047 merged.

tests/sdk/java/await_test.go
```diff
@@ -92,3 +93,33 @@ func TestAwaitPVC(t *testing.T) {
...
+	_ = test.Up()
+	took := time.Since(start)
+	assert.Less(t, took, 2*time.Minute, "didn't skip pod's slow startup")
+
+	start = time.Now()
+	_ = test.Refresh()
+	took = time.Since(start)
+	assert.Less(t, took, 2*time.Minute, "didn't skip pod's slow read")
+
+	test.UpdateSource("testdata/await/skipawait/step2")
```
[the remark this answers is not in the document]
blampe: This is very alarming! Good catch - put together a fix here https://github.com/pulumi/providertest/pull/96

tests/sdk/java/await_test.go
```diff
@@ -92,3 +93,33 @@ func TestAwaitPVC(t *testing.T) {
...
+
+	test := pulumitest.NewPulumiTest(t,
+		"testdata/await/skipawait",
+		opttest.SkipInstall(),
+	)
+
+	start := time.Now()
+	_ = test.Up()
+	took := time.Since(start)
+	assert.Less(t, took, 2*time.Minute, "didn't skip pod's slow startup")
```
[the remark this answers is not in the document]
blampe: Yeah, basically 2 minutes should be generous enough for the YAML provider to do everything it needs, and if it takes longer then it's probably waiting for the pod to come up.

provider/pkg/await/internal/awaiter.go
blampe: Notes to self, nothing to see here 😄

provider/pkg/await/internal/awaiter.go
blampe: The context is plumbed all the way down and respected by the observers, so if it dies all of our informers will shut down and this will resolve. Namely these guys:

https://github.com/pulumi/pulumi-kubernetes/pull/3133/files#diff-c68cee828d9c5172eef833ba32b6185741c858ff507bd2f4d6df8c5a6fb275dbR55-R59

https://github.com/pulumi/pulumi-kubernetes/pull/3133/files#diff-b52d2594c41dd1ff41784e1c6101fcd3c86f51c128ea8b8de43ddc6607977a15R148

Full disclosure I'm very skeptical we need any of this "Hail Mary" logic. There's a comment in the code but I think it largely stems from issues we had handling watch errors. I kept it as-is since we've got tests around it, but I expect we could get rid of it without issue.

blampe: I've simplified this a bit -- we don't need to worry about context cancellation here since the observer's already shut down when that happens.

provider/pkg/await/internal/awaiter.go
```diff
@@ -0,0 +1,146 @@
...
+				return true
+			})
+		}(o)
+	}
+
+	// Block until our condition is satisfied, or until our Context is canceled.
+	aw.condition.Range(func(e watch.Event) bool {
+		err = aw.condition.Observe(e)
+		if err != nil {
+			return false
```
EronWright, line 64: Is there some logging one should do here before discarding `err`?

provider/pkg/await/internal/awaiter.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		if done, _ := aw.condition.Satisfied(); done {
+			return false
+		}
+		return true
+	})
+
+	// Re-evaluate our condition since its state might have changed during the
+	// iterator's teardown.
+	done, err := aw.condition.Satisfied()
```
EronWright, line 74: Is `err` possibly being redefined here?  Am worried that the defer func is writing to the wrong `err`.

EronWright:
I notice that the cli-utils library has a `polling` package, seems similar the core logic in this PR.  Thoughts on it?
https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/polling/doc.go

provider/pkg/await/condition/observer.go
```diff
@@ -0,0 +1,178 @@
...
+
+// Observer acts on a watch.Event Source. Range is responsible for filtering
+// events to only those relevant to the Observer, and Observe optionally
+// updates the Observer's state.
+type Observer interface {
+	// Range iterates over all events visible to the Observer. The caller is
+	// responsible for invoking Observe as part of the provided callback. Range
+	// can be used to customize setup and teardown behavior if the Observer
+	// wraps another Observer.
+	Range(func(watch.Event) bool)
```
EronWright, line 39: Given the bit of awkwardness [with All](https://github.com/pulumi/pulumi-kubernetes/pull/3134/files#diff-2042e56949b8976fc34e1d68bb8c4f9ed01d7f9b5796392b4893ba77e0f5f416R63-R65), would it make sense to factor out the `Range` function, e..g to source (the observed thing).

provider/pkg/await/internal/awaiter.go
blampe: Added a warning.

blampe (commented):
> I notice that the cli-utils library has a watcher package, seems similar the core logic in this PR. Thoughts on it?
https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/watcher/doc.go

@EronWright yes, I saw that as well (more specifically the [polling](https://pkg.go.dev/sigs.k8s.io/cli-utils/pkg/kstatus/polling) portion). We're already using Informers for our other awaiters and they work well enough, so the primary goal is to use them for deletes as well.

provider/pkg/await/daemonset_test.go
```diff
@@ -320,20 +320,26 @@ func TestAwaitDaemonSetDelete(t *testing.T) {
 	}
 
 	for _, tt := range tests {
+		tt := tt
```
EronWright, line 323: BTW I learned that loop variables are now copied as of go 1.22 and you don't need this line anymore.

EronWright (approved):

PR #3136, opened 2024-07-27
fix: Do not set default namespace if explicitly set to empty

cenk1cenk2:
### Proposed changes

Hello,

With the new `renderToYaml` engine that is experimental, I do want to generate ArgoCD resources in a typesafe way, but there is a mechanism in the `yaml` renderer, which gets called in every other resource type like plain, Helm, Kustomize etc. to always set a default namespace if the resource does not have, so now you have to pass around namespaces and ensure that they are matching with the ArgoCD configuration, which is [here](https://github.com/pulumi/pulumi-kubernetes/blob/master/provider/pkg/provider/yaml/v2/yaml.go#L139).

When setting the `namespace` in the kubernetes provider as below empty, the proposed behavior will not put in the default namespace.

```typescript
const provider = new k8s.Provider(options.name, {
  renderYamlToDirectory: options.path,
  namespace: ''
})
```

Since the order of resolution is the resource, then the provider, then active kubernetes context namespace, this can be easily overwritten when desired, but will allow resources to also generate without a namespace field.

### Related issues (optional)

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Changed
+
+- Whenever namespace is explicitly set to an empty string, it will not add the default namespace to the namespaced resource.
+
 ### Added
 
 - `clusterIdentifier` configuration can now be used to manually control the
--- a/provider/pkg/provider/provider_configure_test.go
+++ b/provider/pkg/provider/provider_configure_test.go
@@ -106,6 +106,19 @@ var _ = Describe("RPC:Configure", func() {
 				Expect(helmFlags.Namespace).To(PointTo(Equal("pulumi")))
 			})
 		})
+
+		Context("when explicitly configured to use an empty namespace", func() {
+			JustBeforeEach(func() {
+				req.Variables["kubernetes:config:namespace"] = ""
+			})
+			It("should use the empty namespace as the default namespace", func() {
+				_, err := k.Configure(context.Background(), req)
+				Expect(err).ShouldNot(HaveOccurred())
+				Expect(k.defaultNamespace).To(Equal(""))
+				helmFlags := k.helmSettings.RESTClientGetter().(*genericclioptions.ConfigFlags)
+				Expect(helmFlags.Namespace).To(PointTo(Equal("")))
+			})
+		})
 	})
 
 	Describe("Kubeconfig Parsing", func() {
--- a/provider/pkg/provider/yaml/v2/yaml.go
+++ b/provider/pkg/provider/yaml/v2/yaml.go
@@ -170,7 +170,7 @@ func Normalize(objs []unstructured.Unstructured, defaultNamespace string, client
 		if err != nil {
 			return nil, err
 		}
-		if isNamespaced && obj.GetNamespace() == "" {
+		if isNamespaced && obj.GetNamespace() == "" && defaultNamespace != "" {
 			obj.SetNamespace(defaultNamespace)
 		}
 	}


blampe:
Thanks for the PR @cenk1cenk2!

Do you have a small example that reproduces the issue you were seeing? It would be good to include something like that as a regression test -- I can help you hook it up.

cenk1cenk2:
Hello @blampe,

Thank you for the immediate reply.

The minimum reproduction should be something like at least with typescript as follows.

```typescript
import * as k8s from '@pulumi/kubernetes'
import { join } from 'path'

new k8s.Provider('k8s', {
  renderYamlToDirectory: join(process.cwd(), 'output'),
  namespace: ''
})

new k8s.core.v1.Secret('namespace', {
  metadata: {
    name: 'namespace'
  },
  stringData: {
    test: 'true'
  }
})
```

Where you expect the output for the argocd purpose to be empty, so that it will inject the namespace itself, else it will just try to use the namespace provided and might fail because of the permissions.

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: namespace
data:
  test: dHJ1ZQ== # test
```

However since it insists on putting in a default namespace on an empty string in the go world, the actual output is as follows.

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: namespace
  namespace: default
data:
  test: dHJ1ZQ== # test
```

blampe:
As far as I can tell all of our other use cases of `defaultNamespace` do similar `k.defaultNamespace != ""` checks before using it, so this seems reasonable to me. We will review this as a team next week!

cenk1cenk2:
Thank you very much for the update @blampe. I will update branch with the latest head as soon as possible.

blampe:
@cenk1cenk2 some questions came up:
* Do you know the namespace Argo will use beforehand?
* Is it possible to post-process the manifests?

I think historically we've tried to treat `renderToYaml` as a way to see exactly what manifests Pulumi would apply if the operation was run, whereas here it seems like you're using more as a way to template something.

More specifically, this `Normalize` call also impacts the live (non-rendered) mode of operation, so we have some concerns that there is a risk of changing behavior elsewhere in subtle or surprising ways.

cenk1cenk2:
I think the originally intended use-case for this `renderToYaml` for the user to see what is the actual end results, so it makes perfect sense that it tries to render things as if `pulumi` is applying it directly to the stack.

For the questions:
- Yes, I know the namespace beforehand, that is how I am now using it.
- I have tried postprocessing with `transformations` to no avail since I can only change the arguments there and it was overwritten by the `defaultNamespace` configured in the provider.

I greatly appreciate your time in checking this, of course, I would not know how this might impact other cases that rely on this behavior. But in my humble opinion, if a user explicitly sets something to empty, I just would not babysit them in a failure case and let it be a user error in this case.

I can further modify this pull request to fit the flow and check other cases if you can give me some guidance or we can also scrap the idea as a whole.

blampe:
> Yes, I know the namespace beforehand, that is how I am now using it.

@cenk1cenk2 sorry, I'm still having trouble understanding your use case. If you already know the namespace beforehand, why do you need to unset it on the resources Pulumi generates? I'm wondering if there's something else going on -- is a namespace getting applied somewhere it shouldn't be?

Or are the YAML outputs intended to be used as a sort of template for other not-known namespaces?

cenk1cenk2:
Hello @blampe,

Sorry for the late reply, I just wanted to let ArgoCD to set the namespaces to have the opportunity to use features like [ApplicationSet](https://argo-cd.readthedocs.io/en/stable/user-guide/application-set/).

mikhailshilkov:
@cenk1cenk2 @blampe Is there a next step in this PR or should we close it?

PR #3139, opened 2024-07-29
Consolidate updateAwaitConfig/createAwaitConfig

blampe:
There's no functional difference between our "update" and "create" await logic
-- in every case, our update awaiters simply invoke the create awaiter, or vice
versa.

(The only minor exception to this is `untilCoreV1ResourceQuotaUpdated`, which
has a little logic to [short-circuit] but otherwise still calls into the
corresponding create awaiter.)

This PR consolidates the create/update configs into one. This will make it more
straightforward to glue together our legacy awaiters.

The first commit replaces `updateAwaitConfig` with `createAwaitConfig`, and the
second commit is just renaming.

The third commit adds test coverage for the `Update` path, essentially
copy-pasted from the `Creation` tests. This will help guarantee that subsequent
PRs preserve existing behavior.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/2799.

[short-circuit]: https://github.com/pulumi/pulumi-kubernetes/blob/06b05ce7a3fb5358bdc03a1d105c42a6c695363f/provider/pkg/await/awaiters.go#L686-L691

### DIFF
--- a/provider/pkg/await/awaiters.go
+++ b/provider/pkg/await/awaiters.go
@@ -46,7 +46,8 @@ type createAwaitConfig struct {
 	initialAPIVersion string
 	logger            *logging.DedupLogger
 	clientSet         *clients.DynamicClientSet
-	currentOutputs    *unstructured.Unstructured
+	currentOutputs    *unstructured.Unstructured // The result our our create/update.
+	lastOutputs       *unstructured.Unstructured // The state of the object before we changed it. `nil` if this is a create.
 	timeout           *time.Duration
 	clusterVersion    *cluster.ServerVersion
 	clock             clockwork.Clock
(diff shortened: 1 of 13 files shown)


blampe:

- #3133
    - #3139 ◀

blampe:

- #3133
    - #3139 ◀

provider/pkg/await/awaiters.go
```diff
@@ -448,22 +415,13 @@ func untilCoreV1ResourceQuotaInitialized(c createAwaitConfig) error {
 		WatchUntil(rqInitialized, 1*time.Minute)
 }
 
-func untilCoreV1ResourceQuotaUpdated(c updateAwaitConfig) error {
-	oldSpec, _ := openapi.Pluck(c.lastOutputs.Object, "spec")
-	newSpec, _ := openapi.Pluck(c.currentOutputs.Object, "spec")
-	if !reflect.DeepEqual(oldSpec, newSpec) {
-		return untilCoreV1ResourceQuotaInitialized(c.createAwaitConfig)
-	}
```
EronWright, line 456: The old logic seems to say, if nothing has changed, then skip waiting for anything.  But waiting is harmless I suppose?

EronWright (approved):

provider/pkg/await/awaiters.go
blampe: Confirmed this still works as expected. Tested by adding an annotation to a ResourceQuota to exercise this case where the spec is unchanged -- previously we would have short-circuited but awaiting works just as well.

Issue #799, opened 2019-09-12
Unhandled case - Service without selector + manual Endpoint


NinoFloris:
Pulumi tries to await pods for a podless service (external ip)
Minimal repro is something like this
```js
const config = { friendlyName: "test", privateIp: "10.0.0.5" }

export const service = new k8s.core.v1.Service(config.friendlyName, {
    metadata: {
        name: config.friendlyName,
    },
    spec: {
        ports: [{ port: 5432 }],
    },

});

export const endpoints = new k8s.core.v1.Endpoints(config.friendlyName, {
    metadata: {
        // Same name will, by convention, connect the endpoint to the service
        name: service.metadata.name,
    },
    subsets: [{
        addresses: [{ ip: config.privateIp }],
        ports: service.spec.ports
    }]
});
```

Deploying this quickly shows a green tick in gke dashboards but pulumi thinks the resource is unhealthy.

Tested on latest release libs and cli (1.1.0)

Edit:
Adding `"pulumi.com/skipAwait": "true",` doesn't have any effect either.

PR #3142, opened 2024-07-30
Fix endless await for services without selectors/endpoints

blampe:
Our current await logic for services expects the service to have _some_ endpoints, and all of those endpoints must be ready. This doesn't allow for the valid case where the service doesn't have any endpoints, and in those cases we will end up waiting forever.

This changes our logic to instead expect 0 non-ready endpoints. This is equivalent to the current behavior when there are non-zero endpoints while also allowing the zero-endpoint case.

We also short-circuit the endpoint logic when the service has no selector, since the endpoint will need to be manually configured.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/605.
Fixes https://github.com/pulumi/pulumi-kubernetes/issues/799.


provider/pkg/await/service.go
```diff
@@ -318,33 +321,25 @@ func (sia *serviceInitAwaiter) processEndpointEvent(event watch.Event, settledCh
...
-	if event.Type == watch.Added || event.Type == watch.Modified {
-		subsets, hasTargets := openapi.Pluck(endpoint.Object, "subsets")
-		targets, targetsIsSlice := subsets.([]any)
-		endpointTargetsPod := hasTargets && targetsIsSlice && len(targets) > 0
-
-		sia.endpointsReady = endpointTargetsPod
-	} else if event.Type == watch.Deleted {
-		sia.endpointsReady = false
+	var endpoints corev1.Endpoints
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &endpoints)
```
rquitales, line 340: nit: do we need to attempt conversion from an unstructured to typed Endpoint object?

blampe: I went this way since it made it easier to inspect the subsets, but I can understand the benefit to keeping everything as unstructured and version-agnostic as possible. In this situation the Informer is already using `v1/endpoints` so we shouldn't expect to see any other types, but I'm open to doing whatever you think is best.

rquitales: Ah! I didn't see `endpoints` being used further down (`for _, subset := range endpoints.Subsets {`) during the first pass, so I thought we were only just attempting to validate that the unstructured object could bind properly.

rquitales (approved):
Changes looks reasonable to me.

Issue #2996, opened 2024-05-06
Add a default status awaiter


blampe:
For cases where we don't have custom await logic, we should be able to use a default waiter with some reasonable heuristics around status conditions if the resource provides them.

If the resource doesn't have status conditions, we should be able to gracefully degrade to essentially skipAwait behavior, although eventually we will want to support https://github.com/pulumi/pulumi-kubernetes/issues/1260. 

[Internal doc](https://docs.google.com/document/d/1REV7B4ZxVd3ZIrTQGng_rBiRJ44v_J9NvMFAntFZBzs/edit#heading=h.khd4ks8yxo6).

Related to https://github.com/pulumi/pulumi-kubernetes/issues/1260.

PR #3135, opened 2024-07-26
Emit event logs during await

blampe:
This adds a new `EventAggregator` and hooks it up to all of our new awaiters.

This has the effect of surfacing Warning events to the user while any await
logic is ongoing, including all of our existing/legacy awaiters.

Note how straightforward it is to add new/similar aggregators/observers. In the
future we can easily supply other informers, for example we might want a
generic awaiter that aggregates any pods/deployments/stateful sets owned by the
resource.

For example, attempting to unseal an invalid secret will show a temporary warning:

```yaml
  type: kubernetes:yaml/v2:ConfigGroup
    properties:
      yaml: |
        apiVersion: bitnami.com/v1alpha1
        kind: SealedSecret
        metadata:
          name: mysecret
          namespace: default
          annotations:
            pulumi.com/waitFor: condition=Synced
        spec:
          encryptedData:
            foo: AgBy3i4OJSWK+PiTySYZZA9rO43cGDEq...
```
```
PULUMI_K8S_AWAIT_ALL=true pulumi up
```

Interactive:
```
 +      └─ kubernetes:bitnami.com/v1alpha1:SealedSecret  secret:default/mysecret  creating (7s)     warning: [sealedsecret/mysecret] ErrUnsealFailed: Failed to unseal: illegal base64 data at input byte 32 (foo)
```

Non-interactive:
```
 +  kubernetes:bitnami.com/v1alpha1:SealedSecret secret:default/mysecret creating (0s) warning: Has no .status.conditions
 +  kubernetes:bitnami.com/v1alpha1:SealedSecret secret:default/mysecret creating (0s) Resource has condition Synced=False (want True)
 +  kubernetes:bitnami.com/v1alpha1:SealedSecret secret:default/mysecret creating (0s) warning: [sealedsecret/mysecret] ErrUnsealFailed: Failed to unseal: illegal base64 data at input byte 32 (foo)
@ updating............
```

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/133.
Refs https://github.com/pulumi/pulumi-kubernetes/issues/2824.

PR #3143, opened 2024-07-31
Add generic await logic

blampe:
This adds generic await logic that will be feature-flagged behind an
environment variable `PULUMI_K8S_AWAIT_ALL`.

When the environment variable is set, creates and updates will await as
follows:

  1. If the resource has the `skipAwait` annotation we will no-op.
  2. If the resource has pre-existing await logic we will use it.
  3. If the `PULUMI_K8S_AWAIT_ALL` environment variable isn't set, we will no-op.
  4. If the `PULUMI_K8S_AWAIT_ALL` environment variable is set we will
     await according to the cli-utils [status] package.
     
1-3 preserve existing behavior; only 4 is net-new.

Note we're currently using the status package for our DaemonSet awaiter without issue.

Importantly, the package fails open in cases where a resource's readiness can't
be determined. This means resources not following well-known conventions (like
the Ready condition) will continue to no-op instead of waiting forever.
User-defined await logic will be helpful in some of those situations.

TODO:
- [ ] Fix failing unit test.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/2996.

[status]: https://pkg.go.dev/sigs.k8s.io/cli-utils/pkg/kstatus/status


blampe:

- #3133
    - #3139
        - #3143 ◀
            - #3134
            - #3135

tests/sdk/java/await_test.go
```diff
@@ -106,23 +116,207 @@ func TestAwaitSkip(t *testing.T) {
...
+	someField       string
+	conditionType   string
+	conditionStatus string
+}
+
+func TestAwaitGeneric(t *testing.T) {
+	// No t.Parallel because this touches environment variables.
+
+	waitForCRDs := func(events <-chan events.EngineEvent) {
+		// Wait until we see a resource with a "Waiting for readiness message"
```
EronWright, line 163:
```suggestion
		// Wait until we see a resource with a "Waiting for readiness" message
```

tests/sdk/java/testdata/await/generic/crds.yaml
```diff
@@ -0,0 +1,51 @@
...
+          type: object
+          properties:
+            spec:
+              type: object
+              properties:
+                someField:
+                  type: string
+            status:
+              type: object
+              properties:
```
EronWright, line 29: Consider adding an `observedGeneration` property to trigger additional logic within kstatus.

tests/sdk/java/testdata/await/generic/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  namespace: ns.metadata.name,
+});
+
+const crds = new kubernetes.yaml.v2.ConfigFile(
+  "crds",
+  { file: "crds.yaml" },
+  { provider: provider }
+);
+
+export const wantsReady = new kubernetes.apiextensions.CustomResource(
```
EronWright, line 19: If you used nodejs instead of yaml runtime solely to be able to create a custom resource, another option is to use `ConfigGroup` V2 because it also can create custom resources, in yaml.

Oh, seems you have a yaml alternative below, why both?

tests/sdk/java/await_test.go
```diff
@@ -106,23 +116,207 @@ func TestAwaitSkip(t *testing.T) {
...
+				someField:       "untouched",
+				conditionType:   "Ready",
+				conditionStatus: "False",
+			},
+		}
+		assertExpectations(t, outputs, expect)
+	}
+
+	t.Run("enabled", func(t *testing.T) {
+		t.Setenv("PULUMI_K8S_AWAIT_ALL", "true")
```
EronWright, line 252: Should we also set PULUMI_EXPERIMENTAL?

blampe: Discussed and decided to stick with `PULUMI_K8S_AWAIT_ALL`.

EronWright (approved):
This looks good.

One thing I'd consider is making the `PULUMI_K8S_AWAIT_ALL` flag be more consistent, in that it would apply kstatus logic for all resources, including the 'built-in' ones. Why is the built-in logic better than the kstatus logic?

blampe:
> One thing I'd consider is making the PULUMI_K8S_AWAIT_ALL flag be more consistent, in that it would apply kstatus logic for all resources, including the 'built-in' ones. Why is the built-in logic better than the kstatus logic?

For posterity - we don't want to cause any regressions with existing objects (our existing awaiters are quite a bit more complex than the generic one), and we preserve a lot of the rich logging we get currently.

Issue #3063, opened 2024-06-17
Support k8s 1.31 on release


rquitales:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Kubernetes v1.31 release is in progress. The provider should support v1.31 as soon as its available.

Current tag is `alpha.1`: https://github.com/kubernetes/kubernetes/releases/tag/v1.31.0-alpha.1

The final release is scheduled for 13 August 2024 (ref: https://github.com/kubernetes/sig-release/tree/master/releases/release-1.31), with the first release candidate scheduled to be cut on 30 July 2024.

PR #3144, opened 2024-07-31
Upgrade Kubernetes schema and client libraries to v1.31

blampe:
Followed the runbook [here](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/Release%3A%20Native%20Providers.md#kubernetes-supporting-newer-kubernetes-releases).

Quick notes/observations:
* I don't think updating cloud-ready-checks is strictly necessary.
* Would it be helpful to have Dependabot tracking pre-release tags for k8s dependencies? That would give us a PR that takes care of most of the go.mod updates.
* `k8s.io/kube-openapi` seems like our only dependency that's not tagged and needs to take latest.
* Deprecations were added [here](https://github.com/pulumi/pulumi-kubernetes/pull/3144/files#diff-f91667369538bec3685060562982ea254f622ddac92c9be80f32caa2069baadd).

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/3063.


rquitales:
@blampe We'll need to do something similar to https://github.com/pulumi/pulumi-kubernetes/pull/2690/commits/fb32acbbfadf381b7bbb7c0e859a6aa117065cc1 to handle the removed APIs, and potentially update [deprecated.go](https://github.com/pulumi/pulumi-kubernetes/blob/93b4649cb3e7aaf3c05ad57333518a0dbd36069a/provider/pkg/kinds/deprecated.go#L4).

EronWright (approved):
LGTM

Issue #2551, opened 2023-09-06
skipawait per-operation (create, update, delete)


antdking:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

We have a case where deletions of a resource are slow; around 1 hour (GCP/GKE issue, not Pulumi).
Using `skipAwait` means we won't wait fully for creation.

It would be good to be able to have `skipAwait` for deletion only. In our case, it doesn't matter if the resource is left around for a while, as long as it's been marked for termination.

antdking:
details of the problem is that GKE Service Network Endpoint Groups don't cleanup timely, which slows down Namespace deletion beyond a reasonable timeout.

For now, we're applying the `skipAwait` annotation using a `NamespacePatch`, so that we still await on creation, but don't await for deletion. As it's a Namespace, we rarely update the resource, so there's no real downside outside of code overhead.

blampe:
@antdking I'm curious how you would expect the UX to work around this? I can imagine a couple options:

1. Introduce some additional annotations like `skipAwaitCreate`, `skipAwaitDelete`, `skipAwaitUpdate`.
2. Modify the existing annotation to allow specifying specific operations like `"skipAwait": "create,delete"`, in addition to still supporting all operations with `"skipAwait": "true"`.

My preference is probably (2) but I'm curious what you think. Or did you have something else in mind?

antdking:
(2) seems fine given the limitations in k8s metadata.

The main concern I see with it is transformer usecases.

In most cases, they'll just want to append, so it'd be worth allowing strings like "true,delete" or "delete,delete"; where the resource decides to always skip, but the transformer is only concerned about "create" being skipped.

Would it be better to expose as a list of strings in the sdk, or keep it as the comma separated list to keep the sdk aligned with the underlying data?

blampe:
Question for @antdking, @oktocat and others: do we need a distinction between creates and updates, or are we really just interested in something like `skipAwaitDelete` to compliment `skipAwait`?

In other words, is there a use case where you would want `skipAwaitCreate` but _not_ `skipAwaitUpdate`? I can't really think of an instance where that would be useful because we really only care about "readiness," and that doesn't care how we manipulated the manifest.

[...]

PR #3147, opened 2024-08-01
Allow skipAwait to skip only readiness or deletion

blampe:
Currently, the `pulumi.com/skipAwait: true` annotation tells the provider to
skip await logic whenever a resource is touched _or_ deleted.

However, there are situations where you might want to skip await logic for only
deletes, or only for readiness, but not both. For example a resource might take
a long time to tear down, during which time other resources could also be
getting cleaned up.

This PR expands the acceptable values for the `skipAwait` annotation:
* `"pulumi.com/skipAwait": "true"` continues to skip await logic in all situations.
* `"pulumi.com/skipAwait": "ready"` will skip await logic when creating or
  updating a resource, when we would normally wait for it to become ready. We
  will still wait for deletion.
* `"pulumi.com/skipAwait": "delete"` will skip await logic when deleting the
  resource, but we will continue to wait for it become ready when creating or
  updating it.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/2551.


blampe:

- #3133
    - #3139
        - #3143
            - #3147 ◀

provider/pkg/metadata/annotations.go
```diff
@@ -162,7 +165,10 @@ func GetReadyCondition(
 // SkipReadyCondition returns true if the inputs are annotated such that we
 // should not await readiness.
 func SkipReadyCondition(inputs *unstructured.Unstructured) bool {
-	return IsAnnotationTrue(inputs, AnnotationSkipAwait)
+	if IsAnnotationTrue(inputs, AnnotationSkipAwait) {
+		return true
+	}
+	return strings.EqualFold(GetAnnotationValue(inputs, AnnotationSkipAwait), "ready")
```
EronWright, line 171: I would advocate for parsing the value as a comma-separated string, and then checking for the enum values of true, ready, delete.  This would allow for a user to do `ready,delete` and for us to contemplate treating `true` as `ready`.

EronWright (commented):
I'm pretty reluctant to skip await on deletion unless the user has explicitly opted in with `skipAwait: delete`. Because skipping await on deletion can lead to a race condition in the 'replacement' use case.

I think the safe move is to interpret the value of `true` as `ready`, not as `ready,delete`.  Since that is a breaking change, another possibility would be to interpret `true` differently depending on the kind, to retain existing behavior.  I'm still surprised that skipAwait ever applied to deletes.

blampe:
On pause https://github.com/pulumi/pulumi-kubernetes/issues/2551#issuecomment-2293964331

PR #3159, opened 2024-08-08
Add functionality for non-ephemeral de-duped logs

blampe:
First commit:
* Add a `Log` method to `DedupLogger` to enable emitting persistent logs.
* Modify `LogMessage` to take a severity/string instead of a `checkerlog.Message`. Depending directly on cloud-ready-checks like this is awkward, and we're going have check messages coming from other sources like kstatus so it doesn't make sense to couple these. 
* Remove log methods on `createAwaitConfig` since they were just mucking with `checkerlog.Message`.
* Rename `LogMessage` → `LogStatus` to make it clearer that these are ephemeral/status logs.
* Add a smoke test.

Second commit:
* Update `logMessage` and `logStatus` call sites to use the logger directly.

### DIFF
--- a/provider/pkg/await/awaiters.go
+++ b/provider/pkg/await/awaiters.go
@@ -21,7 +21,6 @@ import (
 	"time"
 
 	"github.com/jonboulle/clockwork"
-	checkerlog "github.com/pulumi/cloud-ready-checks/pkg/checker/logging"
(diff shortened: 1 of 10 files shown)


rquitales (approved):
Changes LGTM

PR #3186, opened 2024-08-22
Make v2 YAML more forgiving when checking GVK for namespaced-ness

### LINKED ISSUE #3176: failed to determine if the following GVK is namespaced if CRD is created in the same run
### What happened?

I'm deploying cert-manager helm release + a ClusterIssuer resource

cert-manager helm release creates the ClusterIssuer CRD, but it seems that the `pulumi.ConfigGroup` that creates the `ClusterIssuer` resources tries to check some properties of the `ClusterIssuer` CRD before it's actually created although there is `depends_on` to the cert-manager.  

```
    Exception: marshaling properties: awaiting input property "resources": failed to determine if the following GVK is namespaced: cert-manager.io/v1, Kind=ClusterIssuer

``` 

If I rerun `pulumi up -y --skip-preview` after then the `ClusterIssuer` will be created fine. That's why I think it's a timing issue between the creation of the `ClusterIssuer` CRD and the actual `ClusterIssuer` resource. 

### Example

```
cert_manager = helmv3.Release(
    "cert_manager",
    helmv3.ReleaseArgs(
        # https://cert-manager.io/docs/installation/helm/
        # https://artifacthub.io/packages/helm/cert-manager/cert-manager
        # https://github.com/cert-manager/cert-manager
        name="cert-manager",
        chart="cert-manager",
        namespace=namespace.id,
        version="1.15.2",
        repository_opts=helmv3.RepositoryOptsArgs(
            repo="https://charts.jetstack.io",
        ),
        values=cert_manager_helm_values,
    ),
    opts=pulumi.ResourceOptions(
        provider=kubernetes_provider,
    ),
)


def generate_clusterissuer_manifest(name, server):
    def func(args):
        template = env.get_template("letsencrypt-clusterissuer.j2.yaml")
        rendered = template.render(
            domain_name=args["domain_name"],
            region=args["region"],
            zone_id=args["zone_id"],
            name=name,
            server=server,
        )
        return rendered

    return pulumi.Output.all(
        domain_name=domain_name,
        region=region,
        zone_id=zone.id,
    ).apply(func)


# https://www.pulumi.com/registry/packages/kubernetes/api-docs/yaml/configgroup/
letsencrypt_stagin_cluster_issuer_cg = kubernetes.yaml.v2.ConfigGroup(
    "letsencrypt-staging",
    yaml=generate_clusterissuer_manifest(
        name="letsencrypt-staging",
        server="https://acme-staging-v02.api.letsencrypt.org/directory",
    ),
    opts=pulumi.ResourceOptions(
        depends_on=[
            cert_manager,
        ],
        provider=kubernetes_provider,
    ),
)


```

### Output of `pulumi about`

```
pulumi about
CLI          
Version      3.129.0
Go Version   go1.22.6
Go Compiler  gc

Plugins
KIND      NAME        VERSION
resource  aws         6.49.1
resource  eks         2.7.8
resource  kubernetes  4.17.1
language  python      unknown
resource  random      4.16.3

Host     
OS       darwin
Version  14.6.1
Arch     x86_64

This project is written in python: executable='/Users/xxx/git/pulumi-aws-ecerulm/venv/bin/python' version='3.12.5'


...

Backend        
Name           xxxxx
URL            file://~
User           xxxx
Organizations  
Token type     personal

Dependencies:
NAME           VERSION
Jinja2         3.1.4
pip            24.2
pulumi_eks     2.7.8
pulumi_random  4.16.3
setuptools     72.2.0
wheel          0.44.0

```

### Additional context

_No response_

### Contributing

blampe:
This changes our error handling in `Normalize` to degrade gracefully in the case when we can't determine namespaced-ness, probably due to the CRD not existing yet. Instead of failing, we assume the resource is namespaced to allow the preview to succeed.

This is consistent with our error handling for this case elsewhere:
* Check https://github.com/pulumi/pulumi-kubernetes/blob/0f834c8b0d89e0003f0dc2d527d4ca8e2cde26e9/provider/pkg/provider/provider.go#L1481-L1488
* Diff https://github.com/pulumi/pulumi-kubernetes/blob/0f834c8b0d89e0003f0dc2d527d4ca8e2cde26e9/provider/pkg/provider/provider.go#L1666-L1674
* Invoke: https://github.com/pulumi/pulumi-kubernetes/blob/0f834c8b0d89e0003f0dc2d527d4ca8e2cde26e9/provider/pkg/provider/invoke_decode_yaml.go#L49-L56

Added a failing unit test.

Fixes #3176

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,10 @@
 
 - Fixed a panic that could occur during deletion. (https://github.com/pulumi/pulumi-kubernetes/issues/3157)
 
+- Fixed an error that could occur when previewing `yaml/v2` resources which
+  depend on CRDs created in the same update.
+  (https://github.com/pulumi/pulumi-kubernetes/issues/3176)
+
 ## 4.17.1 (August 16, 2024)
 
 ### Fixed
--- a/provider/pkg/provider/yaml/v2/yaml.go
+++ b/provider/pkg/provider/yaml/v2/yaml.go
@@ -167,7 +167,14 @@ func Normalize(objs []unstructured.Unstructured, defaultNamespace string, client
 		// determine whether the kind is namespaced, which may involve a call to the API server.
 		// note that the object list is searched for a matching CRD before resorting to a discovery call.
 		isNamespaced, err := clients.IsNamespacedKind(gvk, clientSet, objs...)
-		if err != nil {
+		if clients.IsNoNamespaceInfoErr(err) {
+			// This can happen during previews, or when running concurrent with
+			// a Release which doesn't give us visibility into the CRDs it will
+			// be creating. Since we can't tell for sure at this point, assume
+			// it is namespaced, and correct if required during the Create
+			// step.
+			isNamespaced = true
+		} else if err != nil {
 			return nil, err
 		}
 		if isNamespaced && obj.GetNamespace() == "" {
--- a/provider/pkg/provider/yaml/v2/yaml_test.go
+++ b/provider/pkg/provider/yaml/v2/yaml_test.go
@@ -553,6 +553,28 @@ var _ = Describe("Normalize", func() {
 				))
 			})
 		})
+
+		Context("when the object is an unrecognized custom resource", func() {
+			BeforeEach(func() {
+				objs = []unstructured.Unstructured{{
+					Object: map[string]any{
+						"apiVersion": "test.pulumi.com/v1",
+						"kind":       "Unknown",
+						"metadata": map[string]any{
+							"name": "foo",
+						},
+					},
+				}}
+			})
+
+			It("should assume it is namespaced", func(ctx context.Context) {
+				objs, err := Normalize(objs, defaultNamespace, clientSet)
+				Expect(err).ShouldNot(HaveOccurred())
+				Expect(objs).To(HaveExactElements(
+					matchUnstructured(Keys{"metadata": MatchKeys(IgnoreExtras, Keys{"namespace": Equal("default")})}),
+				))
+			})
+		})
 	})
 
 	Describe("special-case kinds", func() {


EronWright (commented):
I considered this approach during the development of the MLCs. Here were my concerns:
1. the resource URN would be volatile depending on the actual cluster state, since it includes the namespace.
2. a wrong guess about the resource scope (namespaced-ness) would cause noise during preview, depicting a replacement of the child.
3. transformation logic would see unexpected resource names.

Note that the logic in `Check` doesn't affect the resource name, so it is safer with respect to these concerns.

What else can we do? My proposal is that `Release` resource add its CRDs to the cache that's within the dynamic client, similarly to what happens to a CRD resource during Check:
https://github.com/pulumi/pulumi-kubernetes/blob/2cc176d54996c1293525436f3970f1538ab7036f/provider/pkg/provider/provider.go#L1517-L1524

Note that this problem happens for `Release` but not for `Chart` because `Chart` creates child resources such as `CustomResourceDefinition`, even during preview, for which `Check` is called (thus populating the cache).

A Helm chart actually has two ways of providing a CRD, and the `Check` method of `Release` would need to understand both to be a comprehensive solution.  One is described [here](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you), and would be about reading the `crds/` directory within the chart. The other way is the "crd-install" way where the CRDs are templated as ordinary resources, and this means that `Check` would need to execute the template to obtain any CRD(s) (see https://github.com/pulumi/pulumi-kubernetes/pull/2568). Unfortunately, the venerable cert-manager chart is in the latter category.

EronWright:
@blampe I'm closing this PR in favor of different solution(s), e.g.
- https://github.com/pulumi/pulumi-kubernetes/issues/3299
- https://github.com/pulumi/pulumi-kubernetes/issues/3300

PR #3205, opened 2024-09-14
fix: properly determine list objects in schemagen

rquitales:
CRDs can have types that end in `List`. These are not actually List resources, so we should not rely on name alone to determine this. Instead, we should follow the rules set out in the [conventions guide](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds) to determine is a Kind is meant to be a List kind.

### DIFF
--- a/provider/pkg/gen/testdata/identify-list-kinds/list-kind
+++ b/provider/pkg/gen/testdata/identify-list-kinds/list-kind
@@ -1,0 +1,153 @@
+-- definitions --
+{
+  "com.pulumi.k8sversion.test.TestResource": {
+    "properties": {
+      "apiVersion": {
+        "description": "APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources",
+        "type": "string"
+      },
+      "kind": {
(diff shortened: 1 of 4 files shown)


blampe (approved):
Deja vu :P

blampe (approved):
Is it feasible to test this?

rquitales:
> Is it feasible to test this?

Yes, I'll add to this PR next week!

provider/pkg/gen/typegen.go
```diff
@@ -693,6 +705,14 @@ func createGroups(definitionsJSON map[string]any, allowHyphens bool) []GroupConf
 
 				return fmt.Sprintf("%s/%s", gStripped, v)
 			}
+
+			// These resources are hard-coded as lists as they do not adhere to the normal conventions.
```
rquitales, line 709: The list is nested under the `resources` and `groups` field instead of `items` for these two kinds.

provider/pkg/gen/testdata/identify-list-kinds/list-kind
```diff
@@ -0,0 +1,153 @@
...
+        "version": "test"
+      }
+    ],
+    "x-kubernetes-selectable-fields": []
+  }
+}
+
+-- kinds --
+- k8sversion.test.TestResource
+- k8sversion.test.TestResourceNotRealList
```
rquitales, line 149: A custom resource that ends in `List`, but is not a List kind.

blampe (approved):

PR #3210, opened 2024-09-16
fix: correctly set pulumi-kubernetes python dependency

rquitales:
### Proposed changes

When setting p-k as a python dependency for schemagen, we also need to include the `==` token to generate a valid pyproject.toml file.

### Related issues (optional)

Related: https://github.com/pulumi/crd2pulumi/issues/148


blampe (approved):
Ugh!

PR #3213, opened 2024-09-18
Fix Typo in Kubernetes API Docs

nelsonfigueroa:
### Proposed changes
This is a simple typo fix to change `is` to `it's`. There's also some linting included that removes trailing spaces and newlines, but I can remove these if that should be a separate PR.

### DIFF
--- a/sdk/python/pulumi_kubernetes/helm/v4/_inputs.py
+++ b/sdk/python/pulumi_kubernetes/helm/v4/_inputs.py
@@ -101,7 +101,7 @@ if not MYPY:
         """
         repo: NotRequired[pulumi.Input[str]]
         """
-        Repository where to locate the requested chart. If is a URL the chart is installed without installing the repository.
+        Repository where to locate the requested chart. If it's a URL the chart is installed without installing the repository.
         """
         username: NotRequired[pulumi.Input[str]]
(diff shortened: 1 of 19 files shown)


sdk/python/pulumi_kubernetes/helm/v4/_inputs.py
```diff
@@ -212,5 +212,3 @@ def username(self) -> Optional[pulumi.Input[str]]:
     @username.setter
     def username(self, value: Optional[pulumi.Input[str]]):
         pulumi.set(self, "username", value)
-
```
arturl, line 215: I suggest removing non-essential whitespace changes since your fix is only about the typo.
(In some cases trailing empty lines can event be meaningful!)

arturl (approved):

sdk/python/pulumi_kubernetes/helm/v4/_inputs.py
blampe: This is generated code, you can ignore it.

blampe (approved):
Thank you!

blampe:
@nelsonfigueroa heads up this was merged with failing tests - #3223.

Issue #3227, opened 2024-09-25
[schemagen] Parsing group from swagger definition is not specific enough


rquitales:
### What happened?

We need to expand the stored group in: https://github.com/pulumi/pulumi-kubernetes/blob/8a0d0c307e7031f66b3c837262438cf951c1394d/provider/pkg/gen/typegen.go#L263

This should be updated to something like: `strings.Join(split[:len(split)-2], ".")`.

When there are multiple resources with similar group names through their swagger definition paths, eg `com.pulumi.k8s` and `io.test.k8s`, the current logic will cause the canonical groups to be stored under the `k8s` key. Instead, we should expand our key to be more specific.

### Example

Identified in https://github.com/pulumi/crd2pulumi/issues/152
Multiple CRDs generate types tokens with the incorrect group due to how we map group->canonicalGroup

### Output of `pulumi about`

-

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #3237, opened 2024-09-30
Fix schemagen canonical group determination to be more specific

rquitales:
### Proposed Changes
This PR includes a major refactor of the `createGroups` function in the `gen` package and addresses a bug that could lead to map key collisions when storing groups derived from swagger paths. Previously, consumers of the `gen` package could encounter key collisions when different CRDs used the same key. For instance, both `io.k8s.api.apps` and `com.testpackage.apps` would result in the key `apps`. This PR resolves this by using the full group name (e.g., `io.k8s.api.apps`) as the map key instead of just the last segment. Since no such collisions exist in the standard Kubernetes GVKs, no schema changes are expected.

To support these changes, the `typegen.go` file has been significantly refactored. The `createGroups` function has been modularized for easier testing and maintainability. Additionally, the `go-linq` dependency has been removed in favor of idiomatic Go code, allowing for static type checking and reducing runtime type errors.

### Changes Made:

- Refactored `createGroups` into modular subfunctions
- Increased test coverage of the `gen` package from 36.8% to 49.6% with new tests
- Replaced `github.com/ahmetb/go-linq` with idiomatic Go for better maintainability and static type checking
- Fixed map key creation to use the full group name, preventing collisions, with no schema changes required in the p-k provider

### Related Issues (optional)

Fixes: #3227 and resolves https://github.com/pulumi/crd2pulumi/issues/152

### DIFF
--- a/provider/pkg/gen/typegen.go
+++ b/provider/pkg/gen/typegen.go
@@ -19,9 +19,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
-	"github.com/ahmetb/go-linq"
 	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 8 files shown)


provider/pkg/gen/typegen.go
```diff
@@ -260,7 +260,7 @@ func GVKFromRef(ref string) schema.GroupVersionKind {
 	gvk := schema.GroupVersionKind{
 		Kind:    split[len(split)-1],
 		Version: split[len(split)-2],
-		Group:   split[len(split)-3],
+		Group:   strings.Join(split[:len(split)-2], "."),
```
rquitales, line 263: This is the main behavior change. We now store `io.k8s.api.apps` as the key in the `canonicalGroup` map. Previously, this was `apps`. This is an issue for downstream consumers of this package since `io.k8s.api.apps` and `com.mytestpackage.apps` would have collided, and resulted in incorrectly generated package names.

provider/pkg/gen/typegen.go
```diff
@@ -463,339 +463,360 @@ func makeSchemaType(prop map[string]any, canonicalGroups map[string]string) stri
 // --------------------------------------------------------------------------
 
 func createGroups(definitionsJSON map[string]any, allowHyphens bool) []GroupConfig {
```
rquitales, line 465: This function was refactored by splitting it up into multiple sub-functions to make it easier to unit test each sub-action. This made it easier to track down what changes were required so that schema generation in p-k was ultimately not modified.

provider/pkg/gen/typegen.go
```diff
@@ -19,9 +19,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
-	"github.com/ahmetb/go-linq"
```
rquitales, line 24: Removing `go-linq` as a dependency and using idiomatic golang for data processing instead.

blampe (approved):
> Additionally, the go-linq dependency has been removed in favor of idiomatic Go code, allowing for static type checking and reducing runtime type errors.

🙌 🙌 🙌 

Amazing, thank you! Looks like the linq stuff was mostly glue and the underlying logic is essentially untouched 👍

PR #3239, opened 2024-10-01
Do not specify pulumi sdk

t0yv0:
### Proposed changes

Re: https://github.com/pulumi/ci-mgmt/issues/1091

Since the original dependency required higher SDK versions than 3.x.x., the remediation also upgrades SDK generators to latest.


provider/pkg/gen/schema.go
```diff
@@ -656,7 +654,6 @@ Use the navigation below to see detailed documentation for each of the supported
 	})
 
 	requires := map[string]string{
-		"pulumi":   ">=3.109.0,<4.0.0",
```
t0yv0, line 659: Was 109.0 highly intentional here? Are we still OK here?

sdk/python/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
   name = "pulumi_kubernetes"
   description = "A Pulumi package for creating and managing Kubernetes resources."
-  dependencies = ["parver>=0.2.1", "pulumi>=3.109.0,<4.0.0", "requests>=2.21,<3.0", "semver>=2.8.1", "typing-extensions>=4.11; python_version < \"3.11\""]
+  dependencies = ["parver>=0.2.1", "pulumi>=3.134.1,<4.0.0", "requests>=2.21,<3.0", "semver>=2.8.1", "typing-extensions>=4.11; python_version < \"3.11\""]
```
t0yv0, line 4: Moving up it seems 🤞

provider/pkg/gen/schema.go
justinvp: It was intentional to make use of a change in `>=3.109.0` of the core SDK, but this change LGTM as it's now `>=3.134.1` as desired.

justinvp (approved):
LGTM

provider/go.mod
```diff
@@ -21,8 +21,8 @@ require (
 	github.com/pulumi/cloud-ready-checks v1.1.1-0.20240731201114-3a703c6bee71
 	github.com/pulumi/pulumi-kubernetes/sdk/v4 v4.0.0
 	github.com/pulumi/pulumi-kubernetes/tests/v4 v4.0.0-20240302002028-652829a1ed71
-	github.com/pulumi/pulumi/pkg/v3 v3.134.1
-	github.com/pulumi/pulumi/sdk/v3 v3.134.1
+	github.com/pulumi/pulumi/pkg/v3 v3.135.0
```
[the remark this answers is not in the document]
justinvp: Not much changed in sdkgen between v3.134.1 and v3.135.0, other than the min default version and that `_utilities.py` change.

t0yv0:
I need to tidy all the modules, one sec.

blampe:
I'm confused by the expected usage of the [Language](https://github.com/pulumi/pulumi/blob/pkg/v3.135.0/pkg/codegen/schema/schema.go#L2048-L2049) parameter. Are we now saying minimum versions should now longer specified by the provider, so codegen handles everything? Or should we omit minimums only for Python and Node? Whatever it is, let's please document the expected usage 🙏 

FWIW I think it would make more sense for codegen to compute a minimum instead of ignoring the provider's input, but this is not important now.

t0yv0:
Apologies a bit trigger-happy with automerge here, feel free to back out. Not releasing yet though @blampe  - when you are comfortable the recommendation is to release, but need to work through concerns first.

blampe:
> Apologies a bit trigger-happy with automerge here, feel free to back out. Not releasing yet though @blampe - when you are comfortable the recommendation is to release, but need to work through concerns first.

No objections! We can release whenever. I just find this bag of values very opaque considering how important it is.

PR #3275, opened 2024-10-19
Fix SSA conflict when updating old objects

blampe:
Refs https://github.com/pulumi/customer-support/issues/1837

This adds unit tests around:
* SSA updates for objects with existing field managers.
* SSA updates for old/legacy objects with no field managers.

This currently forces the apply when the object has no managers. There's also some code showing where we could impersonate `kubectl` SSA to trigger a server-side upgrade of these fields.

Worth noting `client-go` recently added support for `Apply` but only in typed clients. I use a wrapper around this to test our untyped `ssaUpdate` path.


provider/pkg/await/await.go
```diff
@@ -536,13 +536,22 @@ func csaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dy
...
+
 // ssaUpdate handles the logic for updating a resource using server-side apply.
-func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client dynamic.ResourceInterface) (*unstructured.Unstructured, error) {
+func ssaUpdate(c *UpdateConfig, liveOldObj *unstructured.Unstructured, client patcher) (*unstructured.Unstructured, error) {
 	liveOldObj, err := fixCSAFieldManagers(c, liveOldObj, client)
 	if err != nil {
 		return nil, err
 	}
 
+	fieldManager := c.FieldManager
```
rquitales, line 550: Looks like this is deadcode from testing? Let's drop the `fieldManager` changes in this function.

blampe: Yeah, this is showing where we could impersonate `kubectl` if we wanted to go down that route, but it seems like we're OK forcing the patch.

rquitales (approved):

EronWright (approved):

PR #3285, opened 2024-10-24
[helm.v4.Chart] Proposal: option to deploy helm hooks

### LINKED ISSUE #3284: helm.v4.Chart: make helm hooks optional (instead of force disabled)
## Hello!

## Issue details

`helm.v3.Chart` ignored the `helm.sh/hook*` annotations but still deployed the resources.
The `helm.v4.Chart` resource skips the hooks. This is mentioned [in the blog](https://www.pulumi.com/blog/kubernetes-chart-v4/#not-supported-helm-hooks) but not in the docs BTW.

I think users should have a way to make v4.Chart deploy hooked resources like v3 did, because it's useful and becasue it makes migrating from v3 to v4 less painful.

awoimbee:
### Proposed changes

`helm.v3.Chart` ignored the `helm.sh/hook*` annotations but still deployed the resources.
The `helm.v4.Chart` resource skips the hooks. This is mentioned [in the blog](https://www.pulumi.com/blog/kubernetes-chart-v4/#not-supported-helm-hooks) but not in the docs BTW.

In this PR I propose to make it configurable so a pulumi user can decide to deploy resources, ignoring the hooks.
The end goal being that helm.v4.Chart comes to feature parity with helm.v3.Chart so we can all move all our charts to v4.

Note that:
* I don't known what I'm doing here, feel free to take over !
* `CONTRIBUTING.md` tells me to install stuff in `/opt` which is not happening, so I did not run the tests locally

### Related issues (optional)

Fixes #3284

### DIFF
--- a/provider/pkg/gen/examples/overlays/chartV4.md
+++ b/provider/pkg/gen/examples/overlays/chartV4.md
@@ -56,7 +56,9 @@ Use the `dependencyUpdate` input to have Pulumi update the dependencies (see: [H
 The `Chart` resource renders the templates from your chart and then manages the resources directly with the
 Pulumi Kubernetes provider. A default namespace is applied based on the `namespace` input, the provider's
 configured namespace, and the active Kubernetes context. Use the `skipCrds` option to skip installing the
-Custom Resource Definition (CRD) objects located in the chart's `crds/` special directory.
+Custom Resource Definition (CRD) objects located in the chart's `crds/` special directory. By default,
+resources managed by helm hooks are ignored, use `deployHookedResources` to deploy them as regular resources,
+ignoring their `helm.sh/hook*` annotations.
 
 Use the `postRenderer` input to pipe the rendered manifest through a [post-rendering command](https://helm.sh/docs/topics/advanced/#post-rendering).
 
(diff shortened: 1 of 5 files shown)


awoimbee:
@EronWright you're the only one in the blame of `provider/pkg/provider/helm/v4/chart.go`, WDYT ?

Would you accept it ? And if so do you want me to iterate on it or someone @ pulumi ?

EronWright (changes requested):
I outlined [here](https://github.com/pulumi/pulumi-kubernetes/issues/3284#issuecomment-2450551684) the technical challenges in truly supporting hook resources. I believe it is worse to implement hooks wrongly than to not support them at all.  We have the `Release` resource for cases where chart compatbility is an issue.

Makeshift:
Often, hooks are added to charts for a bit of extra quality-of-life and are not actually required for the charts to install correctly.

For example, I just tried to install the ArgoCD chart with chart v4 - it failed because there's a hook on the [redis secret init job](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/redis-secret-init/job.yaml), so Pulumi ignored the job entirely, resulting in the redis pod spinning forever and the deployment eventually timing out. Had Pulumi just treated _the annotation_ like _an annotation_, it would have been fine.

It doesn't even look like there's a way to use transforms with v4 to remove the hooks, so you just... can't deploy this chart with v4, I guess. Time to write a post-renderer, yay...

PR #3309, opened 2024-11-11
Fix tests

blampe:
The `nginx:1.7.9` image is 9-years-old and we've recently had issues pulling it.

Tests on master time out after this:
```
 +  kubernetes:core/v1:Pod nginx creating (1s) warning: Failed: Failed to pull image "nginx:1.7.9": failed to pull and unpack image "docker.io/library/nginx:1.7.9": no signatures
 +  kubernetes:core/v1:Pod nginx creating (1s) warning: Failed: Error: ErrImagePull
@ updating....
 +  kubernetes:core/v1:Pod nginx creating (2s) warning: Failed: Error: ImagePullBackOff
```

Even locally I'm not able to pull it:
```
❯ docker pull nginx:1.7.9
Error response from daemon: missing signature key
```

Various support articles suggest this is due to an older version of Docker, but I'm running v27.

Rather than attempt to debug this further, let's update these tests to use a more recent image which doesn't have this problem.

Update: the pull error was transient and probably had something to do with this:

> [DEPRECATION NOTICE] Docker Image Format v1 and Docker Image manifest version 2, schema 1 support is disabled by default and will be removed in an upcoming release. Suggest the author of docker.io/library/nginx:1.7.9 to upgrade the image to the OCI Format or Docker Image manifest v2, schema 2. More information at https://docs.docker.com/go/deprecated-image-specs/

### DIFF
--- a/tests/sdk/nodejs/examples/provider/index.ts
+++ b/tests/sdk/nodejs/examples/provider/index.ts
@@ -14,7 +14,7 @@ const myk8s2 = new k8s.Provider("myk8s2", {
 const nginxcontainer = new k8s.core.v1.Pod("nginx", {
     spec: {
         containers: [{
-            image: "nginx:1.7.9",
+            image: "nginx:1.27.2",
             name: "nginx",
             ports: [{ containerPort: 80 }],
         }],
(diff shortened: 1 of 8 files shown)


EronWright (approved):
LGTM and thanks, I too hit upon this issue.

Issue #3329, opened 2024-11-25
StatefulSet update doesn't wait for pods to initialize


daniel-wtfoxtrot:
### What happened?

I experience strange behaviour in our Deployments, when updating StatefulSets. The initial `pulumi up` behaves as expected, everything is deployed and running fine. However when I run a subsequent `pulumi up` to change the StatefulSets Pod's Image, the change of the StatefulSet is immediately considered successful, without waiting for any of the new updated Pods to start.

```
@ updating....
 ~  kubernetes:apps/v1:StatefulSet foo-store updating (0s) [diff: ~spec]; [1/2] Waiting for StatefulSet to create Pods (0/0 Pods ready)
 ~  kubernetes:apps/v1:StatefulSet foo-store updating (0s) [diff: ~spec]; StatefulSet initialization complete
 ~  kubernetes:apps/v1:StatefulSet foo-store updating (0s) [diff: ~spec]; 
 ~  kubernetes:apps/v1:StatefulSet foo-store updated (0.85s) [diff: ~spec]; 
```

At this point new Pods are still pending or haven't even been started, but the StatefulSet update is already considered successful.

### Example

1. Deploy this StatefulSet
```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: foo-store
  namespace: stateful-set-issue
  labels:
    app: foo-store
spec:
  replicas: 2
  selector:
    matchLabels:
      app: foo-store
  template:
    metadata:
      labels:
        app: foo-store
    spec:
      containers:
        - name: foo-store-web
          image: registry.k8s.io/e2e-test-images/agnhost:2.40
          resources:
            limits:
              memory: 2Gi
            requests:
              memory: 1Gi
          readinessProbe:
            httpGet:
              path: /healthz
              port: 8080
            initialDelaySeconds: 5
            timeoutSeconds: 3
            periodSeconds: 5
            successThreshold: 1
            failureThreshold: 20
          imagePullPolicy: Always
      restartPolicy: Always
      terminationGracePeriodSeconds: 30
      automountServiceAccountToken: false
      imagePullSecrets:
        - name: artifactory
      affinity:
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
            - weight: 80
              podAffinityTerm:
                labelSelector:
                  matchLabels:
                    app: foo-store
                topologyKey: kubernetes.io/hostname
      schedulerName: default-scheduler
  serviceName: foo-store
[...]

PR #3344, opened 2024-12-03
Fix await annotation precedence

blampe:
Currently when creating or updating, custom await annotations (including `waitFor` and `skipAwait`) will be ignored for resources which have built-in await logic.

This corrects the behavior by returning a boolean from `metadata.ReadyCondition` indicating whether the user specified some behavior via an annotation. If that's true then we don't use the built-in logic.

We preserve the existing precedence when `PULUMI_K8S_AWAIT_ALL` is `true`. That is, we will continue to use built-in await logic (if the resource has it) by default, even when generic await is enabled.

During testing I also realized we have a bug where if a JSONPath is specified pointing to a non-primitive type (like `.metadata`) it will never resolve. I fixed this to match correctly.

Refs https://github.com/pulumi/pulumi-kubernetes/issues/3329
Fixes https://github.com/pulumi/pulumi-kubernetes/issues/3345

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 - `pulumi.com/waitFor` and other await annotations now correctly take precedence over default await logic.
   (https://github.com/pulumi/pulumi-kubernetes/issues/3329)
 
+- JSONPath expressions used with the `pulumi.com/waitFor` annotation will no longer hang indefinitely if they match non-primitive fields.
+  (https://github.com/pulumi/pulumi-kubernetes/issues/3345)
+
 ## 4.18.3 (October 31, 2024)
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 - [nodejs] Resolves `punycode` deprecation warnings by using native `fetch` instead of `node-fetch`.
   (https://github.com/pulumi/pulumi-kubernetes/issues/3301)
 
+### Fixed
+
```
rquitales, line 9: Should we also add a note about jsonpath matching on complex types?

rquitales (approved):

Issue #3325, opened 2024-11-22
Issue with x_kubernetes_preserve_unknown_fields in  java com.pulumi.kubernetes.apiextensions.v1.CustomResourceDefinition


elonazoulay:
### What happened?

I am attempting to create a CustomResourceDefinition that has a field with `x-kubernetes-preserve-unknown-fields` set to true via the `x_kubernetes_preserve_unknown_fields` method and expected the resource to be created but instead get the following error:
```
Caused by: java.lang.IllegalArgumentException: Expected type 'com.pulumi.kubernetes.apiextensions.v1.outputs.JSONSchemaProps' (annotated with 'com.pulumi.core.annotations.CustomType') to have a setter annotated with @com.pulumi.core.annotations.CustomType$Setter("x-kubernetes-preserve-unknown-fields"), got: $schema,nullable,x_kubernetes_int_or_string,minLength,pattern,description,type,title,required,example,exclusiveMaximum,patternProperties,allOf,not,default,oneOf,additionalItems,x_kubernetes_embedded_resource,id,maxProperties,exclusiveMinimum,x_kubernetes_validations,definitions,multipleOf,maxItems,x_kubernetes_list_type,format,anyOf,enum,minProperties,dependencies,minItems,x_kubernetes_preserve_unknown_fields,x_kubernetes_list_map_keys,uniqueItems,maximum,additionalProperties,x_kubernetes_map_type,externalDocs,minimum,$ref,items,maxLength,properties
    	at com.pulumi.serialization.internal.Converter.lambda$tryConvertObjectInner$7(Converter.java:401)
    	at java.base/java.util.HashMap.forEach(HashMap.java:1429)
    	at com.pulumi.serialization.internal.Converter.tryConvertObjectInner(Converter.java:399)
    	at com.pulumi.serialization.internal.Converter.tryConvertMap(Converter.java:665)
    	at com.pulumi.serialization.internal.Converter.tryConvertObjectInner(Converter.java:281)
    	at com.pulumi.serialization.internal.Converter.lambda$tryConvertObjectInner$7(Converter.java:426)
    	... 34 more
```
If I remove the fields in the example below that contain `x_kubernetes_preserve_unknown_fields` it works.

### Example

This is the removed portion of the spec:
```
.put("coordinator", JSONSchemaPropsArgs.builder()
                                                                                        .type("object")
                                                                                        .properties(ImmutableMap.<String, JSONSchemaPropsArgs>builder()
[...]

PR #3348, opened 2024-12-04
Normalize `x-kubernetes-*` fields for outputs

rquitales:
### Proposed changes

This PR normalizes outputs so that any `x-kubernetes-*` fields are stored in their respective underscored forms. When creating a CRD with Pulumi, we are able to set the `x-kubernetes-*` fields correctly during create/update (except in Java). However, if we attempt to read this field as an output, it would fail since the stored state is not normalized to the underscored forms. In other languages (Go/Python/TS), this was fine - but in Java, this fails as the builder pattern we use for codegen is unable to set the `x-kubernetes-*` fields since we only have setters for the underscored versions.

This also fixes the issue where we cannot read the values of `x_kubernetes_*` fields from outputs.

This PR adds additional unit tests, expands on the existing nodejs e2e test to verify that we can now read `x_kubernetes_*` fields, and creates a new Java e2e test to ensure we can successfully create CRDs that contain `x_kubernetes_*` fields.

### Related issues (optional)

Closes: #3325


blampe:
> This PR normalizes outputs so that any x-kubernetes-* fields are stored in their respective underscored forms

To clarify -- we currently store dashes in state, and will continue to store dashes, right? So underscores are only ever exposed to the SDK?

rquitales:
> > This PR normalizes outputs so that any x-kubernetes-* fields are stored in their respective underscored forms
> 
> To clarify -- we currently store dashes in state, and will continue to store dashes, right? So underscores are only ever exposed to the SDK?

I believe that it is necessary to store underscores for it to be exposed to the SDK. This should not be a breaking change for users.

EronWright:
Complementary to: https://github.com/pulumi/pulumi-kubernetes/pull/2128

EronWright (approved):
Looks great to me.

blampe (approved):

Issue #3363, opened 2024-12-11
Implement autonaming configuration protocol


mikhailshilkov:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Implement the Kubernetes part of https://github.com/pulumi/pulumi/issues/1518. See https://github.com/pulumi/pulumi/discussions/17592 for the full design.

In short, https://github.com/pulumi/pulumi/pull/17810 introduced the protobuf changes required for the provider-side implementation. With those, a provider can:

- Declare that it supports autonaming configurations with a response flag in Configure
- Accept two extra properties in CheckRequest: a proposed name and a mode to apply it

We need to implement autonaming configuration for Kubernetes.

PR #3377, opened 2024-12-18
Implement autonaming configuration protocol

mikhailshilkov:
### Proposed changes

This PR implement the Kubernetes part of https://github.com/pulumi/pulumi/issues/1518. See https://github.com/pulumi/pulumi/discussions/17592 for the full design.

In short, https://github.com/pulumi/pulumi/pull/17810 introduced the protobuf changes required for the provider-side implementation. With those, a provider can:

- Declare that it supports autonaming configurations with a response flag in Configure
- Accept two extra properties in CheckRequest: a proposed name and a mode to apply it

This PR applies those parameters to auto-name calculation if they are specified:

- In Disabled mode, we don't calculate auto-names at all
- In Enforce or Propose mode, we use whatever ProposedName is supplied by the engine

### Related issues (optional)

Resolves https://github.com/pulumi/pulumi-kubernetes/issues/3363

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Unreleased
 
-- Respect autonaming configuration specified by the engine.
+### Added
+
+- Added support for autonaming configuration specified by the engine.
   (https://github.com/pulumi/pulumi-kubernetes/issues/3363)
 
 ## 4.19.0 (December 11, 2024)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+- Respect autonaming configuration specified by the engine.
+  (https://github.com/pulumi/pulumi-kubernetes/issues/3363)
+
 ## 4.19.0 (December 11, 2024)
 
 ### Changed
--- a/provider/pkg/metadata/naming.go
+++ b/provider/pkg/metadata/naming.go
@@ -37,8 +37,9 @@ func AssignNameIfAutonamable(randomSeed []byte, engineAutonaming *pulumirpc.Chec
 	if engineAutonaming != nil {
 		switch engineAutonaming.Mode {
 		case pulumirpc.CheckRequest_AutonamingOptions_DISABLE:
-			return errors.New("autonaming is disabled, please provide a name or generateName value")
+			return errors.New("autonaming is disabled, resource requires the .metadata.name field to be set")
 		case pulumirpc.CheckRequest_AutonamingOptions_ENFORCE, pulumirpc.CheckRequest_AutonamingOptions_PROPOSE:
+			contract.Assertf(engineAutonaming.ProposedName != "", "expected proposed name to be non-empty: %v", engineAutonaming)
 			autoname = engineAutonaming.ProposedName
 		}
 	}
--- a/provider/pkg/metadata/naming.go
+++ b/provider/pkg/metadata/naming.go
@@ -15,6 +15,8 @@
 package metadata
 
 import (
+	"errors"
+
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Unreleased
 
+- Respect autonaming configuration specified by the engine.
+  (https://github.com/pulumi/pulumi-kubernetes/issues/3363)
```
blampe, line 4:
```suggestion
### Added

- Added support for autonaming configuration specified by the engine.
  (https://github.com/pulumi/pulumi-kubernetes/issues/3363)
```

provider/pkg/metadata/naming_test.go
```diff
@@ -81,11 +82,33 @@ func TestAssignNameIfAutonamable(t *testing.T) {
...
+	// o7 has no name, but autonaming is disabled by the engine, so autonaming fails.
+	o7 := &unstructured.Unstructured{}
+	autonamingDisabled := &pulumirpc.CheckRequest_AutonamingOptions{
+		Mode: pulumirpc.CheckRequest_AutonamingOptions_DISABLE,
+	}
+	AssignNameIfAutonamable(nil, autonamingDisabled, o7, pm1, resource.NewURN(tokens.QName("teststack"), tokens.PackageName("testproj"),
+		tokens.Type(""), tokens.Type("bang:boom/fizzle:MajorResource"), "foo"))
+	assert.False(t, IsAutonamed(o7))
+	assert.Equal(t, "", o7.GetGenerateName())
+	assert.Equal(t, "", o7.GetName())
```
blampe, line 111: In this case autonaming doesn't _fail_, the object just doesn't get a name. That will ultimately become an error when we try to apply the object to the cluster, since the name is required.

If we encounter this situation we should return an error to the user instead of letting them apply the invalid object. Something like "Autonaming is disabled so XYZ requires a .metadata.name". It would make sense to modify `AssignNameIfAutonamable` to return `error` for this situation.

provider/pkg/metadata/naming.go
```diff
@@ -17,17 +17,27 @@ package metadata
...
 		prefix := urn.Name() + "-"
 		autoname, err := resource.NewUniqueName(randomSeed, prefix, 0, 0, nil)
 		contract.AssertNoErrorf(err, "unexpected error while creating NewUniqueName")
+		if engineAutonaming != nil {
+			switch engineAutonaming.Mode {
+			case pulumirpc.CheckRequest_AutonamingOptions_DISABLE:
+				return
+			case pulumirpc.CheckRequest_AutonamingOptions_ENFORCE, pulumirpc.CheckRequest_AutonamingOptions_PROPOSE:
+				autoname = engineAutonaming.ProposedName
+			}
```
blampe, line 39: We should also probably validate the proposed name, since the user could specify a template that isn't valid for k8s.

I _think_ that would be [IsDNS1123Subdomain](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Subdomain) but @rquitales should double check this! 

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

blampe (commented):
Is this worth a quick yaml e2e test or are we confident in the unit behavior?

provider/pkg/metadata/naming_test.go
mikhailshilkov: Is it always an error if a `name` is not specified, for all resource types? I assumed that some objects may allow that? But it sounds like that's not true? If name/generateName are always required, an error makes sense to me.

provider/pkg/metadata/naming.go
mikhailshilkov: Do we validate `urn.Name()` anywhere?

rquitales: Yes, [IsDNS1123Subdomain](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Subdomain) is the correct validator.

blampe: > Do we validate `urn.Name()` anywhere?

That's a good question, I don't think we do! I guess we offload all of that to the cluster, so maybe we don't need to validate anything here.

I'm curious what the user experience looks like in that situation? If it's pretty clear how to resolve the issue then this is probably fine.

blampe:
Rebasing should fix your build failure.

provider/pkg/metadata/naming.go
mikhailshilkov: This is the error that one gets in both cases:

> error: resource "urn:pulumi:dev::kubernetes-ts::kubernetes:apps/v1:Deployment::nginx!!!!" was not successfully created by the Kubernetes API server: Deployment.apps "nginx!!!!-6626ce94" is invalid: metadata.name: Invalid value: "nginx!!!!-6626ce94": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

provider/pkg/metadata/naming_test.go
mikhailshilkov: Added an error on Disabled

mikhailshilkov:
> Is this worth a quick yaml e2e test or are we confident in the unit behavior?

@blampe Could you point me to an existing example that I could copy?

blampe:
> > Is this worth a quick yaml e2e test or are we confident in the unit behavior?
> 
> @blampe Could you point me to an existing example that I could copy?

This simplest/fasted would probably be to copy on of the yaml tests that live (confusingly) under tests/java 

https://github.com/pulumi/pulumi-kubernetes/blob/master/tests/sdk/java/yamlv2_test.go
https://github.com/pulumi/pulumi-kubernetes/tree/master/tests/sdk/java/testdata/yamlv2

provider/pkg/metadata/naming.go
EronWright: In Kubernetes, object name validation may differ on a per-kind basis. For example, while many resources use [IsDNS1123Subdomain](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Subdomain), some resources are more relaxed. For example, `ClusterRoleBinding` allows for colons and such names as `system:basic-user`.

I would advise against a general rule, but I see no issue with having built-in rules for specific kinds such as Service.

mikhailshilkov: Yeah... I think we shouldn't add any new validation as a part of this change (autonaming configuration). The experience is consistent across (invalid) explicit names, default auto-names, and custom auto-names.

provider/pkg/metadata/naming.go
```diff
@@ -15,22 +15,36 @@
...
+	prefix := urn.Name() + "-"
+	autoname, err := resource.NewUniqueName(randomSeed, prefix, 0, 0, nil)
+	contract.AssertNoErrorf(err, "unexpected error while creating NewUniqueName")
+	if engineAutonaming != nil {
+		switch engineAutonaming.Mode {
+		case pulumirpc.CheckRequest_AutonamingOptions_DISABLE:
+			return errors.New("autonaming is disabled, please provide a name or generateName value")
+		case pulumirpc.CheckRequest_AutonamingOptions_ENFORCE, pulumirpc.CheckRequest_AutonamingOptions_PROPOSE:
+			autoname = engineAutonaming.ProposedName
+		}
```
EronWright, line 43: Is it definitely true that ProposedName is non-empty in these modes?  The logic isn't totally clear about which modes would use a provider-generated value, would be nice to be more explicit. 
In other words, is it possible that the mode would be "proposed" without an actual proposed value?

provider/pkg/metadata/naming.go
```diff
@@ -15,22 +15,36 @@
...
+	if IsNamed(obj, propMap) || IsGenerateName(obj, propMap) {
+		return nil
+	}
+	prefix := urn.Name() + "-"
+	autoname, err := resource.NewUniqueName(randomSeed, prefix, 0, 0, nil)
+	contract.AssertNoErrorf(err, "unexpected error while creating NewUniqueName")
+	if engineAutonaming != nil {
+		switch engineAutonaming.Mode {
+		case pulumirpc.CheckRequest_AutonamingOptions_DISABLE:
+			return errors.New("autonaming is disabled, please provide a name or generateName value")
```
EronWright, line 40: Note that other error messages within the k provider use a slightly different wording, for example:
> "patch resources require the `.metadata.name` field to be set"

And, why is this different than azure-native w.r.t. to returning an error in the 'disabled' case?
https://github.com/pulumi/pulumi-azure-native/blob/b5b0a217b37ebe605bc871b552d3dc195e882d22/provider/pkg/provider/provider.go#L570-L573

EronWright (approved):

provider/pkg/metadata/naming.go
mikhailshilkov: > Note that other error messages within the k provider use a slightly different wording

Great point. Should it be `autonaming is disabled, resources require the .metadata.name field to be set`?

> why is this different than azure-native w.r.t. to returning an error in the 'disabled' case?

The end user UX is the same - users will get an error from Check. It's more of an implementation detail. In Azure Native, the `getDefaultName` doesn't return an `error`, but it's produced two levels up in Check itself, where it validates required properties. That was already the case before autonaming configuration.

rquitales (approved):

Issue #3153, opened 2024-08-07
v5: consider orphaning CRDs on delete


blampe:
We should consider orphaning CRDs by default in v5, similar to Helm's default behavior. See Eron's comment [here](https://github.com/pulumi/pulumi-kubernetes/pull/3147#pullrequestreview-2222432913).

EronWright:
Note that ArgoCD doesn't cleanup CRDs either:
https://github.com/argoproj/argo-cd/issues/613

EronWright:
Also, should CRDs always be managed with CSA rather than SSA? Seems like SSA is problematic for some CRDs.

PR #3407, opened 2025-01-03
Don't delete resources unless field managers match

blampe:
This fixes the delete-after-create problem by reviving https://github.com/pulumi/pulumi-kubernetes/pull/2999 using field managers (as Eron suggested) instead of annotations to confirm ownership before deleting a resource.

If the resource was created with SSA, and we don't find the expected field manager, then we no-op instead of deleting the resource. The intended effect is to remove it from our state so the "upserted" manager can continue owning it.

This is technically a breaking change in behavior for some edge cases, but I think it is justified by the potential severity of our delete-after-create bug.

### DIFF
--- a/provider/pkg/await/await.go
+++ b/provider/pkg/await/await.go
@@ -48,6 +48,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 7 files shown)


provider/pkg/await/await.go
```diff
@@ -834,6 +835,27 @@ func Deletion(c DeleteConfig) error {
...
+
+	actualSSAManagers := sets.Set[string]{}
+	for _, f := range live.GetManagedFields() {
+		// Ignore fields not managed by pulumi SSA.
+		if !strings.HasPrefix(f.Manager, "pulumi-kubernetes-") {
+			continue
+		}
+		actualSSAManagers = actualSSAManagers.Insert(f.Manager)
+	}
+	if c.ServerSideApply && !actualSSAManagers.Has(c.FieldManager) {
```
blampe, line 851: As written this would also no-op in cases where the object was updated with CSA. 

We can narrow this by also checking `actualSSAManagers` is non-empty, so we know the live object was at least touched by pulumi.

EronWright:
Note that Helm uses labels to track ownership and to drive adoption.
https://github.com/helm/helm/pull/7649

EronWright:
What makes me uncomfortable is the idea that Pulumi being a field manager is evidence of ownership.  Imagine that some other manager has taken over all but one field, say `foo` which Pulumi still manages.  Should we still proceed to delete the resource?

Skipping deletion when there's other field managers would not work either, because the autoscaler that is controlling the replicas has no intention of owning the object.

I would advocate for adopting a Helm-like strategy of using ownership labels.
- https://micarlise.dev/adopt-existing-helm-resources/
- https://github.com/helm/helm/pull/7649

Note that Helm recently added a flag called `--take-ownership` to skip the conflict check. I filed https://github.com/pulumi/pulumi-kubernetes/issues/3595 to track this.
- https://github.com/helm/helm/pull/13439

Interesting to note that Helm has special treatment of CRDs, e.g. they're never deleted. Consider doing the same:
- https://github.com/pulumi/pulumi-kubernetes/issues/3153

See also: https://github.com/pulumi/pulumi-kubernetes/pull/3000

EronWright (commented):
Declined to review.

PR #3550, opened 2025-03-12
Fix race condition in Deleted condition

### LINKED ISSUE #3317: Await logic for deletion is sometimes slow
It seems like the deletion await logic occasionally fails to observe the actual deletion of the resource, and waits until the timeout (and is then successful).  I don't have a repro but here's the example:

```
View in Browser (Ctrl+O): https://app.pulumi.com/eron-pulumi-corp/nested/dev/updates/7

     Type                                                           Name                              Status              
 -   pulumi:pulumi:Stack                                            nested-dev                        deleted (0.33s)     
 -   ├─ kubernetes:yaml/v2:ConfigGroup                              stack                             deleted (0.15s)     
 -   │  └─ kubernetes:pulumi.com/v1:Stack                           stack:default/stack-tidy-raven    deleted (20s)       
 -   ├─ kubernetes:rbac.authorization.k8s.io/v1:ClusterRoleBinding  crb                               deleted (600s)      
 -   ├─ kubernetes:yaml/v2:ConfigGroup                              program                           deleted (0.10s)     
 -   │  └─ kubernetes:pulumi.com/v1:Program                         program:default/stack-tidy-raven  deleted (0.24s)     
 -   ├─ kubernetes:core/v1:ServiceAccount                           ksa                               deleted (0.24s)     
 -   └─ random:index:RandomPet                                      name                              deleted (0.17s)   
Resources:
    - 8 deleted

Duration: 10m24s
```

The program resembled:
```yaml
name: example
runtime: yaml
resources:
  ksa:
    type: kubernetes:core/v1:ServiceAccount
    properties:
      metadata:
        namespace: default
  crb:
    type: kubernetes:rbac.authorization.k8s.io/v1:ClusterRoleBinding
    properties:
      roleRef:
        apiGroup: rbac.authorization.k8s.io
        kind: ClusterRole
        name: system:auth-delegator
      subjects:
        - kind: ServiceAccount
          name: ${ksa.metadata.name}
          namespace: default 
```

There's a possible clue in the diagnostic output.  Observe that the phrase "Resource scheduled for deletion" did not appear for the slow resource (the CRB).  The phrase comes from cli-utils library [here](https://github.com/kubernetes-sigs/cli-utils/blob/c241e717654cc2cc162da63d2c9ba5d5f6754df6/pkg/kstatus/status/generic.go#L33). I would guess that the object was gone before the cli-utils got a chance to observe it.

```
Diagnostics:
  kubernetes:core:ServiceAccount (ksa)
    Resource scheduled for deletion
    
 
  kubernetes:rbac.authorization.k8s.io:ClusterRoleBinding (crb)
    
 
  kubernetes:yaml/v2:ConfigGroup$kubernetes:pulumi.com/v1:Program (program:default/stack-tidy-raven)
    Resource scheduled for deletion
    
 
  kubernetes:yaml/v2:ConfigGroup$kubernetes:pulumi.com/v1:Stack (stack:default/stack-tidy-raven)
    Resource scheduled for deletion
```

EronWright:
### Proposed changes

This PR fixes a race condition in the await logic that leads to Pulumi stalling until the await timeout is reached (e.g. after 600s). The condition logic races between informer events and GET call(s) to observe the deletion of the object. In the edge case that the 'deleted' watch event is observed before the GET call completes, the system may become stuck because the getter overwrites the watcher's observation (as stored in the `cond.deleted` field).

In other words the `deleted` flag should be idempotent, and never transition from true to false.

Also, I notice that the informer could be more reliably started by the factory, since factory.Start is idempotent.  This will ensure that the informer is considered during shutdown, and that WaitForCacheSync would work as expected.

### Related issues (optional)

closes #3317

### DIFF
--- a/provider/pkg/await/condition/deleted.go
+++ b/provider/pkg/await/condition/deleted.go
@@ -58,7 +58,7 @@ func NewDeleted(
 	return dc, nil
 }
 
-// Range establishes an Informer and confirms the object still existsr.
+// Range establishes an Informer and confirms the object still exists.
 // If a Deleted event isn't Observed by the time the underlying Observer is
 // exhausted, we attempt a final lookup on the cluster to be absolutely sure it
 // still exists.
@@ -141,7 +141,6 @@ func (dc *Deleted) refreshClusterState() {
 	_, err := dc.getter.Get(ctx, dc.Object().GetName(), metav1.GetOptions{})
 	if err == nil {
 		// Still exists.
-		dc.deleted.Store(false)
 		return
 	}
 	if k8serrors.IsNotFound(err) {
--- a/provider/pkg/await/condition/source.go
+++ b/provider/pkg/await/condition/source.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/tools/cache"
 )
 
 var (
@@ -57,7 +58,6 @@ func NewDynamicSource(
 			close(stopper)
 			factory.Shutdown()
 		}()
-		factory.Start(stopper)
 		return factory
 	})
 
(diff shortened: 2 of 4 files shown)


provider/pkg/await/condition/source.go
```diff
@@ -95,16 +95,13 @@ func (des *DynamicSource) Start(_ context.Context, gvk schema.GroupVersionKind)
...
-	if check, ok := i.(interface{ HasStarted() bool }); ok {
-		if check.HasStarted() {
-			return events, nil
-		}
-	}
-	go informer.Informer().Run(des.stopper)
-	factory.WaitForCacheSync(des.stopper)
+	// Start the new informer by calling factory.Start (which is idempotent).
+	// This ensures that the informer is started exactly once and is cleaned up later.
+	factory.Start(des.stopper)
```
EronWright, line 100: See implementation of [factory.Start](https://github.com/kubernetes/client-go/blob/c106b23895edc59ff05c65770a17e6a6d3caee66/dynamic/dynamicinformer/informer.go#L91-L113)

provider/pkg/await/condition/source.go
```diff
@@ -57,7 +58,6 @@ func NewDynamicSource(
 			close(stopper)
 			factory.Shutdown()
 		}()
-		factory.Start(stopper)
```
EronWright, line 60: There are no informers to start at this point.

provider/pkg/await/condition/deleted.go
```diff
@@ -141,7 +141,6 @@ func (dc *Deleted) refreshClusterState() {
 	_, err := dc.getter.Get(ctx, dc.Object().GetName(), metav1.GetOptions{})
 	if err == nil {
 		// Still exists.
-		dc.deleted.Store(false)
```
EronWright, line 144: Rationale: the deleted field should transition from false to true, never true to false.

lblackstone (approved):

PR #3598, opened 2025-04-09
adopt-ci-mgmt

mjeffryes:
Adopt the new provider-ci native templates from ci-mgmt for this provider.


mjeffryes:
the test that's failing is "TestQuery", which looks like it maybe depends on functionality that's not longer supported in the `pulumi.runtime`?

rquitales:
@mjeffryes I think we can drop that test given that the feature we're testing is deprecated.

rquitales (approved):

PR #3612, opened 2025-04-28
Adopt ci mgmt (again)

mjeffryes:
Adds the .ci-mgmt.yaml I forgot the first time around. 

https://github.com/pulumi/pulumi-kubernetes/pull/3598 was missing the .ci-mgmt.yaml file so update jobs (eg. [Update GH workflows, ecosystem providers · pulumi/ci-mgmt@627bf35](https://github.com/pulumi/ci-mgmt/actions/runs/14700611890/job/41251076343)) are failing for this provider.

Fixes: https://github.com/pulumi/ci-mgmt/issues/1509

### DIFF
--- a/.ci-mgmt.yaml
+++ b/.ci-mgmt.yaml
@@ -1,8 +1,8 @@
 template: native
 provider: kubernetes
 lint: true
-aws: true
-gcp: true
+aws: true # need AWS credentials to test in EKS
+gcp: true # need GCP credentials to test in GKE
 pulumiVersionFile: .pulumi.version
 major-version: 4
 parallel: 3
--- a/.ci-mgmt.yaml
+++ b/.ci-mgmt.yaml
@@ -1,0 +1,16 @@
+template: native
+provider: kubernetes
+lint: true
+aws: true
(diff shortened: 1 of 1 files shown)


.ci-mgmt.yaml
```diff
@@ -0,0 +1,16 @@
+template: native
+provider: kubernetes
+lint: true
+aws: true
+gcp: true
```
EronWright, line 5: My eyes do a double-take here, can you add a comment about these?

EronWright (approved):

blampe (approved):

.ci-mgmt.yaml
mjeffryes, line 5:
```suggestion
aws: true # need AWS credentials to test in EKS
gcp: true # need GCP credentials to test in GKE
```

mjeffryes: Fair enough; these options enable credentials for AWS and GCP respectively so that the provider can run integration tests against those clouds. I've added inline comments to that effect in the file

PR #3613, opened 2025-04-28
Bump node to 18 and regenerate SDK

blampe:
It seems we introduced a number of type files (possibly in #2248) which were never subsequently used or touched. Removing the SDK directory and the re-generating types causes these files to be deleted, which seems scary but is actually fine.

As it turns out we actually generate types to a couple gigantic files ([inputs](https://github.com/pulumi/pulumi-kubernetes/blob/master/sdk/nodejs/types/input.ts), [outputs](https://github.com/pulumi/pulumi-kubernetes/blob/master/sdk/nodejs/types/output.ts)) which are namespaced in the same way we're used to seeing them on disk.


EronWright (approved):
Wow, is subtle but looks legit.

PR #3634, opened 2025-05-04
Add support for complex conditions in pulumi.com/waitFor annotation

matanbaruch:


CHANGELOG.md
```diff
@@ -1534,7 +1540,7 @@ not be affected by this change.
 
 ### Bug fixes
 
-- Fix deprecation warnings and docs. (<https://github.com/pulumi/pulumi-kubernetes/pull/918> and <https://github.com> /pulumi/pulumi-kubernetes/pull/921).
+- Fix deprecation warnings and docs. (<https://github.com/pulumi/pulumi-kubernetes/pull/918>).
```
matanbaruch, line 1543:
```suggestion
- Fix deprecation warnings and docs. (<https://github.com/pulumi/pulumi-kubernetes/pull/918> and <https://github.com> /pulumi/pulumi-kubernetes/pull/921).```

Comment (matanbaruch) [inline CHANGELOG.md:1543]:
```suggestion
- Fix deprecation warnings and docs. (<https://github.com/pulumi/pulumi-kubernetes/pull/918> and <https://github.com> /pulumi/pulumi-kubernetes/pull/921).
```

CHANGELOG.md
```diff
@@ -1706,7 +1712,7 @@ discovery.v1beta1.EndpointSlice and discovery.v1beta1.EndpointSliceList.
 
 ### Important
 
-This release reverts the default diff behavior back to the pre-`0.25.3` behavior. A new flag has
+This release disables the dry-run diff behavior introduced in `0.25.3` by default. A new flag has
```
matanbaruch, line 1715:
```suggestion
This release reverts the default diff behavior back to the pre-`0.25.3` behavior. A new flag has
```

blampe (commented):
Thank you for putting this together! This is definitely a real limitation.

One of our goals with the waitFor syntax was to stay as close as possible (ideally identical) to the syntax accepted by kubectl. So if you're already comfortable with kubectl then you don't need to learn a new DSL for Pulumi. Similarly if you need to manually debug Pulumi's behavior you can craft a CLI command directly without needing to massage things into a shape kubectl understands.

I found this issue upstream which appears to be the most relevant https://github.com/kubernetes/kubectl/issues/1224. The direction they seem to be taking is to make kubectl's JSONPath parsing more closely follow the spec, in particular the `||` operator.

All of this to say, I completely understand the need for this change, but I'm hesitant to invent new syntax because whenever kubectl eventually gets this functionality we would also need to support _that_ syntax.

Curious to hear your thoughts.

matanbaruch:
Thanks @blampe — totally understand and appreciate the goal of staying aligned with kubectl syntax. That said, we're really struggling with practical, day-to-day use cases that Pulumi is well-positioned to solve as an IaaC tool. The issue you linked (kubernetes/kubectl#1224) is already 3 years old with no clear path to resolution, and in the meantime, we're left without a clean or declarative way to express readiness conditions that are crucial in production scenarios. Pulumi is often wrapping kubectl for us, but when kubectl itself lacks these features, Pulumi can help bridge the gap until upstream support exists — if it ever does.

blampe:
>  Pulumi is often wrapping kubectl for us, but when kubectl itself lacks these features, Pulumi can help bridge the gap until upstream support exists — if it ever does.

For sure. Right now we're using `k8s.io/client-go/util/jsonpath` and I'm wondering if it might be possible to simply use a parser with better spec support. So the workarounds mentioned on that issue which _should_ work but don't would at least be supported by Pulumi. It seems reasonable to expect kubectl's jsonpath support will better match the spec _at some point_.

Issue #3521, opened 2025-02-26
Boolean operators for await annotations


stooj:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently await annotations like `pulumi.com/waitFor` with an array of conditions are true when ALL conditions are true (they are `AND`ed).

Other bools (like `OR`) would allow better control over the flow of a pulumi up and let the operation finish faster.

### Affected area/feature

- Await logic

### Example

```
"pulumi.com/waitFor": JSON.stringify([
    "jsonpath={.status.phase}=Healthy", 
--- OR ---
    "jsonpath={.status.phase}=Paused"
])
```
> We are using Argo Rollouts, In order to Identify we passed the first step or any other step, One of the following needs to be inside the .status.phase for example.

> We have services that are using steps then pause for 10m, and some are not using steps and become Healthy when they finish.
We don't want to wait the 10m as the service already stable in our opinion. So status.phase=Paused is a step which we consider Healthy as it passed the first step.

rquitales:
Thanks for this enhancement request. We'll look into how we can extend this annotation to be more expressive.

PR #3658, opened 2025-05-29
Support RFC9535 JSONPaths

blampe:
This is an alternative to https://github.com/pulumi/pulumi-kubernetes/pull/3634 which switches our JSONPath parser to the excellent [github.com/theory/jsonpath](https://github.com/theory/jsonpath) implementation of [RFC9535](https://www.rfc-editor.org/rfc/rfc9535.html#filter-selector). This allows us to support more complex conditions while still staying close to the existing `kubectl` experience.

I prefer this approach because it's _possible_ `kubectl` will update their own parser in the future to more closely match the spec. Inventing our own syntax/conventions to describe boolean logic guarantees we will drift further from `kubectl`, and it gives our users one more thing to learn. JSONPath is hard enough to grok as it is :)

Note that value comparisons are still static, so something like `{.spec.readyReplicas}={.spec.replicas}` is still not possible. If there's a strong case for it we can re-consider adding it, but I'm skeptical for now.

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/3521

cc @matanbaruch


CHANGELOG.md
```diff
@@ -7,6 +7,34 @@
...
+
+- The `pulumi.com/waitFor` annotation now uses an
+  [RFC9535](https://www.rfc-editor.org/rfc/rfc9535.html)-compliant JSONPath
+  parser. This makes it possible to wait for more complex scenarios.
+
+  For example:
+
+  - To wait for a Pod's `status.phase` to be "Running" or "Succeeded":
+
+    `pulumi.com/waitFor: "jsonpath={.status[?@ == 'Running' || @ == 'Succeeded' ]}"`
```
blampe, line 20: Unfortunately this is a little implicit because we can't specify the "phase" field directly https://mailarchive.ietf.org/arch/msg/jsonpath/ueqHXqXa_2hdjIWcjZRKtu3_bRk/

blampe:
> Is it possible to add it also to v3.x.x ?

@matanbaruch unfortunately not. We aren't supporting v3 beyond critical issues, and too much has changed to make back-porting feasible.

matanbaruch:
@blampe Can this be merged?

EronWright (approved):
Awesome!

matanbaruch:
@blampe @EronWright
Hey,
We were planning to implement this but looks like it wasn't release yet.

Is there any plan for release?

matanbaruch:
@blampe @EronWright Hey any update on release?

Issue #3702, opened 2025-07-08
helm.v4.Chart: allow autonaming resource (allows graceful statefulset replacement)


awoimbee:
Hi,
When I update statefulsets that are part of helm charts, I get errors like:

```
error: Preview failed: resource "urn:pulumi:[...]" was not successfully created by the Kubernetes API server: StatefulSet.apps "valkey-node" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```
Both at preview and at apply time.

The issue is: pulumi wants to recreate the statefulset, but it's using the same name, and the SSA API doesn't differentiate update from creation.
I could delete-before-replace, but that would cause downtime.
I would like a way to autoname this statefulset so pulumi can create-then-delete.
Note that I just want to create-then-delete the statefulset, I don't want to change the name of the whole chart.

I tried:
* Setting `"pulumi.com/autonamed"`: didn't work as it's an annotation that pulumi creates but doesn't read to know if something should be autonamed
* `props.metadata.name = undefined;`: triggers autonaming, but:
  * I have no control over the given name
  * The generated name is invalid: `valkey:base/valkey-node-094adbbd`
* A random name suffix:
  ```typescript
  transforms: [({opts, props}) => {
    if (props.kind === "StatefulSet") {
      const randomStr = Math.random().toString(36).substring(2, 6);
      props.metadata.name = `${props.metadata.name}-${randomStr}`;
      return { opts, props };
    }
    return undefined;
  }]
  ```
  The issue is that this would always cause a change and recreate the sts.
* Using `metadata.generateName`: doesn't work as SSA doesn't differentiate create and update

Proposed solution:
autonaming for chart resources strip the prefix (`valkey:base/valkey-node-094adbbd` -> `valkey-node-094adbbd`)

Issue with this solution: while it unlocks advanced users, it's terrible for newcomers (you need to use a transform and set `props.metadata.name = undefined;`...)

awoimbee:
Wait what I saw this in the code, but it doesn't work for statefulsets as they don't return `FieldImmutableErrorMsg`
https://github.com/pulumi/pulumi-kubernetes/blob/04cc31bf5794ec83cd8643e6ee321e4f0e65903a/provider/pkg/await/await.go#L200

[...]

PR #3703, opened 2025-07-08
Update autonaming to handle composite resource names (from helm charts)

awoimbee:
### Proposed changes

Update `AssignNameIfAutonamable` to strip `valkey:base/` out of the `valkey:base/valkey-node` resource name, so autonaming works on resources that are part of helm charts (`/` is invalid in a k8s resource name, `:` might be too, idk).

### Related issues (optional)

Relates to #3702

---

I hate Go's tooling, couldn't figure out how to `fmt` nor how to run the tests...

### DIFF
--- a/provider/pkg/metadata/naming.go
+++ b/provider/pkg/metadata/naming.go
@@ -16,6 +16,7 @@ package metadata
 
 import (
 	"errors"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


awoimbee:
I guess this is not happening, closing. I still think it would be a cool feature.

PR #3709, opened 2025-07-16
Moving dependencies into mise.toml.

pose:
### Proposed changes

Moving pre-req dependencies out from the `CONTRIBUTING.md` file and adding them to `mise.toml`. Developers can now `mise install` the project which should fetch and install every required build dependency. Modeling dependencies with Mise reduces the amount of global tooling required to be installed manually and enables easily switch between runtime and toolchains versions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,8 @@
 
 Dependencies required are modeled on `mise.toml`. Run `mise install` and `mise
 settings experimental=true` (required for Go binaries such as `golangci-lint`
-and `pulumictl`) to fetch and install them.
+and `pulumictl`) to fetch and install them. Finally, run `mise env` to check if
+env variables are being set correctly.
 
 ### Restore Vendor Dependencies
 


mise.toml
```diff
@@ -0,0 +1,29 @@
+[tools]
+# Project overrides (formerly .pulumi.version)
```
danielrbradley, line 2: Saying "formerly" .. are we removing the .pulumi.version .. or are we still using that for the makefile dependency?

mise.toml
```diff
@@ -0,0 +1,29 @@
...
+nodejs = '20'
+python = '3.11'
+
+# See toolVersions at
+# https://github.com/pulumi/pulumi-kubernetes/blob/master/.ci-mgmt.yaml
+go = '1.24'
+
+# CONTRIBUTING.md depencies
+kubectl = '1.33.2'
+# TODO Find why we cannot just use npm and remove one extra dependency
```
danielrbradley, line 19: Probably worth taking this note out of here. Perhaps open an issue in ci-mgmt for moving from yarn to npm in all repos?

pose: https://github.com/pulumi/ci-mgmt/issues/1624

pose: (and removed from the code)

danielrbradley (approved):

PR #3711, opened 2025-07-16
Fix Go tests (?)

blampe:
We seem to have an issue where `pulumi up` will exit with status code 255 after the preview step (indicating a cancelation). I reproduced this locally with 3.173 and CI also experiences this with 3.171. (Note the provider is built against 3.180.) This only seems to impact Go tests, possibly due to those runners having cached older versions.

The issue seems to resolve when using a newer release of the CLI, so this should at least unblock us.


pose:
Thanks for troubleshooting this 🙏.  Do we need to flag this as a _known issue_ for customers?

pose:
Merging as this is blocking my other PR, but happy to discuss offline!

PR #3716, opened 2025-07-17
Update Java to 1.16.0

t0yv0:
Since OSSRH is sunset, Gradle projects require an update to continue publishing artifacts.

Part of a larger cross-Pulumi rollout:

https://github.com/pulumi/ci-mgmt/issues/1611


blampe:
@t0yv0 this one uses gen-sdk so I think we'll just bump pu/pu. Edit: oh dang are we not generating this anymore?

blampe:
Moving to https://github.com/pulumi/pulumi-kubernetes/pull/3714

PR #3793, opened 2025-08-27
Return live state from the awaiter

blampe:
tl;dr: We've had a synchronization issue for a while that primarily affects tests. It was made worse by #3733, probably due to less overhead from our informers. This PR resolves the problem by simplifying things such that we always return the state of an object as it was observed by the awaiter when we decided it was ready.

For background, we have some generic await logic and we have some legacy awaiters for things like pods, deployments, etc. The legacy awaiters are wrapped in a way that allows them to behave like generic awaiters, but they essentially just consist of some methods which spin up their own informers and block until a condition is met.

Additionally, after waiting for an object to become ready we re-query the cluster for that object's state. Importantly this is a quorum read.

This creates couple minor issues:
1. The generic await layer isn't guaranteed to see all of same events delivered to legacy awaiters by the time the legacy awaiter decides an object is ready. This isn't a user-facing issue because of the quorum read, but it is a source of flakes for tests against the await layer.
2. It's technically possible for an object to be modified after we decide it's ready but before we re-query its state.

We could resolve these issues by waiting for caches to sync, but I opted to address a TODO and simplify things a bit.

Instead, we can return the object's state directly as it was when the awaiter decided it was ready. This state includes all of the same fields (status, managed fields, etc.) we get from the quorum read, but without any room for surprises. IMO this is worthwhile because it's easier to reason about, is less load on the API server, and is "more correct."

Fixes https://github.com/pulumi/pulumi-kubernetes/issues/3794.


provider/pkg/await/deployment_test.go
```diff
@@ -610,7 +610,7 @@ func Test_Apps_Deployment(t *testing.T) {
 		period := make(chan time.Time)
 		go test.do(deployments, replicaSets, pods, timeout)
 
-		err := awaiter.await(
+		_, err := awaiter.await(
```
EronWright, line 613: Consider asserting that the correct live object is returned.

tests/sdk/dotnet/helm-local/step1/nginx/values.yaml
```diff
@@ -12,8 +12,8 @@
 ##
 image:
   registry: docker.io
-  repository: bitnami/nginx
-  tag: 1.19.1-debian-10-r23
+  repository: bitnamisecure/nginx
+  tag: latest
```
EronWright, line 16: nit: use a sha

EronWright (approved):
Consider using legacy source:
https://github.com/bitnami/charts/issues/36215#issuecomment-3233843307

PR #3845, opened 2025-09-23
Add Kustomize URLs support to V2

### LINKED ISSUE #3844: Kustomize V2 support does not support Git URLs
I ran into a bug and it seems that we're missing equivalent logic to our V1 support to handle Git URLs:

https://github.com/pulumi/pulumi-kubernetes/blob/bc3c4569d3ef3997142d0fdd3141cd087df7acd0/provider/pkg/provider/invoke_kustomize.go#L37-L52

joeduffy:
We seem to have supported this in V1, but do not in V2. Krusty [expects](https://pkg.go.dev/sigs.k8s.io/kustomize/api/krusty@v0.20.1#MakeKustomizer) all files to already exist locally:

> Any files referenced by the kustomization must be present on the filesystem.

This change uses the upstream [localize](https://github.com/kubernetes-sigs/kustomize/blob/master/proposals/22-04-localize-command.md) command to fetch the remote directory, and we then act on our local copy as normal.

Fixes #3844.

### DIFF
--- a/provider/pkg/provider/kustomize/v2/directory.go
+++ b/provider/pkg/provider/kustomize/v2/directory.go
@@ -16,17 +16,19 @@ package v2
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/url"
 	"os"
+	"path/filepath"
 
 	providerresource "github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/provider/resource"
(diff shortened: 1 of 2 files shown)


lblackstone:
Tests failing with the following error:
```
error: kubernetes:kustomize/v2:Directory resource 'remote' has a problem: kustomize build error: unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/tmp/pulumi-kustomize2290689844/local'
```

PR #3931, opened 2025-11-05
Enable more node tests to run on run-acceptance-tests

guineveresaenger:
These tests were not running on pull requests because `run-acceptance-tests` runs all tests in short mode by default. The assertion that these tests provision node balancers and thus need more than a KinD cluster seem incorrect at first glance; we should be able to run them on a KinD cluster.
This pull request removes the `tests.SkipIfShort` condition on these tests.

Tests now enabled are:

- TestAccGuestbook
- TestAccHelm
- TestAccHelmLocal
- TestAccPrometheusOperator
- TestHelmRelease
- TestHelmReleaseNamespace

### DIFF
--- a/tests/sdk/nodejs/examples/examples_test.go
+++ b/tests/sdk/nodejs/examples/examples_test.go
@@ -43,7 +43,6 @@ func TestAccMinimal(t *testing.T) {
 }
 
 func TestAccGuestbook(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pose (approved):
(not a k8s expert, but since it's just unskipping tests seems reasonable!)

Issue #3713, opened 2025-07-17
Significant increase in memory usage from v4.18.0 onwards


stooj:
### Describe what happened

`pulumi-kubernetes` consumes far more memory after the v4.18.0 release.

### Sample program

I'm running a kind cluster for testing:

```typescript
import * as kubernetes from "@pulumi/kubernetes";

const provider = new kubernetes.Provider("kubernetes-provider", {
  context: "kind-kind",
  enableServerSideApply: true,
});

for (let i = 0; i < 100; i++) {
  new kubernetes.batch.v1.Job(
    `test-job-${i}`,
    {
      spec: {
        template: {
          spec: {
            containers: [
              {
                name: "hello-world",
                image: "busybox",
                command: ["/bin/sh", "-c", `sleep 10`],
              },
            ],
            restartPolicy: "OnFailure",
          },
        },
        backoffLimit: 4,
      },
    },
    { provider },
  );
}
```

### Log output

```mermaid
xychart-beta
    title "Memory usage for `pulumi-kubernetes` versions"
    x-axis "pulumi/kubernetes version"[4.13.1,4.14.0,4.15.0,4.16.0,4.17.1,4.18.0,4.18.1,4.18.2,4.18.3,4.18.4,4.19.0,4.20.0,4.21.0,4.22.0,4.23.0]
    y-axis "Max memory usage (in Mi)" 0 --> 6000
    bar [779,790,761,784,796,5909,5889,6109,6349,4478,4383,4326,4704,4763,5133]
```

### Output of `pulumi about`

```
CLI
Version      3.177.0
Go Version   go1.24.4
Go Compiler  gc

Plugins
KIND      NAME        VERSION
resource  kubernetes  4.21.0
language  nodejs      3.177.0

Host
OS       nixos
Version  25.05 (Warbler)
Arch     x86_64

This project is written in nodejs: executable='/nix/store/c8jxsih8yy2rnncdmx2hyraizf689nvp-nodejs-22.14.0/bin/node' version='v22.14.0'

Current Stack: stooj/7442/dev

Found no resources associated with dev

Found no pending operations associated with dev

Backend
Name           pulumi.com
URL            https://app.pulumi.com/stooj
User           stooj
Organizations  stooj, team-ce, higara, pequod, demo, higara-free
Token type     personal

Dependencies:
NAME                VERSION
@pulumi/pulumi      3.181.0
@types/node         18.19.118
typescript          5.8.3
[...]

PR #3980, opened 2025-11-21
Add more safeguards around stale generations

blampe:
#3713 introduced a race condition where a watch could already be established and serve some stale events for an object (i.e. events with the wrong generation).

This PR adds safeguards around our Deployment, StatefulSet, DaemonSet, and generic awaiters to ignore any events without a current generation.

I intentionally check `observed < expected` instead of `observed == expected` because I've been in situations where I realize a Pulumi deployment is stuck and I manually tweak something (incrementing a generation) to un-stick it. Using `==` would cause the update to hang instead, which is undesirable.

Fixes https://github.com/pulumi/customer-support/issues/2509.


provider/pkg/await/condition/observer.go
```diff
@@ -84,6 +84,12 @@ func (oo *ObjectObserver) Observe(e watch.Event) error {
 	oo.mu.Lock()
 	defer oo.mu.Unlock()
 	obj, _ := e.Object.(*unstructured.Unstructured)
+
+	// Do nothing if this is a stale object.
+	if obj.GetGeneration() < oo.obj.GetGeneration() {
+		return nil
+	}
```
EronWright: Nit: Would you agree that a more precise comparison would be of the resourceVersion than the generation? You're trying to observe versions that are later than what was returned from the mutating create/update. Your logic would needlessly re-observe the initial version, right?

In other words, **generation** is a concept of versioned specs, but the awaiter logic is primarily about **status changes** (where generation doesn't change).

EronWright (changes requested):

EronWright:
Alright, given that the resource Version is supposed to be opaque, I suppose generation is the best one can do. please revert the other stuff.

Issue #3965, opened 2025-11-18
`PulumiSchema` should rename properties & types named `pulumi`


ringods:
### Describe what happened

A customer reported that `crd2pulumi` v1.6.0 failed to generate SDKs for the External Secrets Operator CRD. See issue https://github.com/pulumi/crd2pulumi/issues/248. The External Secrets Operator CRD file contains a property named `pulumi` because we contributed Pulumi ESC support to this operator:

https://github.com/external-secrets/external-secrets/blob/main/deploy/crds/bundle.yaml#L5833-L5890

Investigation leads to this commit not allowing the name `pulumi` as a property or type name in Pulumi schema:

https://github.com/pulumi/pulumi/commit/4c73d80b9baae2edc0ed1814e7c8f656764c1576#diff-9ab9df002b36270fe280ee8e8d162d61a0946708e23a60733f30159b8727b937

This commit was introduced in pulumi/pulumi v3.172.0.

We can prevent this by renaming any such property or type in `crd2pulumi`. However, the better place is for the schema creation function `PulumiSchema(...)` in `provider/pkg/gen/schema.go` to return schema with this property already renamed.

### Sample program

See the example in https://github.com/pulumi/crd2pulumi/issues/248

### Output of `pulumi about`

See stack trace in other issue.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

dancmeyers:
Ahha, I clearly searched the wrong things when raising my bug as I didn't find a prior report, but I was raising it as a `crd2pulumi` error rather than a `@pulumi/kubernetes` error. I think https://github.com/pulumi/crd2pulumi/issues/251 is a duplicate of this.

dancmeyers:
I've resolved my issue in favour of this one.

pgavlin:
Fixed by https://github.com/pulumi/pulumi-kubernetes/pull/4036.

PR #4036, opened 2025-12-19
gen: rename properties named 'pulumi'

pgavlin:
Attempt to rename these properties to 'pulumi_', and keep adding
underscores if the name already exists. This absorbs a breaking change
from the fix to https://github.com/pulumi/pulumi/issues/18994, which
made 'pulumi' a reserved property name.

Motivated by https://github.com/pulumi/crd2pulumi/issues/253, as
`crd2pulumi` uses this infra for its own schema generation.

### DIFF
--- a/provider/pkg/gen/typegen.go
+++ b/provider/pkg/gen/typegen.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"slices"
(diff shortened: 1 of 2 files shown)


pgavlin:
Fixes https://github.com/pulumi/pulumi-kubernetes/issues/3965.

blampe (approved):

PR #4063, opened 2026-01-07
Free more disk space and also clear go cache

guineveresaenger:
Experiment to see if we can squeeze more compute out of the publish runner.

Any merge would be overwritten by ci-mgmt, so we probably don't want to merge.


guineveresaenger:
Closing in favor of https://github.com/pulumi/pulumi-kubernetes/pull/4069.

PR #4094, opened 2026-01-23
Check observedGeneration when awaiting StatefulSet to avoid race

yhrn:
### Proposed changes

When awaiting a StatefulSet change completion also verify that `status.observerdGeneration == metadata.generation` or otherwise treat the the object as not ready. 

Without this check there is a chance that Pulumi will look at the status before the controller has had the chance to update the status so it still looks ready. `metadata.generation` is incremented by the API server on update so this prevents the described race condition. Currently all comparisons are just between fields in the status, i.e. fields that are set by the controller.

We see this problem around 50% ov the time so this is not just a theoretical edge case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 
 ### Fixed
 
+- Fixed a race condition in StatefulSet await logic where Pulumi could
+  incorrectly report a rollout as complete before the controller had processed
+  the update. This occurred when Pulumi checked status fields before the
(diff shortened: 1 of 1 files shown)


guineveresaenger:
Hi @yhrn - thank you for this pull request. Running tests now.

/run-acceptance-tests

guineveresaenger:
Tests pass. Thank you for this change, and thank you for adding tests! We will pull these changes into the next release.

PR #4122, opened 2026-01-31
Add new `alwaysRender` option to provider config for use with yaml render mode

### LINKED ISSUE #4121: renderYamlToDirectory does not render unchanged resources
## Description

When using `renderYamlToDirectory` to generate Kubernetes manifests as YAML files, only resources that are being **created** or **updated** have their YAML files generated. Resources that are **unchanged** (no diff detected) do not get rendered because the `Update` method is never called for them.

This is problematic when using `renderYamlToDirectory` with tools like **ArgoCD Config Management Plugin (CMP)**, which expects to receive all manifests on every run, regardless of whether resources have changed.

## Current Behavior

1. First `pulumi up` with `renderYamlToDirectory`: All resources are rendered (they're all new)
2. Subsequent `pulumi up` with no code changes: **No YAML files are generated** because:
   - `Diff()` returns `DIFF_NONE` for unchanged resources
   - Pulumi engine skips `Update()` when `DIFF_NONE`
   - `renderYaml()` is only called inside `Update()`

## Expected Behavior

When `renderYamlToDirectory` is enabled, all Kubernetes resources should have their YAML files generated on every run, regardless of whether they have changed.

## Use Case

Using Pulumi with ArgoCD CMP to:
1. Define Kubernetes resources in Pulumi (TypeScript)
2. Use `renderYamlToDirectory` to generate YAML manifests
3. Have ArgoCD manage the rendered manifests

This pattern allows:
- Pulumi to manage both cloud infrastructure (GCP, AWS) and Kubernetes resources
- ArgoCD to track and visualize the Kubernetes resources
- Cross-stack references to work (K8s resources can reference cloud outputs)

## Proposed Solution

In YAML render mode, the `Diff()` function should always return `DIFF_SOME` to force the `Update()` method to be called. This is safe because:
- `Update()` in `yamlRenderMode` only writes files to disk
- It does not modify any actual cluster resources
- The `deleteBeforeReplace` and `replaces` logic is irrelevant in render mode

## Environment

- Pulumi Kubernetes Provider version: 4.25.0
- Pulumi CLI version: 3.x
- OS: macOS/Linux

mateo-moon:
edited Feb 6 2025 by @guineveresaenger:

Implements a new provider config option, `alwaysRender`, which sets `Diff_Some` to true when used in combination with `renderYAMLtoDirectory`.

This allows for user-defined opt-in behavior for YAML manifests to be idempotently generated on every `pulumi up`, and for Update() to show these changes to a user.

Fixes #4121

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -96,6 +96,7 @@ const (
 	fieldManagerKey      = "__fieldManager"
 	secretKind           = "Secret"
 	clusterIdentifierKey = "clusterIdentifier"
+	trueStr              = "true"
 )
 
 type cancellationContext struct {
@@ -309,7 +310,7 @@ func (k *kubeProvider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequ
 	if pConfig, ok := k.loadPulumiConfig(); ok {
 		if v, ok := pConfig["strictMode"]; ok {
 			if v, ok := v.(string); ok {
-				strictMode = v == "true"
+				strictMode = v == trueStr
 			}
 		}
 	}


guineveresaenger:
Hi @mateo-moon, thank you for opening this pull request.

While I'm running the tests, I recommend you run the [TestRenderYAML](https://github.com/pulumi/pulumi-kubernetes/blob/a1cf1b37af69abe3191f2b8a6613dca1dc185d84/tests/sdk/nodejs/nodejs_test.go#L1115) locally with this change (possibly with `SkipRefresh: true` since I ran into some issues locally) and see what turns up.

That would be the place to test whether this functionality is operating as expected.

I was also wondering if it might make sense to perform the render in Diff instead of Update? Maybe conditionally when `pulumirpc.DiffResponse_DIFF_NONE` is true?

mateo-moon:
Thanks for the review and the suggestion! I've updated the implementation as you suggested.

Instead of forcing `DIFF_SOME` to trigger `Update()` for all resources, the fix now renders the YAML directly in `Diff()` when there are no actual changes:

```go
if k.yamlRenderMode && hasChanges == pulumirpc.DiffResponse_DIFF_NONE {
    if err := renderYaml(newInputs, k.yamlDirectory); err != nil {
        return nil, err
    }
}
```

This approach:
- Preserves correct diff semantics (`refresh --expect-no-changes` will work correctly)
- Resources with actual changes are still rendered via `Update()` as before
- Resources without changes get rendered via `Diff()`
- All resources end up in the YAML output directory

This should fix the `TestRenderYAML` test failure since `pulumi refresh --expect-no-changes` will no longer see phantom changes.

guineveresaenger:
hi! I apologize for the back and forth here. It turns out that this is a bit less straightforward than I initially thought, since this is essentially a new feature.

On second thought it feels incorrect to quietly rewrite files during Diff() by default. Anyone who uses `renderYamlToDirectory` should see updates explicitly, even if they're idempotent file writes. So your initial approach is more in line with what one would expect.

That said, enabling this in Update() across the board would cause a lot of noise for users who would like to continue using `renderYamlToDirectory` as they have been - showing only actual updates. In the past we did actually fix noise around this.

Therefore, neither option is ideal.

I'm going to discuss this internally with the team, but one thought I had would be to introduce a new provider input property `alwaysRender` that would control this behavior conditionally, so you could configure your preference. But this now being an entirely new feature, we have to think about maintainer burden before we take on this change.

Thank you very much for your patience!

guineveresaenger:
update! 

We have decided to implement your desired functionality, but as an additional flag `alwaysRender`. I hope you don't mind if I add that change to this pull request. We really appreciate your contribution, and that you highlighted this use case.

We'll continue to render manifests in Update() - this is more honest to the paradigm than doing so in Diff() although if the output noise is too great, we may reconsider this decision. But with an explicit option, you can opt-in to the desired behavior; current behavior remains backwards compatible, and there will be no invisible surprises, because opt-in is explicit.

mateo-moon:
> update!
> 
> We have decided to implement your desired functionality, but as an additional flag `alwaysRender`. I hope you don't mind if I add that change to this pull request. We really appreciate your contribution, and that you highlighted this use case.
> 
> We'll continue to render manifests in Update() - this is more honest to the paradigm than doing so in Diff() although if the output noise is too great, we may reconsider this decision. But with an explicit option, you can opt-in to the desired behavior; current behavior remains backwards compatible, and there will be no invisible surprises, because opt-in is explicit.

Not at all. Thank you for a quick response and fix.

provider/pkg/provider/provider.go
```diff
@@ -657,6 +666,8 @@ func (k *kubeProvider) Configure(_ context.Context, req *pulumirpc.ConfigureRequ
 	k.yamlDirectory = renderYamlToDirectory()
 	k.yamlRenderMode = len(k.yamlDirectory) > 0
 
+	k.alwaysRender = vars["kubernetes:config:alwaysRender"] == trueStr
```
pgavlin: is there a difference between comparing against `trueStr` and the `truthyValue` helper used in `CheckConfig`?

PR #4167, opened 2026-02-20
Automate alerts for upstream Kubernetes releases

### LINKED ISSUE #4165: Automate kubernetes/kubernetes upgrades
We're being entirely manual in our upgrades, but large parts can be automated via a regular update check.

We've also missed upstream upgrades because our alert system was not persistent.

We have a [playbook](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/Release%3A%20Native%20Providers.md#kubernetes-supporting-newer-kubernetes-releases) for releases; we should attempt to simplify and automate it as much as possible.

The end result should be:

1. A regular (weekly, daily?) Workflow check for a new version of kubernetes/kubernetes
2. Upon detecting an update, open an issue "Update k8s to v$NEXT"
3. Attempt to open a PR that has followed the automatable steps for an upgrade as much as possible
4. Include remaining manual steps in the issue body (reduce this effort over time with more automation.)

guineveresaenger:
This pull request adds a Workflow to open an issue and attempt a pull request to automatically upgrade the Kubernetes APIs when a new upstream release is made. This workflow is closely modeled after `weekly-pulumi-update.yml`.

I've allowed the Workflow to run not just on a cron, but also on push and dispatch so hopefully we can iron out any wrinkles before merging to main. I'll tweak that once we find things look OK.

Fixes #4165.

- **First pass at weekly kube upgrade check**
- **tweaks**
- **Link issue number to PR fix**

### DIFF
--- a/.github/workflows/kubernetes-update-check.yml
+++ b/.github/workflows/kubernetes-update-check.yml
@@ -21,25 +21,6 @@ jobs:
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           lfs: true
-      - env:
-          ESC_ACTION_ENVIRONMENT: github-secrets/${{ github.repository_owner }}-${{ github.event.repository.name }}
-          ESC_ACTION_EXPORT_ENVIRONMENT_VARIABLES: "false"
-          ESC_ACTION_OIDC_AUTH: "true"
-          ESC_ACTION_OIDC_ORGANIZATION: pulumi
-          ESC_ACTION_OIDC_REQUESTED_TOKEN_TYPE: urn:pulumi:token-type:access_token:organization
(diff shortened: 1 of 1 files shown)


.github/workflows/kubernetes-update-check.yml
```diff
@@ -0,0 +1,125 @@
+name: kubernetes-update
+on:
+  push:
+    branches:
+      - 'guin/automate-upgrades'
+  schedule:
+    - cron: 35 12 * * 4
```
blampe: It's helpful to leave a comment on crons to make them easy to grok at a glance, and to make it easier for reviewers to confirm it matches your intent.
```suggestion
    - cron: 35 12 * * 4 # 12:35 UTC on Thursdays
```

PR #4215, opened 2026-03-09
fix: Surface exec credential plugin stderr on cluster unreachable errors

### LINKED ISSUE #4212: Exec credential plugin stderr is hidden from users when the cluster is unreachable
## Problem

When a kubeconfig uses an exec-based credential plugin (e.g., `aws eks get-token`, `gke-gcloud-auth-plugin`, `kubelogin`) and that plugin fails, users only see a generic error like:

```
error: configured Kubernetes cluster is unreachable: unable to load schema
information from the API server: command terminated with exit code 1
```

The actual subprocess stderr — which contains the actionable error (expired credentials, missing IAM role, wrong region, etc.) — is piped by client-go directly to the provider's `os.Stderr`, which the Pulumi engine does not forward to the terminal. It is silently discarded.

The only workaround today is `pulumi up -v=9`.

## Root Cause

client-go runs exec credential plugins lazily, on the first HTTP request. In this provider that happens inside `getResources()` during `Configure()`. client-go sets the subprocess's stderr to `os.Stderr` of the provider process with no public API to intercept it. The error returned to the caller only contains client-go's own message, not the plugin's output.

## Proposed Fix

On the failure path only, re-run the exec plugin with stderr captured in a buffer and append it to `clusterUnreachableReason`. Since this only runs after a failure is already detected, there is no performance impact on the happy path.

**Example before:**
```
error: unable to load schema information from the API server: command terminated with exit code 1
```

**Example after:**
```
error: unable to load schema information from the API server: command terminated with exit code 1

exec credential plugin output:
Error: token refresh failed: credentials have expired, please run `aws sso login`
```

### 1. Add `execCredentialStderr` helper to `provider/pkg/provider/util.go`

```go
func execCredentialStderr(kubeconfigClient clientcmd.ClientConfig, overrides *clientcmd.ConfigOverrides) string {
    if kubeconfigClient == nil {
        return ""
    }
    rawConfig, err := kubeconfigClient.RawConfig()
    if err != nil {
        return ""
    }

    contextName := rawConfig.CurrentContext
    if overrides != nil && overrides.CurrentContext != "" {
        contextName = overrides.CurrentContext
    }

    ctx, ok := rawConfig.Contexts[contextName]
    if !ok || ctx == nil {
        return ""
    }
    authInfo, ok := rawConfig.AuthInfos[ctx.AuthInfo]
    if !ok || authInfo == nil || authInfo.Exec == nil || authInfo.Exec.Command == "" {
        return ""
    }

    var stderrBuf bytes.Buffer
    cmd := exec.Command(authInfo.Exec.Command, authInfo.Exec.Args...)
    cmd.Stderr = &stderrBuf
    cmd.Env = os.Environ()
    for _, e := range authInfo.Exec.Env {
        cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", e.Name, e.Value))
    }
    _ = cmd.Run()
    return strings.TrimSpace(stderrBuf.String())
}
```

### 2. Enrich the error in `provider/pkg/provider/provider.go` (lines 914-917)

```go
if _, err = k.getResources(); err != nil {
    k.clusterUnreachable = true
    k.clusterUnreachableReason = fmt.Sprintf(
        "unable to load schema information from the API server: %v", err)
    if execStderr := execCredentialStderr(kubeconfig, overrides); execStderr != "" {
        k.clusterUnreachableReason += fmt.Sprintf(
            "\n\nexec credential plugin output:\n%s", execStderr)
    }
}
```

### 3. Add unit tests

In `provider_configure_test.go` (Ginkgo), cover:
- Auth info with no exec config → returns `""`
- Exec plugin writes to stderr and exits non-zero → returns that stderr
- Context override selects the correct user
- Nil overrides → no panic

## Notes

- The `kubeconfig.ClientConfig()` failure path (line 870) is out of scope — that path fires for kubeconfig structure errors (bad TLS certs, malformed YAML), not exec plugin failures, since client-go does not invoke the exec plugin during `ClientConfig()`.
- **Security:** exec plugin args come from the user's trusted kubeconfig. We only capture stderr (not stdout, which contains tokens). Well-behaved plugins only write diagnostics to stderr.

rshade:
## Summary

Fixes #4212.

When a kubeconfig uses an exec-based credential plugin (`aws eks get-token`, `gke-gcloud-auth-plugin`, `kubelogin`, etc.) and authentication fails, users previously saw only a generic error:

```
error: configured Kubernetes cluster is unreachable: unable to load schema
information from the API server: command terminated with exit code 1
```

The actionable diagnostic — expired credentials, missing IAM role, wrong region — was written by the subprocess to `os.Stderr` of the provider process, which the Pulumi engine does not forward to the terminal, silently discarding it.

After this change, the same failure surfaces as:

```
error: configured Kubernetes cluster is unreachable: unable to load schema
information from the API server: command terminated with exit code 1

exec credential plugin output:
Error: token refresh failed: credentials have expired, please run `aws sso login`
```

## Changes

- **`provider/pkg/provider/util.go`**: Adds `execCredentialStderr`, which re-runs the exec credential plugin with stderr captured in a buffer. Called only on the failure path — no performance impact on the happy path.
- **`provider/pkg/provider/provider.go`**: On `getResources()` failure, calls `execCredentialStderr` and appends any output to `clusterUnreachableReason`.
- **`provider/pkg/provider/util_test.go`**: Unit tests covering nil input, no-exec config, stderr capture, context override selection, and nil-panic safety.

## Test plan

- [x] `go test ./pkg/provider/... -run TestExecCredentialStderr` — all 5 cases pass
- [ ] Manual test: configure a kubeconfig with an exec plugin that fails (e.g., expired AWS SSO session) and run `pulumi preview` — verify the plugin's stderr appears in the error message without needing `-v=9`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 
 ## 4.28.0 (March 12, 2026)
 
+### Fixed
+
+- [#4215](https://github.com/pulumi/pulumi-kubernetes/pull/4215) When a kubeconfig uses an exec-based credential plugin (`aws eks get-token`, `gke-gcloud-auth-plugin`, `kubelogin`, etc.) and the cluster is unreachable due to an authentication failure, the plugin's stderr output is now included in the error message. Previously, actionable diagnostics such as "credentials have expired" were silently discarded.
+
 ### Changed
 
 - Upgrade Kubernetes schema and libraries to v1.35.2.
--- a/provider/pkg/await/checker/checker.go
+++ b/provider/pkg/await/checker/checker.go
@@ -92,6 +92,9 @@ func (s *StateChecker) Ready(state interface{}) bool {
 
 func (s *StateChecker) ReadyStatus(state interface{}) (bool, Result) {
 	ok, results := s.readyDetails(state)
+	if len(results) == 0 {
+		return ok, Result{Ok: ok}
+	}
 	return ok, results[len(results)-1]
 }
 
--- a/provider/pkg/await/checker/job/check_test.go
+++ b/provider/pkg/await/checker/job/check_test.go
@@ -21,7 +21,6 @@ import (
 	"github.com/stretchr/testify/require"
 	batchv1 "k8s.io/api/batch/v1"
 
-	"github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/await/checker"
 	"github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/await/checker/internal"
 )
 
@@ -148,16 +147,12 @@ func Test_Job_Checker(t *testing.T) {
 			jobChecker := NewJobChecker()
 
 			ready := false
-			var details checker.Results
 			jobStates := loadWorkflows(t, tt.workflowPaths...)
 			for _, jobState := range jobStates {
-				ready, details = jobChecker.ReadyDetails(jobState)
-				if ready {
+				if ready, _ = jobChecker.ReadyDetails(jobState); ready {
 					break
 				}
 			}
-			fmt.Printf("Expect Ready() = %t\n", tt.expectReady)
-			fmt.Println(details)
 			if ready != tt.expectReady {
 				t.Errorf("Ready() = %t, want %t", ready, tt.expectReady)
 			}
--- a/provider/pkg/await/checker/pod/check.go
+++ b/provider/pkg/await/checker/pod/check.go
@@ -72,7 +72,7 @@ func podInitialized(obj interface{}) checker.Result {
 		return result
 	}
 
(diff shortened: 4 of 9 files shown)


guineveresaenger:
Hi @rshade, thank you for working on this! Do you need assistance? 💟 

I think we can justify a little CHANGELOG entry here, since it's user-facing.

rshade:
Hey @guineveresaenger I think I am going to ask claude if it can figure out everywhere this might leak a secret first before I convert it to a full PR. I didn't really like it having to do a separate exec just to bubble up the error, but it looks like the only way.  Thank you for the offer of help. I will get it out, and definitely on the changelog.

PR #4270, opened 2026-04-08
Block server-side apply upsert of existing objects on create

### LINKED ISSUE #2926: Fail-fast when a program attempts to create a namespace that already exists.

guineveresaenger:
This pull request addresses the issues laid out in https://github.com/pulumi/pulumi-kubernetes/issues/2926. 

Specifically, here we are approaching #2926 as a **bug**. When a user attempts to delete an object that is in Pulumi state, such a delete should succeed. Because Create defaults to using server-side apply, we can wind up with undeleteable objects in Pulumi state. **I would argue that this is a bug.**

Therefore, this PR does the following:

- Add a GET call before each Create to assert that an object already exists. This is perhaps the most controversial choice since it adds a full ; however it allows us to keep using Server-Side Apply on Create, rather than reverting to client-side apply, which I understand has some migration issues wrt field managers.

- Fail Create when an object already exists. The user is suggested three options:
    - use an alias if the operation is a rename so we don't accidentally delete objects
    - use `DeleteBeforeReplace` resource option for renames
    - use a Patch resource to only manage certain object fields.
  We don't tell the user to explicitly put the object into state via Import, because that will result in Deletes hitting the Forbidden erorr.

- ~Relinquish on Forbidden delete: When the API server forbids deletion (e.g. protected namespaces like `kube-system`), fall back to relinquishing managed fields instead of failing. This is for situations where a user explicitly imported a protected object.~ This was too broad. Users should be directed towards patch resources, which can be safely deleted.

- **Also adds `upsertExistingObjects` provider config**: [Users have reported they like this behavior for bulk import purposes](https://docs.google.com/document/d/1Tvz1rUvEMnBwOAPlLsnnmGmx_dzUUmA5iK0PrtqKdD8/edit?tab=t.0#heading=h.9wa2rg4iusal). Therefore, we also provide an opt-in flag for those users.

Fixes #2926
Refs #2948

## Test plan

- [x] Unit tests for create-side existence check (existing object blocks create, preview also blocks, new objects proceed, Patch resources skip check, `upsertExistingObjects` skips check)
- [x] E2E test: creating the built-in `default` namespace fails with "already exists" error
- [x] E2E test: upsert + destroy of protected namespace succeeds via Relinquish fallback
- [x] Existing Creation/Update/Deletion unit tests pass


tests/sdk/yaml/testdata/existing-object/Pulumi.yaml
```diff
@@ -0,0 +1,11 @@
+name: existing-object
```
pose, line 1: YAML tests make me happy :)

provider/pkg/provider/provider.go
```diff
@@ -629,6 +630,19 @@ func (k *kubeProvider) Configure(
 		k.enablePatchForce = true
 	}
 
+	upsertExistingObjects := func() bool {
+		if enabled, exists := vars["kubernetes:config:upsertExistingObjects"]; exists {
+			return enabled == trueStr
+		}
+		if enabled, exists := os.LookupEnv("PULUMI_K8S_UPSERT_EXISTING_OBJECTS"); exists {
+			return enabled == trueStr
```
pose, line 638: (nitpick) If I understand this correctly, would using in bash`export PULUMI_K8S_UPSERT_EXISTING_OBJECTS=1` would fail this check since we are comparing exclusively for `== true`. Is that desirable? I would expect `=1` to be a valid way to set it to true. What do you think?

pose (approved):

provider/pkg/await/await.go
```diff
@@ -186,6 +187,31 @@ func Creation(c CreateConfig) (*unstructured.Unstructured, error) {
...
+								"(2) use deleteBeforeReplace if replacing a resource with an explicit name; "+
+								"(3) use `pulumi import` to adopt this resource; "+
+								"(4) use a Patch resource to manage specific fields",
+							c.Inputs.GetNamespace(), name, c.Inputs.GetKind(),
+						)
+					}
+					if !apierrors.IsNotFound(getErr) {
+						logger.V(1).Infof("pre-create GET for %s/%s failed: %v",
+							c.Inputs.GetNamespace(), name, getErr)
+					}
```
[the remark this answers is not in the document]
guineveresaenger: Addressed.

provider/pkg/provider/provider.go
guineveresaenger: It's the way that the other environment variables are set up. :/ I think to change it we should change it everywhere.

guineveresaenger:
Closing this in favor of https://github.com/pulumi/pulumi-kubernetes/pull/4278.

PR #4278, opened 2026-04-13
Use client-side create to prevent silent upsert of existing objects

### LINKED ISSUE #2926: Fail-fast when a program attempts to create a namespace that already exists.

guineveresaenger:
This is basically the reverse of #3000 - we make upsert behavior an opt-in flag, and use client-side create for creating resources.

This will cause some new stacks to run into Create errors, but it closes multiple data loss issues. Customers wanting to retain server-side apply may do so via the `upsertExistingObjects` flag.

Pulumi updates will continue to use server-side apply.

I've verified no diff is observed on a SSA Update after CSA Create. 

## Summary

When server-side apply is enabled, the provider used SSA Apply (Patch) for all creates, which silently updated existing objects instead of returning an error. This caused data loss when renaming a Pulumi resource with the same `metadata.name` — the engine planned a create-then-delete, the create upserted the existing object, and the delete removed it.

This PR switches to client-side create (which returns `AlreadyExists` natively) when server-side apply is enabled but `upsertExistingObjects` is false (the default). Server-side apply is still used for updates and diffs. The existing `fixCSAFieldManagers` migration handles the field manager transition from client-side create to server-side apply update transparently.

- A new provider config `upsertExistingObjects` (default `false`) restores the previous upsert behavior for users who intentionally adopt existing cluster resources into Pulumi.
- Can also be set via `PULUMI_K8S_UPSERT_EXISTING_OBJECTS=true`.
- To manage fields on existing objects without owning their lifecycle, use Patch resources (e.g. `NamespacePatch`).

Fixes #2926
Refs #2948

## Test plan

- [x] Unit test: create fails with AlreadyExists when object exists
- [x] Unit test: create succeeds with `upsertExistingObjects: true`
- [x] Unit test: create succeeds when object does not exist
- [x] E2E test: creating the built-in `default` namespace fails
- [x] E2E test: client-side create to server-side apply update transition produces no spurious diffs
- [x] E2E test: renaming a Pulumi resource with the same metadata.name fails instead of causing data loss
- [x] Existing SSA conflict and creation tests pass
- [x] Manual repro: CSA create → SSA update → no changes on re-run


blampe (approved):
If this makes it harder to get into a state that could cause #2948 then I think the behavioral change is well justified -- #2948 is a huge foot-gun / outage waiting to happen.

pgavlin (approved):

PR #4333, opened 2026-05-12
Implement List for Kubernetes resources

### LINKED ISSUE #4332: Implement `List` for Kubernetes resources
[pulumi/pulumi#22693](https://github.com/pulumi/pulumi/pull/22693) added the [`List` RPC](https://github.com/pulumi/pulumi/blob/b8d0a0595a16729017728bc81a4cc0fd2e77d702/proto/pulumi/provider.proto#L175) (`pulumirpc.ResourceProvider.List`) and the [`listInputs` field on `ResourceSpec`](https://github.com/pulumi/pulumi/blob/b8d0a0595a16729017728bc81a4cc0fd2e77d702/pkg/codegen/schema/schema.go#L1877) to the provider protocol. `pulumi-kubernetes` does not yet declare `listInputs` on its resources or serve `List`.

[Frassle](https://github.com/Frassle) has implementation work in flight on the [`fraser/list`](https://github.com/pulumi/pulumi-kubernetes/compare/master...fraser/list) branch (not yet open as a PR).

Work:

- [ ] Decide which underlying Kubernetes API(s) to enumerate against, and which resource types to enable first.
- [ ] Define `listInputs` for the enabled types.
- [ ] Implement `List` dispatch, returning streamed `Result` and `Continuation` responses per the proto contract.
- [ ] Tests covering paginated and filter-driven enumeration.

Tracked by [pulumi/pulumi#3742](https://github.com/pulumi/pulumi/issues/3742). Engine support landed in [pulumi/pulumi#22693](https://github.com/pulumi/pulumi/pull/22693).

guineveresaenger:
This pull request implements the [`List` provider RPC](https://github.com/pulumi/pulumi/pull/22693)
for Kubernetes resources.

- Adds `kubeProvider.List method that resolves the given token to a GVK, maps to a RESTMapping, and streams `Result` / `Continuation` from the K8s dynamic client.
- Defines `listInputs`:
  - `name`
  - `labelSelector`
  - `fieldSelector`
  - `namespace` (will error for cluster-scoped kinds)
ListInputs are set on any Kubernetes kind that is Listable, that is every non-nested resource that is not a Patch resource.
- Query supports top-level keys and a `metadata.{namespace,name}` fallback.
- Unit test coverage of validation, filters, Patch-suffix stripping, cluster-scoped guard, 
- Round-trip in-process gRPC test

## Test plan
- [x] `go test ./provider/pkg/provider/ -run "TestList_|TestParseListQuery|TestGvkFromTypeToken"` passes
- [x] Manually exercised against a real cluster via a direct gRPC client
- [x] `make schema` regenerates 341 `listInputs` blocks on non-Patch resources

## Known gaps
- Unit-level verification that `page_size` is forwarded to `ListOptions.Limit` is deferred because the fake k8s client doesn't honor Limit.
- No user-facing language SDK surface yet; schema metadata only. Awaiting language codegen upstream.

Builds on the runtime draft from @Frassle.

Closes #4332.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Added
 
 - [#2744](https://github.com/pulumi/pulumi-kubernetes/issues/2744) Advertise a `helm` mapping for `terraform` conversion so `pulumi import --from terraform` and `pulumi convert --from terraform` recognize `helm_release` and emit `kubernetes:helm.sh/v3:Release`.
+- [#4332](https://github.com/pulumi/pulumi-kubernetes/issues/4332) Implement the `List` provider RPC and advertise `listInputs` (`namespace`, `name`, `labelSelector`, `fieldSelector`) on every non-nested resource. `namespace` is omitted from `listInputs` on cluster-scoped kinds.
 
 ### Fixed
 
--- a/provider/pkg/gen/schema.go
+++ b/provider/pkg/gen/schema.go
@@ -516,7 +516,11 @@ func PulumiSchema(swagger map[string]any, opts ...schemaGeneratorOption) pschema
 					ObjectTypeSpec:     objectSpec,
 					DeprecationMessage: kind.DeprecationComment(),
 					InputProperties:    map[string]pschema.PropertySpec{},
-					ListInputs:         listInputsSpec(kind.kind),
+				}
+				// *List envelope kinds (PodList, DeploymentList, ...) are not addressable REST
+				// resources; only their base kind can be listed.
+				if !kind.isList {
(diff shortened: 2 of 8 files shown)


Frassle (commented):
The limit logic looks suspect, and isn't tested. Given a cluster with 100 resources I should be able to do a list query with page_size=25, limit=50 and get back one response with 25 items and a continuation token, then using that token get another response of 25 items and _no_ token, because we're finished.
Likewise if page_size was 20 we'd get two responses of 20 items each, then a final response of 10 items for the overall 50 item limit.

Frassle (approved):
Looks reasonable to me from core view of things

Issue #4356, opened 2026-05-18
Move integration tests to `KinD` on default


guineveresaenger:
Three PoCs (#4335, #4342, #4344) show every `SkipIfShort`-gated test can run on `KinD` with the right setup. The PR/master coverage gap is exactly these five tests:

| Gate | Closed by |
|---|---|
| MetalLB | `TestPreviewWithApply`, `TestIstio`, `TestDotnet_Guestbook` (#4335) |
| + Traefik | `TestAccIngress` (#4342, also trims bloated testdata) |
| + Calico (replaces kindnet) | `TestEmptyItemNormalization` (#4344) |

## Proposal
PRs run on `KinD` with MetalLB + Traefik + Calico. `release.yml` / `prerelease.yml` keep running against GKE unchanged.

## How
1. Bundle the three PoC setups into one `scripts/setup-kind-cluster.sh` in this repo. ci-mgmt calls it as a single step — heredocs out of YAML, version pins owned here, same pattern as `scripts/ci-cluster-{create,destroy}.sh`.
2. Drop `SkipIfShort` on the five tests; trim `TestAccIngress` testdata. PoCs do this individually; consolidate on merge.

guineveresaenger:
Closed via https://github.com/pulumi/pulumi-kubernetes/pull/4369

PR #4368, opened 2026-05-19
Add `scripts/setup-kind-cluster.sh` for `KinD`-on-default work

guineveresaenger:
Setup for #4356.

- `scripts/setup-kind-cluster.sh` — installs Calico → MetalLB → Traefik on a freshly-created `KinD` cluster. Idempotent (uses `helm upgrade --install`).
- `scripts/kind-config.yaml` — `disableDefaultCNI: true` so Calico can replace kindnet.

[Companion ci-mgmt PR](https://github.com/pulumi/ci-mgmt/pull/2249) wires this into the `run-acceptance-tests.yml` template. 

Once ci-mgmt changes land in this repo, follow-up PRs will unskip the five `SkipIfShort` tests (proofs of concept: #4335, #4342, #4344).

### DIFF
--- a/scripts/setup-kind-cluster.sh
+++ b/scripts/setup-kind-cluster.sh
@@ -5,12 +5,14 @@ CALICO_VERSION="v3.28.2"
 METALLB_VERSION="v0.14.8"
 
 echo "Installing Calico..."
-kubectl apply -f "https://raw.githubusercontent.com/projectcalico/calico/${CALICO_VERSION}/manifests/calico.yaml"
+gh api -H "Accept: application/vnd.github.raw" "/repos/projectcalico/calico/contents/manifests/calico.yaml?ref=${CALICO_VERSION}" > /tmp/calico.yaml
+kubectl apply -f /tmp/calico.yaml
 kubectl -n kube-system rollout status daemonset/calico-node --timeout=5m
 kubectl -n kube-system rollout status deployment/calico-kube-controllers --timeout=5m
 
 echo "Installing MetalLB..."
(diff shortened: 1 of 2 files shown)


scripts/setup-kind-cluster.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+set -o nounset -o errexit -o pipefail
+
+CALICO_VERSION="v3.28.2"
+METALLB_VERSION="v0.14.8"
+
+echo "Installing Calico..."
+kubectl apply -f "https://raw.githubusercontent.com/projectcalico/calico/${CALICO_VERSION}/manifests/calico.yaml"
```
pose: Does `kubectl` retry if it fails to download the file? I am asking if it's worth using the `gh` CLI to obtain that artifact instead and account for retries.

scripts/setup-kind-cluster.sh
```diff
@@ -0,0 +1,41 @@
...
+kubectl -n kube-system rollout status daemonset/calico-node --timeout=5m
+kubectl -n kube-system rollout status deployment/calico-kube-controllers --timeout=5m
+
+echo "Installing MetalLB..."
+kubectl apply -f "https://raw.githubusercontent.com/metallb/metallb/${METALLB_VERSION}/config/manifests/metallb-native.yaml"
```
pose: (same as above)

pose (approved):
Comments are more to increase resiliency

scripts/setup-kind-cluster.sh
guineveresaenger: Kubectl does not - good catch!

PR #4387, opened 2026-05-25
Fix docs for Kubernetes version support and local SDK install

immanuwell:
### Proposed changes

Small docs fix, but its a real footgun.

- align the documented minimum Kubernetes version with the provider check of `v1.13`
- refresh stale docs links
- fix the local `yarn link` flow in `CONTRIBUTING.md` so it matches `make install`

### Repro

1. Open `README.md`. It says clusters `>=1.9.0` are supported.
2. The provider rejects anything below `v1.13` in `provider/pkg/provider/provider.go`.
3. The existing provider suite covers that path too:
   `go test ./pkg/provider -run TestSuite -count=1 -args -ginkgo.focus='server version is < 1.13'`

So yeah, docs were a bit off. This makes them line up again.

### Related issues (optional)

None found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## Unreleased
 
+### Changed
+
+- Align the documented minimum supported Kubernetes version and local SDK install instructions with current provider behavior.
+
 ## 4.31.1 (May 19, 2026)
 
 ### Changed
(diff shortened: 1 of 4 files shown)


guineveresaenger (approved):
Thank you! Really appreciate this cleanup. <3

Issue #3498, opened 2025-02-17
renderYamlToDirectory is still dependent on a kubernetes cluster connection


trondhindenes:
### What happened?

There are several situations where pulumi errors when using `renderYamlToDirectory` - althought it's supposed to be a completely "offline" way to generate kubernetes manifests:

If the "KUBECONFIG" environment variable is set (a very common way of pointing kubectl to a config), pulumi errors out:
```
error: pulumi:providers:kubernetes resource 'minioProvider': property kubeconfig value {<path-to-kubeconfig-file>} has a problem: "kubeconfig" arg is not compatible with "renderYamlToDirectory" arg
```

if the active kubeconfig is pointing at a cluster that is not reachable for some reason (for example, expired credentials): This does not cause errors, but the error from kubeconfig is printed, so it's a clear indication that pulumi attempts to communicate with the configured cluster

### Example

See above

### Output of `pulumi about`

```
CLI          
Version      3.136.1
Go Version   go1.23.2
Go Compiler  gc

Plugins
KIND      NAME        VERSION
resource  kubernetes  4.21.1
language  nodejs      unknown

```

### Additional context

When using `renderYamlToDirectory`, pulumi should avoid any cluster communication using kubectl.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Thanks for reporting this issue @trondhindenes and apologies that you're facing it. Please note that `rendYamlToDirectory` support is [still a beta feature](https://www.pulumi.com/registry/packages/kubernetes/api-docs/provider/#renderyamltodirectory_nodejs). We'll add this to our backlog to further investigate and resolve.

trondhindenes:
thanks!

jkodroff:
Could we at least document that the 2 options are incompatible and state why? I remember hearing that `rendYamlToDirectory` is not as offline as one might think b/c the provider would need cluster context for CRs or CRDs (I forget which).

mjeffryes:
I think @jkodroff is right, that this is "by design", but we really ought to document this clearly. I'm going to adjust the labels to reflect that.

PR #4406, opened 2026-05-30
fix: allow ambient KUBECONFIG in render mode

### LINKED ISSUE #4399: KUBECONFIG env var is rejected but default ~/.kube/config works
# Pulumi Kubernetes: KUBECONFIG env var incompatible with render_yaml_to_directory but default ~/.kube/config works

## Summary
When using `render_yaml_to_directory` mode, the Pulumi Kubernetes provider rejects the `KUBECONFIG` environment variable but successfully uses the default `~/.kube/config` location. This creates confusion and breaks common CI/CD workflows where tools set `KUBECONFIG` explicitly.

## Environment
- **Pulumi CLI**: v3.243.0
- **pulumi-kubernetes**: v4.31.1
- **Python**: 3.12.13
- **OS**: Ubuntu 22.04 (GitHub Actions ubuntu-latest)
- **Helm Chart version**: v4

## Error Message
```
error: pulumi:providers:kubernetes resource 'context-render': 
property kubeconfig value {/home/runner/work/oi-context/oi-context/gha-kubeconfig-504481f49fd78a91} has a problem: 
"kubeconfig" arg is not compatible with "renderYamlToDirectory" arg
```

### Scenario 1: With KUBECONFIG env var (fails)

```yaml
# GitHub Actions workflow
- name: GKE auth
  uses: google-github-actions/get-gke-credentials@v3
  with:
    cluster_name: "my-cluster"
    location: "us-central1"
  # This action sets KUBECONFIG=/some/temp/path

- name: Render with Pulumi
  run: pulumi up --yes
  # FAILS with "kubeconfig arg is not compatible with renderYamlToDirectory"
```

**Python code:**
```python
import pulumi_kubernetes as k8s
from pulumi_kubernetes.helm.v4 import Chart, ChartArgs

provider = k8s.Provider(
    "k8s-render",
    render_yaml_to_directory="./dist",
)

chart = Chart(
    "my-chart",
    ChartArgs(
        chart="./helm-chart",
        namespace="default",
    ),
    opts=pulumi.ResourceOptions(provider=provider),
)
```

### Scenario 2: With default ~/.kube/config (works)

```yaml
# GitHub Actions workflow
- name: GKE auth
  uses: google-github-actions/get-gke-credentials@v3
  with:
    cluster_name: "my-cluster"
    location: "us-central1"

- name: Copy kubeconfig to default location
  run: |
    mkdir -p ~/.kube
    cp "$KUBECONFIG" ~/.kube/config

- name: Render with Pulumi
  env:
    KUBECONFIG: ""  # Explicitly unset
  run: pulumi up --yes
  # SUCCEEDS - renders manifests correctly
```

**Same Python code as above - no changes needed**

## Expected Behavior

When `render_yaml_to_directory` is set:
1. The provider should accept `KUBECONFIG` environment variable just like it accepts `~/.kube/config`
2. OR the error message should clearly explain that `KUBECONFIG` env var is not supported and suggest using the default location
3. The behavior should be consistent: either reject both or accept both

## Actual Behavior

- ❌ **With `KUBECONFIG` env var set**: Provider rejects with error even though the file is valid
- ✅ **With `~/.kube/config` (default location)**: Provider works correctly
- The provider reads from the same kubeconfig file in both cases, but treats them differently based on how it's referenced

## Root Cause Analysis

The provider appears to:
1. Check if `KUBECONFIG` env var is set
2. If set, validate it against `render_yaml_to_directory` compatibility and reject
3. If not set, fall back to `~/.kube/config` and allow it

This creates an inconsistency where the **source** of the kubeconfig path matters more than whether it's actually being used.

### High Impact on CI/CD Workflows

This breaks common CI/CD patterns where:
- **GKE credentials action** sets `KUBECONFIG` automatically (Google's official action)
- **ACK credentials action** sets `KUBECONFIG` automatically (Alibaba's official action)
- **EKS credentials action** sets `KUBECONFIG` automatically (AWS's patterns)
- Users need cluster connectivity for CRD validation (see related issue #3498)

### Workaround Required

Teams must add an extra step to copy kubeconfig:
```bash
mkdir -p ~/.kube
cp "$KUBECONFIG" ~/.kube/config
unset KUBECONFIG
```

This is unintuitive and not documented.

### Option 1: Accept KUBECONFIG env var (Recommended)
Allow `KUBECONFIG` environment variable in render mode, treating it the same as `~/.kube/config`:
```python
# Should work in render mode
os.environ['KUBECONFIG'] = '/path/to/kubeconfig'
provider = k8s.Provider(
    "k8s-render",
    render_yaml_to_directory="./dist",
)
```

### Option 2: Improve error message
If `KUBECONFIG` must remain unsupported, provide a clearer error:
```
error: KUBECONFIG environment variable is not supported with renderYamlToDirectory.
Please copy your kubeconfig to ~/.kube/config or unset the KUBECONFIG environment variable.
```

### Option 3: Add skip_kubeconfig_env flag
```python
provider = k8s.Provider(
    "k8s-render",
    render_yaml_to_directory="./dist",
    skip_kubeconfig_env=True,  # Ignore KUBECONFIG env var
)
```

## Related Issues

- #3498: renderYamlToDirectory depends on kubernetes cluster connection (explains why kubeconfig is needed)
- #3176: CRD validation in same run

### Why Cluster Access is Needed in Render Mode

Per issue #3498, `render_yaml_to_directory` mode requires cluster connectivity to:
1. Query CRD definitions to determine if resources are namespaced or cluster-scoped
2. Validate custom resource manifests

This means kubeconfig **is actually required** for render mode when using CRDs, but the provider's handling of `KUBECONFIG` env var creates unnecessary friction.

## Test Case

```python
import os
import tempfile
import pulumi
import pulumi_kubernetes as k8s
from pulumi_kubernetes.helm.v4 import Chart, ChartArgs

# Test 1: With KUBECONFIG env var (currently fails)
def test_with_kubeconfig_env():
    os.environ['KUBECONFIG'] = '/tmp/test-kubeconfig'
    
    provider = k8s.Provider(
        "test-render",
        render_yaml_to_directory="/tmp/render-output",
    )
    
    chart = Chart(
        "test-chart",
        ChartArgs(chart="./chart", namespace="default"),
        opts=pulumi.ResourceOptions(provider=provider),
    )
    # Expected: Should work
    # Actual: Fails with "kubeconfig arg is not compatible"

# Test 2: With ~/.kube/config (currently works)
def test_with_default_kubeconfig():
    # Ensure KUBECONFIG is not set
    os.environ.pop('KUBECONFIG', None)
    
    # Copy kubeconfig to default location
    os.makedirs(os.path.expanduser('~/.kube'), exist_ok=True)
    # ... copy file ...
    
    provider = k8s.Provider(
        "test-render",
        render_yaml_to_directory="/tmp/render-output",
    )
    
    chart = Chart(
        "test-chart",
        ChartArgs(chart="./chart", namespace="default"),
        opts=pulumi.ResourceOptions(provider=provider),
    )
    # Expected: Should work
    # Actual: Works correctly
```

## Expected Result
Both test cases should succeed - the provider should treat `KUBECONFIG` env var and `~/.kube/config` equivalently when `render_yaml_to_directory` is set.

immanuwell:
### Proposed changes

Allow render mode to accept `kubeconfig` when it is only the ambient `KUBECONFIG` value injected by the SDK.

This path is pretty normal in CI. `google-github-actions/get-gke-credentials` exports `KUBECONFIG` out of the box. Before this PR, render mode tripped on that and failed with `"kubeconfig" arg is not compatible with "renderYamlToDirectory" arg`. After this PR, that ambient path is allowed, while explicit `kubeconfig` is still blocked. Added a regression test too.

Repro:
1. Set `KUBECONFIG` to a valid kubeconfig file.
2. Create a provider with `renderYamlToDirectory` and do not pass `kubeconfig` yourself.
3. The SDK fills `kubeconfig` from `KUBECONFIG`, then `CheckConfig` fails.
4. With this PR, render mode works as expected.

### Related issues (optional)

Fixes #4399
Related #3498

### DIFF
--- a/provider/pkg/provider/provider_checkconfig_test.go
+++ b/provider/pkg/provider/provider_checkconfig_test.go
@@ -151,6 +151,27 @@ var _ = gk.Describe("RPC:CheckConfig", func() {
 			})
 		})
 
+		gk.Context("when kubeconfig does not match ambient KUBECONFIG", func() {
+			gk.BeforeEach(func() {
+				path := WriteKubeconfigToFile(config)
+				gm.Expect(os.Setenv("KUBECONFIG", path)).To(gm.Succeed())
+				gk.DeferCleanup(func() {
+					os.Unsetenv("KUBECONFIG")
+				})
+				news["kubeconfig"] = resource.NewStringProperty(WriteKubeconfigToString(config))
+			})
+			gk.It("should fail because an explicit kubeconfig is disallowed", func() {
+				resp, err := k.CheckConfig(context.Background(), req)
+				gm.Expect(err).ToNot(gm.HaveOccurred())
+				gm.Expect(resp.Failures).To(gm.HaveExactElements(
+					CheckFailure(
+						"kubeconfig",
+						gm.Equal(`"kubeconfig" arg is not compatible with "renderYamlToDirectory" arg`),
+					),
+				))
+			})
+		})
+
 		gk.Context("when context is specified", func() {
 			gk.BeforeEach(func() {
 				news["context"] = resource.NewStringProperty(config.CurrentContext)
--- a/provider/pkg/provider/provider_checkconfig_test.go
+++ b/provider/pkg/provider/provider_checkconfig_test.go
@@ -16,6 +16,7 @@ package provider
 
 import (
 	"context"
+	"os"
 
 	_ "embed"
 
@@ -131,6 +132,22 @@ var _ = gk.Describe("RPC:CheckConfig", func() {
 						gm.Equal(`"kubeconfig" arg is not compatible with "renderYamlToDirectory" arg`),
 					),
 				))
+			})
+		})
+
+		gk.Context("when kubeconfig matches ambient KUBECONFIG", func() {
+			gk.BeforeEach(func() {
+				path := WriteKubeconfigToFile(config)
+				gm.Expect(os.Setenv("KUBECONFIG", path)).To(gm.Succeed())
+				gk.DeferCleanup(func() {
+					os.Unsetenv("KUBECONFIG")
+				})
+				news["kubeconfig"] = resource.NewStringProperty(path)
+			})
+			gk.It("should succeed", func() {
+				resp, err := k.CheckConfig(context.Background(), req)
+				gm.Expect(err).ToNot(gm.HaveOccurred())
+				gm.Expect(resp.Failures).To(gm.BeEmpty())
 			})
 		})
 
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -415,7 +415,9 @@ func (k *kubeProvider) CheckConfig(
 				Reason:   fmt.Sprintf(errTemplate, "context"),
 			})
 		}
-		if truthyValue("kubeconfig", news) {
+		// SDKs default kubeconfig from $KUBECONFIG. In render mode we still want ambient
+		// kubeconfig loading to work, so don't reject that specific defaulted value.
+		if truthyValue("kubeconfig", news) && !matchesAmbientKubeconfigEnv(news["kubeconfig"]) {
 			failures = append(failures, &pulumirpc.CheckFailure{
 				Property: "kubeconfig",
 				Reason:   fmt.Sprintf(errTemplate, "kubeconfig"),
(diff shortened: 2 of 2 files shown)


provider/pkg/provider/provider_checkconfig_test.go
```diff
@@ -134,6 +135,22 @@ var _ = gk.Describe("RPC:CheckConfig", func() {
 			})
 		})
 
+		gk.Context("when kubeconfig matches ambient KUBECONFIG", func() {
```
guineveresaenger, line 138: Could you add a negative case here as well?

guineveresaenger (changes requested):
hi @immanuwell - thank you for this pull request.

Could you add a negative scenario check as well? Otherwise this LGTM; running tests to confirm.

guineveresaenger (approved):
I went ahead and added the test for you; going ahead and merging this..

PR #4417, opened 2026-06-04
Resolve custom resource namespace scope from Helm release CRDs

### LINKED ISSUE #3176: failed to determine if the following GVK is namespaced if CRD is created in the same run

### LINKED ISSUE #3299: Release resource should integrate with the CRD cache
## Hello!

## Issue details

The provider maintains a cache of CRDs across resources, to enable various resources to determine whether a given kind is namespaced. For example, the `ConfigFile` resource creates a child resource for each object in the manifest file, and assigns a resource base derived from the object's name. If the kind is namespaced, the context namespace (e.g. `default`) must be included in the name, in order to arrive at a canonical name.  This, in turn, affects other features like the supported `config.kubernetes.io/depends-onn` annotation ([ref](https://www.pulumi.com/blog/kubernetes-yaml-v2/)). See https://github.com/pulumi/pulumi-kubernetes/pull/2904 for more information.

At present, the Release resource doesn't contribute to the CRD cache, because it doesn't create child resources (e.g. of type [kubernetes.apiextensions.k8s.io/v1.CustomResourceDefinition](https://www.pulumi.com/registry/packages/kubernetes/api-docs/apiextensions/v1/customresourcedefinition/)). 
 
As I outlined in https://github.com/pulumi/pulumi-kubernetes/pull/3186#pullrequestreview-2255975782, the `Release` resource should populate the CRD cache.

guineveresaenger:
When a custom resource's CRD is installed by a Helm Release in the same program, the provider can't tell at preview whether the resource is namespaced or cluster-scoped: a Release installs its CRDs server-side, so they never land in the provider's CRD cache the way a CustomResourceDefinition resource or a helm.sh/v4:Chart child does. 

This reads a release's CRDs into the shared cache during Check, from both the chart's crds/ directory and its rendered templates, so a custom resource elsewhere in the program resolves the real scope on the first preview. Rendering is gated by a cheap text scan, so only charts that actually template a CustomResourceDefinition are rendered, and it's best-effort. 
  
Fixes #3176.
Fixes https://github.com/pulumi/pulumi-kubernetes/issues/3299.
Also fixes https://github.com/pulumi-labs/pulumi-nvidia-aicr/issues/6.

### DIFF
--- a/provider/pkg/provider/helm_release.go
+++ b/provider/pkg/provider/helm_release.go
@@ -15,7 +15,6 @@
 package provider
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -498,28 +497,19 @@ func (r *helmReleaseProvider) cacheReleaseCRDs(ctx context.Context, chart *helmc
 	if r.clientSet == nil || chart == nil {
 		return
 	}
-	cache := &r.clientSet.CRDCache
-
-	// CRDs in the chart's crds/ directory.
-	if err := cacheChartCRDs(ctx, chart, cache); err != nil {
-		logger.V(9).Infof("caching crds/ CRDs for release %q: %v", rel.Name, err)
-	}
-
-	// CRDs templated into the chart's manifest.
-	if !chartTemplatesMentionCRD(chart) {
-		return
-	}
 	manifest, err := r.renderChartManifest(chart, rel)
 	if err != nil {
 		logger.V(9).Infof("rendering chart to find CRDs for release %q: %v", rel.Name, err)
 		return
 	}
-	if err := cacheCRDsFromManifest(ctx, manifest, cache); err != nil {
-		logger.V(9).Infof("caching templated CRDs for release %q: %v", rel.Name, err)
+	if err := cacheCRDsFromManifest(ctx, manifest, &r.clientSet.CRDCache); err != nil {
+		logger.V(9).Infof("caching CRDs for release %q: %v", rel.Name, err)
 	}
 }
 
-// renderChartManifest runs a client-side helm template (no cluster) and returns the rendered manifest.
+// renderChartManifest runs a client-side helm template (no cluster) and returns the rendered
+// manifest. IncludeCRDs (honoring skipCrds) makes Helm emit the crds/ CRDs across the chart and its
+// subcharts; templated CRDs render regardless.
 func (r *helmReleaseProvider) renderChartManifest(chart *helmchart.Chart, rel *Release) (string, error) {
 	conf, err := r.getActionConfig(rel.Namespace)
 	if err != nil {
@@ -528,6 +518,7 @@ func (r *helmReleaseProvider) renderChartManifest(chart *helmchart.Chart, rel *R
 	install := action.NewInstall(conf)
 	install.ClientOnly = true
 	install.DryRun = true
+	install.IncludeCRDs = !rel.SkipCrds
 	install.ReleaseName = rel.Name
 	if install.ReleaseName == "" {
 		install.ReleaseName = "release-name"
@@ -545,18 +536,6 @@ func (r *helmReleaseProvider) renderChartManifest(chart *helmchart.Chart, rel *R
 		return "", nil
 	}
 	return rendered.Manifest, nil
-}
-
-func cacheChartCRDs(ctx context.Context, chart *helmchart.Chart, cache *clients.CRDCache) error {
-	if chart == nil || cache == nil {
-		return nil
-	}
-	for _, crd := range chart.CRDObjects() {
-		if err := cacheCRDsFromManifest(ctx, string(crd.File.Data), cache); err != nil {
-			return fmt.Errorf("parsing CRD file %q: %w", crd.Name, err)
-		}
-	}
-	return nil
 }
 
 func cacheCRDsFromManifest(ctx context.Context, manifest string, cache *clients.CRDCache) error {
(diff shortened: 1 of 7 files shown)


provider/pkg/provider/helm_release.go
```diff
@@ -487,6 +491,112 @@ func (r *helmReleaseProvider) helmLoad(
...
+// elsewhere in the program can resolve their namespace scope at preview, before the CRD is installed
+// on the cluster.
+func (r *helmReleaseProvider) cacheReleaseCRDs(ctx context.Context, chart *helmchart.Chart, rel *Release) {
+	if r.clientSet == nil || chart == nil {
+		return
+	}
+	cache := &r.clientSet.CRDCache
+
+	// CRDs in the chart's crds/ directory.
+	if err := cacheChartCRDs(ctx, chart, cache); err != nil {
```
corymhall: Do we need to take into account `skipCrds`? e.g. if `skipCrds = true` should we skip writing to the cache?

provider/pkg/provider/helm_release.go
```diff
@@ -487,6 +491,112 @@ func (r *helmReleaseProvider) helmLoad(
...
+		return "", nil
+	}
+	return rendered.Manifest, nil
+}
+
+func cacheChartCRDs(ctx context.Context, chart *helmchart.Chart, cache *clients.CRDCache) error {
+	if chart == nil || cache == nil {
+		return nil
+	}
+	for _, crd := range chart.CRDObjects() {
```
corymhall: Does this take into account disabled subchart CRDs?

tests/sdk/yaml/helm_release_crd_scope_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/common/apitype"
+)
+
+func TestHelmReleaseCRDScopeWritesToCache(t *testing.T) {
+	test := pulumitest.NewPulumiTest(t, "testdata/helm-release-crd-scope", opttest.SkipInstall())
+	t.Cleanup(func() {
+		test.Destroy(t)
+	})
+	test.Preview(t)
+	test.Up(t)
```
corymhall, line 25: Should this test assert on the preview results instead of `up` (or both)? Is it possible that the preview could still show the wrong value?

provider/pkg/provider/helm_release.go
```diff
@@ -386,6 +388,8 @@ func (r *helmReleaseProvider) Check(
 			// as opposed to the program input (which is a constraint such as ">= 1.2.3").
 			// with this we may determine whether the Helm release needs to be upgraded.
 			newRelease.Version = chart.Metadata.Version
+
+			r.cacheReleaseCRDs(ctx, chart, newRelease)
```
corymhall, line 391: Is it possible for this to be more permissive? e.g. do we only need certain fields to be known?

corymhall (approved):
I think my main questions are around whether it matters if we cache a CRD that is skipped or disabled. I don't understand enough about how this provider works, but just thought to call it out in case it does matter.

tests/sdk/yaml/helm_release_crd_scope_test.go
guineveresaenger: due to https://github.com/pulumi/pulumi-kubernetes/pull/4400, the preview will not fail but emit a warning instead. We can assert on the warning but it's not really focused on the caching then anymore.

provider/pkg/provider/helm_release.go
guineveresaenger: heh - this made me realize we can make this a whole lot simpler by just using install.IncludeCRDs instead of handrolling discovery.

PR #4426, opened 2026-06-08
Render Helm hooks in helm.v4.Chart when using renderYamlToDirectory

### LINKED ISSUE #3284: helm.v4.Chart: make helm hooks optional (instead of force disabled)

ehsan310:
## Scope

> **This does not fully resolve #3284.** It does not implement Helm hook *lifecycle* support (ordering, weights, delete policies, execution) — that remains open in #3284 (and #555 for live apply). This PR only brings `helm.v4.Chart` **render mode** up to par with `helm template` by letting hook resources be emitted instead of silently dropped. Relates to #3284.

## Problem

`kubernetes.helm.sh/v4:Chart` renders a chart with the Helm SDK but discards all Helm hook resources (anything annotated `helm.sh/hook`): only `release.Manifest` is parsed and `release.Hooks` is ignored.

For a normal apply this is intentional (Pulumi can't currently honor hook semantics). **But** in `renderYamlToDirectory` mode Pulumi is *not* applying anything — it only emits YAML for another tool (e.g. Argo CD) to apply. There, the output should mirror `helm template`, which includes hook resources so the downstream tool can act on them. Today they vanish, which breaks GitOps-style workflows that pack the rendered YAML into an OCI artifact for Argo CD.

## Change

Add an explicit, opt-in `includeHooks` boolean to `helm.v4.Chart` (default `false`).

- When `includeHooks: true` **and** the provider is in `renderYamlToDirectory` mode, hook resources are appended to the rendered output as-is, with their `helm.sh/hook` annotations preserved.
- If `includeHooks` is set **without** render mode, it is ignored and a warning is logged — hooks are never silently registered as live Pulumi-managed resources.
- Default behavior is unchanged: with `includeHooks` unset, hooks are dropped exactly as before.
- Test hooks (`helm.sh/hook: test`) are always excluded (detected via `hook.Events` / `release.HookTest`, not a regexp).

The flag lives on the Chart (mirroring the existing chart-level `skipCrds`, and v3's `includeTestHookResources`). Render mode is plumbed to the component via a `RenderYAMLToDirectory` field on `ResourceProviderOptions`, wired from the provider's existing `yamlRenderMode`.

## Tests

- Unit (`chart_test.go`): `includeHooks` unset → all hooks dropped; `includeHooks: true` + render mode → non-test hook included, test hook excluded; `includeHooks: true` without render mode → hooks ignored + warning logged.
- Full render integration (`TestHelmChartV4RenderYamlIncludesHooks`, modeled on `TestHelmChartV4WithYamlRenderModeRendersWithoutClusterConnection`) with a new local chart fixture containing a normal resource, a `pre-install` hook, and a `test` hook; asserts the pre-install hook is rendered under `1-manifest/` and the test hook is not.

Schema and all language SDKs were regenerated for the new `includeHooks` input.

### DIFF
--- a/provider/pkg/provider/helm/v4/chart.go
+++ b/provider/pkg/provider/helm/v4/chart.go
@@ -17,11 +17,14 @@ package v4
 import (
 	"context"
 	"fmt"
+	"regexp"
+	"strings"
 
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chartutil"
 	helmkube "helm.sh/helm/v3/pkg/kube"
 	"helm.sh/helm/v3/pkg/postrender"
+	"helm.sh/helm/v3/pkg/release"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/discovery"
 
@@ -248,8 +251,21 @@ func (r *ChartProvider) Construct(
 	}
 
 	// Parse the YAML file into an array of Kubernetes objects.
+	//
+	// Helm hook resources (those annotated with `helm.sh/hook`) are separated
+	// from the regular manifest by the Helm SDK and are not applied by Pulumi.
+	// In render-only mode (`renderYamlToDirectory`) Pulumi is not in charge of
+	// the resource lifecycle; it merely emits YAML for another tool (e.g. ArgoCD)
+	// to apply. In that case, include the hook resources in the output as-is so
+	// they are not silently dropped, mirroring `helm template`. Test hooks
+	// (`helm.sh/hook: test`) are excluded since they are not part of a normal
+	// deployment.
+	manifest := release.Manifest
+	if r.opts.RenderYAMLToDirectory {
+		manifest = manifestWithHooks(release)
+	}
 	parseOpts := provideryamlv2.ParseOptions{
-		YAML: release.Manifest,
+		YAML: manifest,
 	}
 	objs, err := provideryamlv2.Parse(ctx.Context(), parseOpts)
 	if err != nil {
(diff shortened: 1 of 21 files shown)


ehsan310:
Hi @guineveresaenger , are you able to check on this PR and tell me if that make sense ?

provider/pkg/provider/helm/v4/chart.go
```diff
@@ -316,6 +332,25 @@ func preregister(ctx *pulumi.Context, comp *ChartState, obj *unstructured.Unstru
 	return obj, resourceOpts
 }
 
+// testHookAnnotation matches test-related Helm hook annotations (test, test-success, test-failure).
+var testHookAnnotation = regexp.MustCompile(`"?helm.sh/hook"?:.*test`)
```
guineveresaenger: is it possible to use a helper function via hook.Events to determine if it's a `test` Hook? I'd love to not use regexp here.

guineveresaenger (changes requested):
hi @ehsan310 - thank you for this pull request.

A housekeeping note: this does not (fully) address https://github.com/pulumi/pulumi-kubernetes/issues/3284. This is fine, from a design standpoint!
Rather, it brings Helm v4 up to par with `helm template`, where hooks are rendered. #3284 is a much more involved feature, if it is even desirable from a Pulumi standpoint.

I'd prefer seeing full render tests for this feature, like TestRenderYAML (tests/sdk/nodejs/nodejs_test.go) or TestHelmChartV4WithYamlRenderModeRendersWithoutClusterConnection (tests/sdk/nodejs/helm_test.go).

Also, and this is due to the design of render mode, I think this feature needs to be an explicit top-level opt-in flag. Rendering Hooks by default would silently register the hook resources as Pulumi-managed resources into stack state. This would break workflows that expect that anything that is rendered is able to be consumed by Pulumi, which Hooks are currently not.

provider/pkg/provider/helm/v4/chart_test.go
```diff
@@ -410,6 +410,77 @@ var _ = gk.Describe("Construct", func() {
...
+							gm.Not(gm.ContainElement(preInstallHook)),
+							gm.Not(gm.ContainElement(testHook)),
+						)),
+					}))
+					var msgs []string
+					for _, l := range tc.engine.Logs() {
+						msgs = append(msgs, l.GetMessage())
+					}
+					gm.Expect(msgs).To(gm.ContainElement(gm.ContainSubstring("includeHooks")))
+				})
```
guineveresaenger: Can we add one more scenario, to show what happens when `includeHooks` is set but render mode isn't?

guineveresaenger (commented):
this looks better - one more test scenario

provider/pkg/provider/helm/v4/chart_test.go
ehsan310: there are 3 test cases 


```
includeHooks unset (default)  = all hooks dropped
includeHooks: true + render mode = non-test hook rendered, test hook excluded
includeHooks: true, render mode off = hooks dropped, normal resources still rendered, warning containing includeHooks logged
```

guineveresaenger (approved):

ehsan310:
@guineveresaenger pushed the linter fix

guineveresaenger:
ha! thank you; I wasn't going to bother you with it 😅 🙇‍♀️

PR #4434, opened 2026-06-11
feat: use OpenAPI v3 for schema loading with v2 fallback

### LINKED ISSUE #4433: Performance issue when creating a new instance of a provider
Customer is reporting that whenever they create an new instance of an explicit provider, it takes 4-5 seconds each time. 

Internal slack thread here: https://pulumi.slack.com/archives/C03Q8JM1EBF/p1781162744370819

tl;dr they do have a lot of CRDs and according to my trusty research assistant, this may mean that when we create a new instance of the provider the OpenAPI spec that it downloads can be quite large.

pierskarsenbarg:
## Summary

- Switches `getResources()` to fetch schemas via the OpenAPI v3 per-GV endpoint (`/openapi/v3/{group}/{version}`) instead of the monolithic v2 endpoint (`/openapi/v2`)
- The v3 path benefits from the per-GV caching already wired up in `cachedopenapi.NewClient` inside `memCacheClient.OpenAPIV3()`; the v2 path had no caching at all and re-downloaded the full schema on every `Configure()` call
- Falls back to v2 automatically for clusters older than k8s v1.24 (where v3 is unavailable)

## Test plan

- [x] `go test ./pkg/openapi/...` — 5 new tests covering v3 lookup, unknown GVK, multiple GVs, partial error resilience, and v2/v3 parity
- [x] `go test ./pkg/clients/...` — existing fake client tests pass with new `OpenAPIV3()` override
- [x] `go test ./pkg/provider/...` — existing provider/await tests pass unmodified

Fixes: #4433

### DIFF
--- a/provider/pkg/provider/provider.go
+++ b/provider/pkg/provider/provider.go
@@ -234,9 +234,16 @@ func (k *kubeProvider) getResources() (k8sopenapi.Resources, error) {
 	k.resourcesMutex.Lock()
 	defer k.resourcesMutex.Unlock()
 
-	rs, err := openapi.GetResourceSchemasForClient(k.clientSet.DiscoveryClientCached)
+	// Try v3 first: schemas are fetched per-GV and benefit from the per-GV cache
+	// in cachedopenapi.NewClient, avoiding the full schema download on every Configure().
+	rs, err := openapi.GetResourceSchemasForClientV3(k.clientSet.DiscoveryClientCached.OpenAPIV3())
 	if err != nil {
-		return nil, err
+		// Fall back to v2 for older clusters (< v1.24) or transient v3 failures.
+		logger.V(9).Infof("OpenAPI v3 schema unavailable (%v), falling back to v2", err)
+		rs, err = openapi.GetResourceSchemasForClient(k.clientSet.DiscoveryClientCached)
+		if err != nil {
+			return nil, err
+		}
 	}
 	k.resources = rs
 	return k.resources, nil
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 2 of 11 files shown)


provider/pkg/provider/provider.go
```diff
@@ -234,9 +234,16 @@ func (k *kubeProvider) getResources() (k8sopenapi.Resources, error) {
 	k.resourcesMutex.Lock()
 	defer k.resourcesMutex.Unlock()
 
-	rs, err := openapi.GetResourceSchemasForClient(k.clientSet.DiscoveryClientCached)
+	// Try v3 first: schemas are fetched per-GV and benefit from the per-GV cache
+	// in cachedopenapi.NewClient, avoiding the full schema download on every Configure().
+	rs, err := openapi.GetResourceSchemasForClientV3(k.clientSet.DiscoveryClientCached.OpenAPIV3())
```
guineveresaenger, line 239: What happens if `rs` is empty?

guineveresaenger (commented):
I don't hate this change, but is there a way to benchmark any performance improvements?

guineveresaenger:
Also, the test should pass after you pull in the most recent default branch.

provider/pkg/provider/provider.go
pierskarsenbarg: That's a good question and I will check, but my immediate answer would be that it's no change as if the v2 endpoint returns an empty we don't do any checks on that either. 

At least in this function

pierskarsenbarg:
> I don't hate this change, but is there a way to benchmark any performance improvements?

According to the [k8s docs](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#openapi-interface-definition) the openapi v3 docs are preferred. 

Again, doesn't answer your question  - there could be something we do to add a test that creates a _lot_ of CRDs and then create providers then test the before and after

pierskarsenbarg:
I've done some basic performance testing using this PR vs the released version of the k8s provider. Using the v3 endpoint of the k8s API seems significantly slower than just using the v2, more than likely because it's making multiple HTTP requests, rather than the v2 which gets everything in one request.

So closing this

PR #4437, opened 2026-06-12
Generate extension-parameterized SDKs from Kubernetes CRDs

guineveresaenger:
Adds provider-side support for extension parameterization.

Usage: 
`pulumi package add kubernetes --extension "name=gateway-networking version=1.2.1 crd-manifest=gateway-api-crds.yaml"`

Notes:
- This pull request filters the base `meta/v1` group out of each extension schema and references those types via external ref from the base Kubernetes SDK to avoid collision.
- The `version` parameter is optional; a default version is provided.
- The `name` parameter is optional; we infer a name from the manifest filename as fallback.
- It is possible to specify multiple manifests.

## Test plan
- `pkg/gen` and `pkg/provider` unit tests pass locally.
- An e2e test was added
 but is awaiting https://github.com/pulumi/pulumi-yaml/pull/1162 in order to run.
 - Per https://github.com/pulumi/home/issues/3595, we will add additional e2e tests.


tests/sdk/yaml/extension_test.go
```diff
@@ -0,0 +1,60 @@
...
+	// without downloading an (unpublished) dev provider.
+	add := exec.Command("pulumi", "package", "add", providerBin,
+		"--extension", "name=gateway-networking crd-manifest=gateway-api-crds.yaml")
+	add.Dir = test.WorkingDir()
+	add.Env = os.Environ()
+	if out, err := add.CombinedOutput(); err != nil {
+		t.Fatalf("pulumi package add --extension failed: %v\n%s", err, out)
+	}
+
+	test.Preview(t)
```
pose: (minor) Why is this required? Could we assert anything after this?

tests/sdk/yaml/extension_test.go
```diff
@@ -0,0 +1,60 @@
...
+	base.Env = os.Environ()
+	if out, err := base.CombinedOutput(); err != nil {
+		t.Fatalf("pulumi package add (base) failed: %v\n%s", err, out)
+	}
+
+	// Add the extension from the local provider binary: `package add --extension`
+	// parameterizes the base provider and records the extension in Pulumi.yaml,
+	// without downloading an (unpublished) dev provider.
+	add := exec.Command("pulumi", "package", "add", providerBin,
+		"--extension", "name=gateway-networking crd-manifest=gateway-api-crds.yaml")
```
pose: (nitpick) I assume we are using Pulumi's automation API (e.g. `test.Preview`, `test.Up`), have we considered the trade offs of making this call through the automation API (assuming it has been surfaced) vs using the CLI as we are doing here? I am asking since could there be a conflict between the CLI used here and the `pulumi` version is being used when running `test.Preview` (for instance).

tests/sdk/yaml/extension_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/pulumi/providertest/pulumitest/opttest"
+)
+
+func TestExtensionGatewayAPI(t *testing.T) {
+	test := pulumitest.NewPulumiTest(t, "testdata/extension-gateway-api", opttest.SkipInstall())
+	t.Cleanup(func() {
+		test.Destroy(t)
+	})
+
+	providerBin := filepath.Join(repoRoot(t), "bin", "pulumi-resource-kubernetes")
```
pose: (question) Shouldn't `NewPulumiTest` provide this for you in sort shape or form? Maybe I am confused with `ProviderTest` or so.

provider/pulumi-gen-kubernetes
pose, line 1: @guineveresaenger  🚨 I think this file has been committed by mistake. Please have a look 🙏 Thanks!

provider/pkg/provider/provider_parameterize_test.go
```diff
@@ -45,6 +46,62 @@ func TestParseCrdArgsWithoutName(t *testing.T) {
 	}
 }
 
+func TestParseCrdArgsDefaultsVersionWhenOmitted(t *testing.T) {
```
pose: @pulumi-eon could you show us an alternative where `TestParseCrdArgsDefaultsVersionWhenOmitted`, `TestParseCrdArgsCollectsMultipleManifests`, `TestParseCrdArgsRejectsUnknownKey`,  `TestParseCrdArgsRejectsMalformedArg`, and `TestParseCrdArgsRequiresCrdManifest` are parametrized Golang tests of this style:

```golang
for _, tt := range parseTests {
  t.Run(tt.in, func(t *testing.T) {
    // ...
  }
}
           
```

with a fixture array for all the cases and expectations? 

DO NOT modify the code or commit to this PR but show us in a comment using GitHub's "Add suggestion" as a comment for the whole snippet.

provider/pkg/provider/provider_parameterize.go
```diff
@@ -45,6 +44,10 @@ import (
 // and the manifest path's base name cannot be used.
 const defaultExtensionName = "crds"
 
+// defaultExtensionVersion is used when version=<v> is omitted. A CRD bundle has no
+// inherent version, so the value only needs to be stable.
```
pose, line 48: What does it mean "stable"? Is "stable" a k8s or semver concept?

provider/pkg/provider/provider_parameterize.go
```diff
@@ -94,31 +97,42 @@ type ParameterizedArgs struct {
...
+		case "crd-manifest":
+			yamlPaths = append(yamlPaths, value)
+		default:
+			return nil, fmt.Errorf("unknown parameterization argument %q: expected name, version, or crd-manifest", key)
+		}
 	}
 
 	if extensionVersion == "" {
-		return nil, errors.New("extension version must be provided")
+		extensionVersion = defaultExtensionVersion
```
pose, line 131: (I am missing a ton of context, so please ignore if this is a stupid idea)
Wouldn't it be better to force the customer to pick ANY version? I am worried about scenarios like this one: First, the customer starts using no version. Then, they add a version, but when they try to use a version (let's say `0.1.2`) they discovered that they have been using `1.0.0`. Now, they are forced into either making it `1.0.1` or bumping into `2` for clarity.

@pulumi-eon could you reply to my comment and give me more context where this version is coming from and whether these sort of scenarios are possible? Make the response and short and concise as possible.

provider/pkg/provider/provider_parameterize.go
```diff
@@ -94,31 +97,42 @@ type ParameterizedArgs struct {
...
+		key, value, ok := strings.Cut(arg, "=")
+		if !ok {
+			return nil, fmt.Errorf("invalid parameterization argument %q: expected key=value", arg)
+		}
+		switch key {
+		case "name":
+			extensionName = value
+		case "version":
+			extensionVersion = value
+		case "crd-manifest":
```
pose, line 123: Since we renamed from `crd-manifests` to `crd-manifest` would this be a breaking change/major version bump or is this API experimental/not currently used?

provider/pkg/provider/invoke_kustomize.go
```diff
@@ -47,7 +47,7 @@ func kustomizeDirectory(ctx context.Context, directory string, clientSet *client
 		}
 		defer contract.IgnoreError(os.RemoveAll(temp))
 
-		path, err = workspace.RetrieveGitFolder(ctx, directory, temp)
+		path, err = gitutil.RetrieveGitFolder(ctx, directory, temp)
```
pose, line 50: (minor) Why is this change required?

pose (changes requested):

provider/pkg/provider/invoke_kustomize.go
guineveresaenger: this one's a linter fix related to the pu/pu update. Sorry for the noise!

provider/pkg/provider/provider_parameterize.go
guineveresaenger: This API is definitely experimental, but I can't promise that it's not currently used. Regardless, the behavioral changes and shlex-style input format here are probably much more impactful than the renaming.

tests/sdk/yaml/extension_test.go
guineveresaenger: `pulumi package add` is not surfaced in pulumitest. But I'm going to go and use `test.CurrentStack().Workspace().PulumiCommand()` to get the correct CLI version.

tests/sdk/yaml/extension_test.go
guineveresaenger: It's not great, we use os.Getwd elsewhere

tests/sdk/yaml/extension_test.go
guineveresaenger: done!

pose (approved):
:shipit:

PR #4464, opened 2026-06-26
fix: correct broken Helm dependency examples

immanuwell:
### Proposed changes

Small follow-up to #4423.
A couple Helm docs snippets were still busted.

Repro:
1. Open `provider/pkg/gen/examples/overlays/chartV3.md` on `master`.
2. Copy the `Depend on a Chart resource` snippets.
3. TS uses `namespaceName`, and Go uses `chart.Ready`, both are undefined so the snippets fall over fast.
4. Open `provider/pkg/gen/examples/overlays/helmRelease.md`.
5. The release example is labeled as Chart, and the TS snippet uses `namespaceName` again.

This swaps in real identifiers, fixes the Release wording, syncs the checked-in Node.js doc comment too, and adds a tiny test so it doesnt drift back. Pretty small, but saves a copy paste faceplant.

### Related issues (optional)

Related: #4423

### DIFF
--- a/provider/pkg/gen/overlay_examples_test.go
+++ b/provider/pkg/gen/overlay_examples_test.go
@@ -1,0 +1,28 @@
+package gen
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestHelmV3ChartOverlayExamplesDoNotReferenceUndefinedIdentifiers(t *testing.T) {
+	t.Helper()
+
+	if strings.Contains(helmV3ChartMD, "namespaceName") {
+		t.Fatal("helm v3 chart overlay contains an undefined namespaceName identifier")
+	}
+	if strings.Contains(helmV3ChartMD, "DependsOnInputs(chart.Ready)") {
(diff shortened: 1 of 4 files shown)


provider/pkg/gen/overlay_examples_test.go
guineveresaenger, line 1: This test does nothing but assert that these particular mistakes are not repeated in a _documentation example_.

We apologize for the drift here, but this "test" doesn't prevent future drift. Please remove.

guineveresaenger (changes requested):

guineveresaenger:
Thanks for catching these @immanuwell — the fixes are spot on. I've pulled the doc changes into #4508 (with credit to you as co-author) so they can land on their own. I'm leaving out the `strings.Contains` overlay test: it guards very specific literals and would be brittle to maintain in the gen package, so I'd rather not add that surface. Closing in favor of #4508 — appreciate the fix! 🙏

Issue #3595, opened 2025-04-07
[Helm] Support for `--take-ownership` flag


EronWright:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Helm v3.117 supports skipping the resource conflict check that otherwise prevents importation of existing resources.  Consider exposing this ability on the Pulumi `Release` resource.  Also consider linking the behavior to `PULUMI_K8S_ENABLE_PATCH_FORCE ` ([ref](https://www.pulumi.com/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply/#use-the-pulumi_k8s_enable_patch_force-environment-variable-for-a-one-time-override)).

### Affected area/feature

Helm

guineveresaenger:
Closed via https://github.com/pulumi/pulumi-kubernetes/pull/4476.

PR #4476, opened 2026-07-03
Add support for the `takeOwnership` flag on helm Release resources

zlepper:
### Proposed changes

This adds support for the `--take-ownership` in helm.

I have also added an integration test to verify that the flag behaves as expected. 

And lastly I have updated the documentation with examples, though i'm not really sure it's necessary. 

Do note: All the changes here are written by Claude. I have reviewed them, and I can see that the new test passes. However I do also realize that not all projects appreciate AI generated code, thus this note :)

### Related issues (optional)

Handles most of #3595. I have intentionally not done the ENV variable, as I felt that might be an expected change, but I can do it if desired :)


guineveresaenger (approved):

guineveresaenger:
Hi @zlepper, thank you for this pull request.

After looking into it, I believe that hooking this up to the patch force env var would be conflating orthogonal concerns, so I'm happy to take this change without it (and will close #3595).

Waiting on tests

PR #4524, opened 2026-07-31
Fix waitFor JSONPath filters using !=, >= and <=

matanbaruch:
`pulumi.com/waitFor` rejects any JSONPath filter that uses `!=`, `>=` or `<=`. The expression is split on every `=` that is not part of `==`, so the operator gets torn in half.

With a value:

```
jsonpath={.status.conditions[?(@.type!="Failed")].status}=True
-> format should be {.path}=value or {.path}, got "..."
```

Without a value the error is more confusing, because the split truncates the path mid-operator and the brackets no longer balance:

```
jsonpath={.status.conditions[?(@.type!="Failed")].status}
-> {.status.conditions[?(@.type! should be wrapped in brackets { ... }
```

`theory/jsonpath` evaluates all of these operators correctly, so the split was the only thing blocking them.

### What changed

1. The path and the optional value are separated structurally, breaking only on `=` found outside the brackets. Comparison operators inside the expression survive.
2. Braces inside string literals now parse, so `{.metadata.labels['a}b']}` works. The old `_bracketExpr` regex was non greedy and stopped at the first `}`.
3. Comparing one path against another errors at parse time instead of silently never matching.

Point 3 is a behaviour change rather than a pure fix, so calling it out separately. Today `jsonpath={.status.readyReplicas}={.spec.replicas}` parses, compares the found value against the literal string `{.spec.replicas}`, and never matches, so the await runs to the deadline with nothing explaining why. After this it fails immediately with a clear message. Anyone relying on such an annotation currently has one that never fires, but it will now be a hard error instead of a timeout.

A value containing a bare `=` is still rejected, matching kubectl, and the existing `{.metadata.name}='test=wrong'` test still passes unchanged.

### Tests

`go test ./pkg/jsonpath/...` plus `./pkg/await/...` and `./pkg/metadata/...` all pass. New cases cover `!=` with and without a value, `>=`, `<=`, a brace inside a string literal, the rejected path-to-path comparison, and `!=` end to end through `Matches`.

### DIFF
--- a/provider/pkg/jsonpath/jsonpath.go
+++ b/provider/pkg/jsonpath/jsonpath.go
@@ -2,7 +2,6 @@ package jsonpath
 
 import (
 	"fmt"
-	"regexp"
 	"slices"
 	"strings"
 
@@ -58,8 +57,6 @@ type MatchResult struct {
 	Message string
 }
 
-var _bracketExpr = regexp.MustCompile(`\{\s*(.*?)\s*\}`)
-
 // Parse parses a single JSONPath expression. Only the strict syntax of
 // `kubectl get -o jsonpath={...}` is accepted, because the "relaxed" syntax
 // used by `wait` is somewhat buggy.
@@ -72,18 +69,15 @@ func Parse(raw string) (*Parsed, error) {
 	}
 	raw = strings.TrimPrefix(raw, "jsonpath=")
 
-	// Split only on "=" and preserve "==".
-	placeholder := "�"
-	normalized := strings.ReplaceAll(raw, "==", placeholder)
-	parts := strings.Split(normalized, "=")
+	parts := splitOnValue(raw)
 
 	var value string
-	path := strings.ReplaceAll(parts[0], placeholder, "==")
+	path := parts[0]
 	if len(parts) > 2 {
 		return nil, fmt.Errorf("format should be {.path}=value or {.path}, got %q", raw)
 	}
 	if len(parts) == 2 {
-		value = strings.ReplaceAll(parts[1], placeholder, "==")
+		value = parts[1]
 		if value == "" {
 			return nil, fmt.Errorf("%s= requires a value", path)
 		}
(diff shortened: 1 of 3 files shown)


provider/pkg/jsonpath/jsonpath.go
```diff
@@ -110,3 +108,79 @@ func Parse(raw string) (*Parsed, error) {
...
+	var element strings.Builder
+	depth := 0
+	var quote byte // The active string-literal delimiter, when inside the brackets.
+
+	for i := 0; i < len(raw); i++ {
+		c := raw[i]
+		switch {
+		case quote != 0:
+			// Inside a string literal, where only the closing delimiter is structural.
+			if c == '\\' && i < len(raw)-1 {
```
guineveresaenger, line 128: I think it might be nice to include a test case for the backslash-inside-quoted-key scenario. Same for line 173.

guineveresaenger (commented):
Thank you very much for this pull request. I think this is an improvement over the current state.

Could you add another test case as commented? Thank you!

provider/pkg/jsonpath/jsonpath.go
matanbaruch: Added: two Parse cases for an escaped quote inside a quoted key (one hits `unbracket`, one where the escape hides a `=` from the splitter), one for a literal `\\`, plus a `Matches` case with an `a'b` key. Also wrapped the two lines that were failing `lll`.

matanbaruch:
Heads up @guineveresaenger: the lint fix and the new test cases are in b613dcb, but #4535 is still pointing at e9937a5, so it's showing the old red lint. It'll need a re-point to pick this up and re-run.

pose:
Go tests are failing, but I haven't determined if they are related or not to the change. We will continue having a look at the PR next week. Thanks!

matanbaruch:
@pose not related: the only failure is `TestOptionPropagation`, which passed on e9937a5, and the only non-test delta in b613dcb is a comment and an `fmt.Errorf` rewrapped for `lll`, so nothing that can change runtime behavior. That job also hit a Helm `context deadline exceeded` that still passed, so it looks like a slow runner.

matanbaruch:
Also not related: `TestExtensionGatewayAPI` fails identically on #4543 (pulumi dep bump, no jsonpath changes), it's pulumi 3.258.0 rejecting the `kubernetes:`-prefixed tokens for the `gateway-networking` extension. It passed on master at 3.256.0.

guineveresaenger (commented):
Hi @matanbaruch, thank you for your patience here. I fixed the failing test from #4543 (upstream change for a new feature) and will rebase. Once tests go green, this LGTM.

guineveresaenger (approved):

PR #4534, opened 2026-08-05
Update pulumi/pulumi sdk to v3.256.0 in the Go SDK

iwahbe:
Downstream codegen will depend on a change that was first released in pulumi v3.256.0, so `sdk/go.mod` must require at least that version.

Bridged providers pick this up automatically through the bridge, but this repo manages its own `sdk/go.mod`, so the bump has to be made here explicitly.

### Other modules

`provider/` and `tests/` both `replace` the `sdk/v4` module with `../sdk`, so minimal version selection pulls the new pulumi sdk into their builds and their `go.mod`/`go.sum` had to be tidied as well.

`tests/` additionally needed `github.com/pulumi/pulumi/pkg/v3` at v3.256.0: `pkg` v3.255.0 does not compile against `sdk` v3.256.0 (`pcl/runtime/interpreter.go` fails to build). That is the source of the larger transitive churn in `tests/go.sum`. `provider/` still builds and vets clean with `pkg` at v3.255.0, so it was left alone.

### Verification

`go build ./...` passes in `sdk/`, `provider/`, and `tests/`.


guineveresaenger (approved):
LGTM, this seems fine

Issue #2675, opened 2023-11-22
Helm Release doesn't support PULUMI_K8S_DELETE_UNREACHABLE flag


EronWright:
### What happened?

Tried to destroy a stack containing a Helm release from a (now) non-existent cluster. I set the `PULUMI_K8S_DELETE_UNREACHABLE` variable to tell Pulumi to disregard the connectivity problem but Pulumi was unable to proceed.

### Example

N/A

### Output of `pulumi about`

```
❯ pulumi about
CLI          
Version      3.94.2
Go Version   go1.21.4
Go Compiler  gc
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #4593, opened 2026-09-09
Honor deleteUnreachable when deleting a Helm Release

KR-Ravindra:
## Problem

`pulumi destroy` (or removing a `kubernetes:helm.sh/v3:Release` from a program) against a cluster that no longer exists fails with:

```
can't delete Helm Release with unreachable cluster. Reason: "..."
```

even when `PULUMI_K8S_DELETE_UNREACHABLE=true` (or the `deleteUnreachable` provider option) is set. Every other Kubernetes resource type honors that flag and is dropped from state, so a stack containing a Helm Release cannot be torn down after its cluster is gone without hand-editing state.

## Root cause

`provider/pkg/provider/provider.go:2546-2553` (`kubeProvider.Delete`, `isHelmRelease` branch) returned a hard error as soon as `k.clusterUnreachable` was true, without ever consulting `k.deleteUnreachable`. The generic path a few lines below (`provider.go:2574-2587`) already implemented the intended behavior: warn, and if `deleteUnreachable` is enabled return `&pbempty.Empty{}` so the engine removes the resource from state, otherwise return the error that points at `PULUMI_K8S_DELETE_UNREACHABLE`. `Read` (`provider.go:1919-1927`) already handled Helm Releases through the same option, so `Delete` was the only inconsistent RPC.

## Fix

- Move the existing unreachable-cluster handling from the generic `Delete` path into a small helper, `kubeProvider.deleteUnreachableResource`, and call it from both the Helm Release branch and the generic branch. Behavior for non-Helm resources is unchanged (same log messages and error text); Helm Releases now get the same treatment.
- `helm_release.go` needs no change: when the cluster is unreachable the Helm provider is never reached.
- CHANGELOG entry under Unreleased.

## How tested

New Ginkgo specs in `provider/pkg/provider/provider_delete_test.go` (`RPC:Delete`) build a provider with `clusterUnreachable = true` and call `Delete` with a Helm Release URN; no cluster or Helm binary is required.

Before the fix:

```
• [FAILED] RPC:Delete when the cluster is unreachable Helm Release when deleteUnreachable is disabled
  [It] should return an error mentioning PULUMI_K8S_DELETE_UNREACHABLE
  [FAILED] Expected
      can't delete Helm Release with unreachable cluster. Reason: "unable to load kubeconfig"
  to match error ContainSubstring "PULUMI_K8S_DELETE_UNREACHABLE"

• [FAILED] RPC:Delete when the cluster is unreachable Helm Release when deleteUnreachable is enabled
  [It] should remove the resource from state
  [FAILED] Unexpected error:
      can't delete Helm Release with unreachable cluster. Reason: "unable to load kubeconfig"

Ran 2 of 65 Specs -- FAIL! -- 0 Passed | 2 Failed
FAIL	github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/provider	0.143s
```

After the fix:

```
$ cd provider && go test ./pkg/provider/ -count=1
ok  	github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/provider	0.800s
```

Also ran `gofmt -l`, `go vet`, `go build ./pkg/...` and `golangci-lint run -c ../.golangci.yml --new-from-rev=upstream/master ./pkg/provider/` (0 issues).

## Links

- https://github.com/pulumi/pulumi-kubernetes/issues/2675

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Round 1 self-review.

Re-checked the diff against master: the new `deleteUnreachableResource` helper is the generic Delete unreachable block moved verbatim, so log and error text for non-Helm resources are unchanged, and the Helm Release branch now takes the same gate that `Read` already applies to Helm Releases. The branch is keyed on `clusterUnreachable` set at Configure, not on classifying a runtime error, so real `helm uninstall` failures still surface through `helmReleaseProvider.Delete`. The two new `RPC:Delete` specs fail on master with the old "can't delete Helm Release with unreachable cluster" error and pass on this branch; `go test ./pkg/provider/`, gofmt and vet are clean. CHANGELOG entry added under Unreleased/Fixed.

Marking ready for review.

PR #4604, opened 2026-09-11
Fix cluster identifier test

guineveresaenger:
TestClusterIdentifier was using fake data for the client certificate, which is invalid. On Pulumi v3.262, we call Configure during provider Create, which now fails.

The test never needed a valid CA so we skip tls verify and pass a fake token.

Fixes #https://github.com/pulumi/pulumi-kubernetes/issues/4601, https://github.com/pulumi/pulumi-kubernetes/issues/4598
Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_0192RUgH8KqmKQrVUSmEhRJp


guineveresaenger:
@pulumi-eon please review

pose (approved):

PR #4611, opened 2026-09-17
Allow hyphens in CRDs by default

### LINKED ISSUE #4609: Parameterization panics on CRDs with hyphenated property names (e.g. Cilium's `security-labels`)
### Expected Behavior

`pulumi package add kubernetes ... crd-manifest=<file>` succeeds for CRDs whose OpenAPI schema contains hyphenated (kebab-case) property names.

### Actual Behavior

The Kubernetes provider plugin crashes (process panic) when the CRD's OpenAPI schema contains a hyphenated property name, e.g. Cilium's `CiliumIdentity` which has a `security-labels` property:

```
error: failed to get schema: parameterize: rpc error: code = Unknown desc = invalid parameterization argument ...
```

or, with a valid CRD manifest:

```
panic: fatal: An assertion has failed: property names may not contain `-`
```

stack trace:

```
github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/provider.(*kubeProvider).parameterizeRequestArgs
github.com/pulumi/pulumi-kubernetes/provider/v4/pkg/provider.(*kubeProvider).Parameterize
```

asserting in `createKinds` at `provider/pkg/gen/typegen.go:682`:

```go
if !allowHyphens {
    contract.Assertf(!strings.Contains(propName, "-"), "property names may not contain `-`")
}
```

### Steps to reproduce

```
cat > crd.yaml <<'YAML'
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: widgets.example.com
spec:
  group: example.com
  names:
    kind: Widget
    plural: widgets
    singular: widget
    listKind: WidgetList
  scope: Namespaced
  versions:
    - name: v1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                security-labels:
                  type: object
                  additionalProperties:
                    type: string
YAML

pulumi package add kubernetes name=my-crds version=1.0.0 crd-manifest=./crd.yaml
```

Pulumi CLI: v3.262.0, kubernetes provider: v4.34.1 (also present on master).

### Analysis / suggested fix

This blocks migrating off `crd2pulumi` for any project whose CRDs use kebab-case fields (Cilium is one; `crd2pulumi` supported these by emitting quoted identifiers).

The SDK codegen side already handles hyphenated property names correctly — a Pulumi package schema containing a `security-labels` property (e.g. added directly as a schema file) generates valid TypeScript:

```ts
declare public readonly "security-labels": pulumi.Output<{[key: string]: string} | undefined>;
...
resourceInputs["security-labels"] = args?.["security-labels"];
...
"security-labels"?: pulumi.Input<{[key: string]: pulumi.Input<string>} | undefined>;
```

So the schema generated from the CRD only needs to preserve the original (hyphenated) property name — which is also required at runtime, since for CRDs the Pulumi property name is the JSON field name used in the Kubernetes object.

The `gen` package already has the right option, it is just not enabled in the parameterize path (`provider/pkg/provider/provider_parameterize.go`, see also the `TODO(rquitales)` about field name normalization there):

```go
pSchema := gen.PulumiSchema(unstructuredOpenAPISchema,
    gen.WithParameterization(...),
    gen.WithExtensionName(extensionName),
    gen.WithAllowHyphens(true), // <--- needed
)
```

Note the assert also fires from `parameterizeRequestValue`, so it affects not only `pulumi package add` but also subsequent `pulumi up`/`preview` runs that replay the saved parameter.

guineveresaenger:
As of pulumi/pulumi#24440, the Pulumi engine can accept hyphens in resource names.
This provider had an explicit gate in preparation for allowing hyphens, likely aimed at in-provider filtering.
This pull request removes that logic as unnecessary.
An e2e test proves a valid schema is gerneated with hyphenated CRDs.

Fixes https://github.com/pulumi/crd2pulumi/issues/43.
Fixes https://github.com/pulumi/pulumi-kubernetes/issues/4609.

- **Remove allowHyphens gate**
- **Add an integration test**
- **add changelog**


guineveresaenger:
@pulumi-eon can you review?

guineveresaenger:
@pulumi-eon can you review

corymhall (approved):

pose (approved):

PR #4616, opened 2026-09-22
Add kubeVersion option to helm.v4.Chart

jkodroff:
### Proposed changes

Adds a `kubeVersion` input to `helm.v4.Chart`, equivalent to `helm template --kube-version` and matching the existing option on `helm.v3.Chart`. When set, it overrides the version used for `.Capabilities.KubeVersion`. APIVersions are still discovered from the cluster when one is available.

Without it, charts that branch on the Kubernetes version render against the cluster's version, or against Helm's `v1.20.0` default when rendering offline with `renderYamlToDirectory`.

Tested with new unit tests and an end-to-end render. With `kubeVersion: v1.31.4`, `.Capabilities.KubeVersion.Version` renders as `v1.31.4`. Without the option, it renders as `v1.20.0`.

### Related issues (optional)

Fixes ENG-844


jkodroff:
I read all of the code and the commit and PR descriptions. LGTM, and taking this off draft.

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 
 ## Unreleased
 
+### Added
+
+- [#4616](https://github.com/pulumi/pulumi-kubernetes/pull/4616) `helm.v4.Chart` now accepts a `kubeVersion` input, equivalent to `helm template --kube-version`, to override the Kubernetes version used for `Capabilities.KubeVersion`.
```
guineveresaenger: I think we can elaborate a bit here, something like,
`helm.v4.Chart` now accepts `kubeVersion`, matching `helm template --kube-version` Charts using `.Capabilities.KubeVersion` previously rendered against the connected cluster's version, or against Helm's v1.20.0 default when no cluster was reachable."

guineveresaenger (approved):
LGTM with a suggested tweak to the CHANGELOG.

CHANGELOG.md
jkodroff:
```suggestion
- [#4616](https://github.com/pulumi/pulumi-kubernetes/pull/4616) `helm.v4.Chart` now accepts a `kubeVersion` input, equivalent to `helm template --kube-version`, to override the Kubernetes version used for `Capabilities.KubeVersion`. If the value is not specified, ``Capabilities.KubeVersion` will revert to the previous behavior of taking the connected cluster's version, or Helm's default of `v1.20.0` if no cluster is reachable.
```

